PR #185, opened 2022-06-22
feat: 👀 Use 100 unit for percentage charge calculation

MorganeLecurieux:
image

Original discussion : https://getlago.slack.com/archives/C03B4D0AANS/p1655804775928309


ansmonjol (approved):
🤑

PR #230, opened 2022-07-07
feat: 🏷 display app version in side nav

ansmonjol:
This MR prompts the app version in side menu popper, bellow logout button.

image

### DIFF
--- a/src/layouts/SideNavLayout.tsx
+++ b/src/layouts/SideNavLayout.tsx
@@ -1,15 +1,25 @@
+import { gql } from '@apollo/client'
 import { Outlet } from 'react-router-dom'
 import { useState, useRef, useEffect } from 'react'
 import styled from 'styled-components'
-import { ClickAwayListener } from '@mui/material'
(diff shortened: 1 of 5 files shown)


src/layouts/SideNavLayout.tsx
```diff
@@ -162,6 +174,17 @@ const SideNav = () => {
                       onClick={() => logOut(client, true)}
                     />
                   </Logout>
+                  <Version>
```
MorganeLecurieux, line 177: So, since it's an endpoint, what happens when there is an error and we have no link and no version ?
(Same regarding the loading)

MorganeLecurieux (approved):

src/layouts/SideNavLayout.tsx
ansmonjol: It has been handled this was
- Loading have receive a skeleton, in place of the expected text to be dispayed
- If an error occur, the version section is hidden and the global error catch will display the error toast

PR #264, opened 2022-07-19
fix: 🔥 ButtonsLink + NavigationLink

MorganeLecurieux:
Fix the bug [CMD + Click does not open link in a new tab](https://www.notion.so/getlago/Delivery-cockpit-d3511c278d574a3aa82986b0d7caf8e1?p=6c4deec62b4f4d8695a45fa104363a81)

### DIFF
--- a/src/components/designSystem/ButtonLink.tsx
+++ b/src/components/designSystem/ButtonLink.tsx
@@ -1,0 +1,127 @@
+import { forwardRef, ReactNode, MouseEvent } from 'react'
+import { Link } from 'react-router-dom'
+import styled, { css } from 'styled-components'
+
+import { theme } from '~/styles'
+
+import { Icon, IconName } from './Icon'
+import { Typography } from './Typography'
+
+export interface ButtonLinkProps {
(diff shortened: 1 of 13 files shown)


src/components/designSystem/ButtonLink.tsx
```diff
@@ -0,0 +1,127 @@
...
+  $active?: boolean
+  $disabled?: boolean
+  children: ReactNode
+  onMouseDown: (e: MouseEvent<HTMLAnchorElement>) => void
+}
+
+const ConditionalWrapper = forwardRef<HTMLAnchorElement, ConditionalWrapperProps>(
+  ({ external, to, children, ...props }: ConditionalWrapperProps) => {
+    return external ? (
+      <ExternalLink href={to} {...props}>
```
ansmonjol, line 30: For external link should we force by default rel="noopener noreferrer" ? 
And target="_blank" could be given during invocation in case we do not want to open in new tab in some cases WDYT ?

ansmonjol (approved):

PR #269, opened 2022-07-20
feat: Create wallet for customer

ansmonjol:
This MR allows the creation of a wallet attached to a customer


src/components/customers/AddWalletToCustomerDialog.tsx
```diff
@@ -0,0 +1,221 @@
...
+        rateAmount: '1.00',
+      },
+      validationSchema: object().shape({
+        expirationDate: date(),
+        grantedCredits: string().required(''),
+        name: string(),
+        paidCredits: string().required(''),
+        rateAmount: string().required(''),
+      }),
+      enableReinitialize: true,
```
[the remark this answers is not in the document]
ansmonjol: Nope indeed, removed

src/components/customers/AddWalletToCustomerDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+            },
+          },
+          refetchQueries: ['getCustomer'],
+        })
+
+        const { errors } = res
+
+        if (!errors) {
+          ;(ref as unknown as RefObject<DialogRef>)?.current?.closeDialog()
+        }
```
MorganeLecurieux, line 77: You don't need that here, just do it on the onClick after `formikProps.submitForm` 
We use this only if we need to display an error in the modal

src/components/customers/AddWalletToCustomerDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+          <TextInputField
+            name="name"
+            label={translate('text_62d18855b22699e5cf55f875')}
+            placeholder={translate('text_62d18855b22699e5cf55f877')}
+            formikProps={formikProps}
+          />
+          <InlineFields>
+            <TextInput value="1" label={translate('text_62d18855b22699e5cf55f879')} />
+            <TextInput value="=" />
+            <TextInputField
```
MorganeLecurieux, line 113: This one should be disabled according to the design no ?

src/components/customers/AddWalletToCustomerDialog.tsx
ansmonjol: won't that close the modal if any backend error occcur ? 500 or whatever

src/components/customers/AddWalletToCustomerDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+              name="rateAmount"
```
MorganeLecurieux, line 114: Same here, should be disabled

src/components/customers/AddWalletToCustomerDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+            />
+          </InlineFields>
+
+          <TextInputField
+            name="paidCredits"
+            beforeChangeFormatter={['positiveNumber', 'decimal']}
+            label={translate('text_62d18855b22699e5cf55f885')}
+            placeholder="0.00"
+            formikProps={formikProps}
+            helperText={translate('text_62d18855b22699e5cf55f88b', {
```
MorganeLecurieux, line 131: IMO the placeholders should be a translation as well (for the other fields too)

src/components/customers/AddWalletToCustomerDialog.tsx
ansmonjol: Yes good catch ! done

src/components/customers/AddWalletToCustomerDialog.tsx
ansmonjol: Done

src/components/customers/AddWalletToCustomerDialog.tsx
MorganeLecurieux: There is a toast that will appear but yes it will close the modal. This is the behaviour we have everywhere right now, but feel free to challenge it with Mike if you think it's not ok :)

ansmonjol: Ok got it. Done

PR #297, opened 2022-08-02
Wallet modals

ansmonjol:
This PR adds 3 modals in customer wallet page.

They allow to Topup a wallet, edit their basic informations or terminate them, only if an active wallet is present.

<div style="display: flex">
image
image
image
</div>


src/components/wallets/AddWalletToCustomerDialog.tsx
```diff
@@ -21,16 +17,25 @@ gql`
       id
     }
   }
+
+  fragment SubscriptionForWalletCreation on Subscription {
```
MorganeLecurieux, line 21: I don't understand the use of this fragment ?

src/components/wallets/WalletTransactionList.tsx
```diff
@@ -46,149 +50,162 @@ interface WalletTransactionListProps {
   wallet: WalletInfosForTransactionsFragment
```
MorganeLecurieux, line 50: You're just using wallet.id, so you can just pass the wallet id and remove the fragment (that is not really usefull here since status and currency are not used)

src/components/wallets/AddWalletToCustomerDialog.tsx
ansmonjol: I need the `amountCurrency` from the `plan` in this component, passed in props named `userCurrency`

MorganeLecurieux: Yes but `userCurrency` is a string, you don't pass the wallet, so no need for a fragment here

MorganeLecurieux (approved):

src/components/wallets/AddWalletToCustomerDialog.tsx
ansmonjol: indeed ! Code updated

PR #319, opened 2022-08-09
bug: fix display for customer without name

ansmonjol:
This PR fixes the UI to properly display customer if no name is provided. 
This is a rare case if you attach a subscription to a customer without name, it will create an empty user in DB and lead to a wrong display in UI.

BEFORE
Name and avatar were not displayed

AFTER
image


src/components/customers/CustomerItem.tsx
```diff
@@ -54,11 +54,11 @@ export const CustomerItem = memo(({ rowId, customer }: CustomerItemProps) => {
           <ListAvatar
             variant="user"
             identifier={name as string}
-            initials={(name || '').split(' ').reduce((acc, n) => (acc = acc + n[0]), '')}
+            initials={(name || '-').split(' ').reduce((acc, n) => (acc = acc + n[0]), '')}
```
MorganeLecurieux, line 57:
```suggestion
            initials={!name ? '-' : name.split(' ').reduce((acc, n) => (acc = acc + n[0]), '')}
```

MorganeLecurieux (approved):

PR #337, opened 2022-08-10
[Chore]: Cypress setup + first tests

### LINKED ISSUE #325: [CHORE]: Setup Cypress

MorganeLecurieux:
Closes #325

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -1,0 +1,56 @@
+name: TestE2E
+
+on:
+  pull_request:
+    types: [review_requested, ready_for_review, synchronize]
+
+jobs:
+  cypress-run: 
+    if: github.event.pull_request.draft == false
+    name: Run Test E2E
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
+
+      - name: Install Node.js dependencies
+        run: yarn
+
+      - name: Build Front local image
+        run: |
+          docker build -t getlago/front:ci ./
+
+      - name: Checkout API Git repository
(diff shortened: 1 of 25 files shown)


cypress/e2e/auth/t0-signup.cy.ts
```diff
@@ -0,0 +1,97 @@
...
+    cy.visit('sign-up')
+  })
+  /** TODO -The following test is skipped
+   * It will work once as this email is not used yet
+   * We need to re-seed between any test to make it work.
+   */
+  it('should create a new user and redirect to the home after signup', () => {
+    cy.visit('sign-up')
+
+    cy.get('input[name="organizationName"]').type('Company name')
```
ansmonjol, line 12: Just a suggestion here.

I'm wondering if we should always use `data-test` and avoid any other kind of selector.

package.json
```diff
@@ -62,6 +63,7 @@
     "babel-plugin-prismjs": "2.1.0",
     "babel-plugin-styled-components": "2.0.7",
     "css-loader": "6.7.1",
+    "cypress": "^10.4.0",
```
ansmonjol, line 66: We'll have a renovabot MR following to fix the version. BTW not sure if we should 🤔 but that's a different topic ˆˆ

You could fix it in this PR, as you wish

ansmonjol (approved):
Few minor comments

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,56 @@
...
+    name: Run Test E2E
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
```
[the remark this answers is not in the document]
MorganeLecurieux: Nop :p 
We can do it later on a different PR but let's update it for all the CIs + test that everything works fine

PR #353, opened 2022-08-17
feat: 🔣 percentage charge model with free units

ansmonjol:
## ⁉️ Problem or Concept

The initial solution for applying free units or fixed units to a percentage pricing model is not working for most of the companies out there. It’s a mandatory feature for Lydia.

## 🗺 Current situation

Applying a fixed units or free units to the % pricing is based on the `sum` aggregation initially used to calculate the total amount. However, the free units or fixed units are not directly calculated on this monetary sum value, it is calculated over a `count` of distinct events/transactions.

## 🛑 Out of scope

- Display free units on the invoice
- Changing the `unit` value in invoice items. For now, we keep this as `units`

## ✅ Success

- [ ]  One user implementing a % charge

## 👩‍🦳 User Preferences

- [ ]  Calculating the number of events for a specific period
- [ ]  Applying a fixed fee to the number of events, and not to the total amount of units (1.5% + 0.50$ per transaction)
- [ ]  Applying free units
    - [ ]  Apply free units to the number of events (3 free transactions, then 1.5%)
    - [ ]  Apply free units to the total amount of units (500$ free, then 1.5%)
    - [ ]  The first to reach the threshold between the number of events or the total amount of units (3 free transactions OR 500$ free, then 1.5%)
- [ ]  Update the invoice. For each % charge model, return the number of events and the total amount of units (3 transactions for X units)

Full specs: https://www.notion.so/getlago/Spec-Percentage-charge-model-with-free-units-d9c7d5c3d24a4df0bc156b340b03e2ee


src/components/plans/ChargePercentage.tsx
```diff
@@ -192,25 +274,35 @@ const Container = styled.div`
...
 
-  > *:first-child {
+  > *:not(:last-child) {
     margin-right: ${theme.spacing(3)};
-    flex: 1;
   }
 
   > *:last-child {
-    transform: translateY(50%);
+    margin-top: 34px;
```
MorganeLecurieux, line 293: Hum... This is not from the spacing system 🤔

src/components/plans/ChargePercentage.tsx
```diff
@@ -192,25 +274,35 @@ const Container = styled.div`
...
   }
 `
 
-const StyledComboBox = styled(ComboBox)`
-  max-width: 120px;
-  margin-top: ${theme.spacing(6)};
-  margin-right: ${theme.spacing(3)};
+const OrText = styled(Typography)`
+  flex: initial;
+  margin-top: 34px;
```
MorganeLecurieux, line 299: This also

src/hooks/plans/useCreateEditPlan.ts
```diff
@@ -48,7 +48,8 @@ gql`
       packageSize
       rate
       fixedAmount
-      fixedAmountTarget
+      freeUnitsPerEvents
+      freeUnitsPerTotalAggregation
```
MorganeLecurieux, line 52: This one is not used here

MorganeLecurieux (approved):

src/hooks/plans/useCreateEditPlan.ts
ansmonjol: As discussed together, I created a refactor issue to fix this (and other in this file) later

https://github.com/getlago/lago-front/issues/354

src/components/plans/ChargePercentage.tsx
ansmonjol: Seen together,  the use case was a bit special and we decided to keep those values as they are

src/components/plans/ChargePercentage.tsx
ansmonjol: Same as above

ansmonjol:
Converted as draft until BA is ready for QA

PR #360, opened 2022-08-22
refactor: display currency symbol instead of code

ansmonjol:
We decided to display currency symbol instead of currency code in app

EUR -> €


MorganeLecurieux:
Closing this as it's to old to redo it from the actual branch

PR #368, opened 2022-08-24
feat: volume charge

ansmonjol:
## 🗺 Current situation

We offer 4 price models for a charge: 

- Standard
- Percentage
- Graduated
- Package

## ✅ Success

- [ ]  Propose volume charge as a new price model

## 👩‍🦳 User Preferences

As a user I can …

- [ ]  Choose `volume` charge model for a billable metic in the plan definition
- [ ]  Add as many tiers I want
- [ ]  Define a last unit of each tiers
- [ ]  Define a price per unit of each tiers
- [ ]  Define a flat fee of each tiers
- [ ]  Validate my volume matrice only if each tiers got either a price per unit or flat fee
- [ ]  Validate my volume matrice only if each tiers got a last unit defined


src/components/plans/PlanForm.tsx
```diff
@@ -159,6 +161,41 @@ export const PlanForm = ({ loading, plan, children, onSave, isEdition }: PlanFor
                 .required(''),
             })
             .nullable(),
+          volumeRanges: array()
+            .when('chargeModel', {
+              is: (chargeModel: ChargeModelEnum) =>
+                !!chargeModel && chargeModel === ChargeModelEnum.Volume,
+              then: array()
+                .test({
+                  test: (graduatedRange) => {
```
MorganeLecurieux, line 170: Not graduated charges here then

MorganeLecurieux (approved):

src/components/plans/PlanForm.tsx
ansmonjol: Good catch! Fixed

PR #415, opened 2022-09-16
Members page

ansmonjol:
This PR adds the new Members section in the settings.

This section allows to perform multiple things:
- List member and pending invites
- Create a new invite
- revoke a membership (except yourself)
- revoke an invite

image

Note: the invitation flow for new joiners will be added in a following PR


src/components/GenericPlaceholder.tsx
```diff
@@ -22,6 +22,17 @@ interface GenericPlaceholderNoButtonProps
   buttonVariant?: never
 }
 
+interface GenericPlaceholderNoButtonAndTitleProps
```
MorganeLecurieux, line 25: Well... No ^^
Just change `GenericPlaceholderProps` and make `title` optional. No need to have a full other type for that.

src/components/GenericPlaceholder.tsx
```diff
@@ -31,7 +42,10 @@ export const GenericPlaceholder = ({
...
   buttonVariant,
   buttonAction,
-}: GenericPlaceholderProps | GenericPlaceholderNoButtonProps) => (
+}:
+  | GenericPlaceholderProps
+  | GenericPlaceholderNoButtonProps
+  | GenericPlaceholderNoButtonAndTitleProps) => (
   <Container className={className} $noMargins={noMargins}>
     {image}
     <Title variant="subhead">{title}</Title>
```
MorganeLecurieux, line 51: If it's optional, you need to check it `{title && <Title variant="subhead">{title}</Title>}`

src/components/GenericPlaceholder.tsx
ansmonjol: Fine with that ! 😄 Will do

src/components/GenericPlaceholder.tsx
ansmonjol: Done

src/components/GenericPlaceholder.tsx
ansmonjol: Done

src/pages/settings/Members.tsx
```diff
@@ -0,0 +1,284 @@
...
+  } = useGetMembersQuery()
+  const hasInvites = !!invitesData?.invites.metadata.totalCount
+  const invitesMetadata = invitesData?.invites.metadata
+  const membersMetadata = membersData?.memberships.metadata
+
+  return (
+    <Page>
+      <Title variant="headline">{translate('text_63208b630aaf8df6bbfb2657')}</Title>
+      <Subtitle>{translate('text_63208b630aaf8df6bbfb2659')}</Subtitle>
+      {(!!invitesLoading || !!invitesError || !!hasInvites) && (
```
MorganeLecurieux, line 91: This condition is weird...
There's no case where you don't display this line no ?
Should be just 
```
<Head>
  {!!invitesLoading ? loadingBlock : normalBlock }
</Head>
```

src/pages/settings/Members.tsx
```diff
@@ -0,0 +1,284 @@
...
+      )}
+      {!!invitesLoading ? (
+        <LoadingListWrapper>
+          {[1, 2].map((i) => (
+            <InviteItemSkeleton key={i} />
+          ))}
+        </LoadingListWrapper>
+      ) : (
+        <InvitationsListWrapper>
+          {!invitesLoading && !!invitesError ? (
```
MorganeLecurieux, line 122: No need to reuse `!invitesLoading` since you're testing that condition 5lines before

src/pages/settings/Members.tsx
```diff
@@ -0,0 +1,284 @@
...
+            <ErrorPlaceholder
+              noMargins
+              title={translate('text_6321a076b94bd1b32494e9e6')}
+              subtitle={translate('text_6321a076b94bd1b32494e9e8')}
+              buttonTitle={translate('text_6321a076b94bd1b32494e9ea')}
+              buttonVariant="primary"
+              buttonAction={() => invitesRefetch()}
```
MorganeLecurieux, line 129:
```
buttonAction={invitesRefetch}
```

no need to recreate a wrapper function here

src/pages/settings/Members.tsx
ansmonjol: In case we don't have invites, we won't display this header. But if we don't have invites but are in error or loading state it should be visible

src/pages/settings/Members.tsx
```diff
@@ -0,0 +1,284 @@
...
+  const revokeInviteDialogRef = useRef<RevokeInviteDialogRef>(null)
+  const revokeMembershipDialogRef = useRef<RevokeMembershipDialogRef>(null)
+  const createInviteDialogRef = useRef<CreateInviteDialogRef>(null)
+  const {
+    data: invitesData,
+    error: invitesError,
+    loading: invitesLoading,
+    refetch: invitesRefetch,
+    fetchMore: invitesFetchMore,
+  } = useGetInvitesQuery()
```
MorganeLecurieux, line 75: You need to set limit here

src/pages/settings/Members.tsx
```diff
@@ -0,0 +1,284 @@
...
+  const {
+    data: membersData,
+    error: membersError,
+    loading: membersLoading,
+    refetch: membersRefetch,
+    fetchMore: membersFetchMore,
+  } = useGetMembersQuery()
```
MorganeLecurieux, line 82: You need to set limit here

src/pages/settings/Members.tsx
ansmonjol: But that's a different condition block 🤔 Not sure to understand

src/pages/settings/Members.tsx
```diff
@@ -0,0 +1,284 @@
...
+      </Head>
+      {!!membersLoading ? (
+        <LoadingListWrapper>
+          {[1, 2, 3].map((i) => (
+            <MembershipItemSkeleton key={i} />
+          ))}
+        </LoadingListWrapper>
+      ) : (
+        <div>
+          {!membersLoading && !!membersError ? (
```
MorganeLecurieux, line 191: No need to check `membersLoading` you're check in it on the parent block

src/pages/settings/Members.tsx
```diff
@@ -0,0 +1,284 @@
...
+            <ErrorPlaceholder
+              noMargins
+              title={translate('text_6321a076b94bd1b32494e9ee')}
+              subtitle={translate('text_6321a076b94bd1b32494e9f0')}
+              buttonTitle={translate('text_6321a076b94bd1b32494e9f2')}
+              buttonVariant="primary"
+              buttonAction={() => membersRefetch()}
```
MorganeLecurieux, line 198: buttonAction={membersRefetch}

src/pages/settings/Members.tsx
ansmonjol: ok nice ! Updated

src/pages/settings/Members.tsx
ansmonjol: updated thanks

src/pages/settings/Members.tsx
ansmonjol: indeed 😬 thanks

src/pages/settings/Members.tsx
```diff
@@ -0,0 +1,284 @@
...
+                }}
+              >
+                {translate('text_63208b630aaf8df6bbfb265b')}
+              </Button>
+            )}
+          </>
+        )}
+      </Head>
+      {!!membersLoading ? (
+        <LoadingListWrapper>
```
MorganeLecurieux, line 184: On an infinite scroll, we put the loading state at the end of the list so when you "fetchMore", you can still see the items previously fetched

And don't forget to add `notifyOnNetworkStatusChange` on your query so you'll see the loading state change

src/pages/settings/Members.tsx
ansmonjol: Good point 🙏

src/pages/settings/Members.tsx
```diff
@@ -0,0 +1,284 @@
...
+
+const Head = styled.div`
+  height: ${NAV_HEIGHT}px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+`
+
+const TitleSkeleton = styled(Skeleton)`
+  margin: 30px 0;
```
MorganeLecurieux, line 254: You have a preperty `marginBottom` on the skeletons to avoid styling them (plus with a spacing out of the DS)

src/components/settings/members/CreateInviteDialog.tsx
```diff
@@ -0,0 +1,193 @@
...
+            })
+          : translate('text_63208c701ce25db78140749b')
+      }
+      onOpen={() => {
+        formikProps.validateForm()
+      }}
+      onClickAway={() => {
+        setInviteToken('')
+        formikProps.resetForm()
+        formikProps.validateForm()
```
MorganeLecurieux, line 89: Why do you need that ?

src/components/settings/members/CreateInviteDialog.tsx
```diff
@@ -0,0 +1,193 @@
...
+      }}
+      actions={({ closeDialog }) => (
+        <>
+          <Button
+            variant="quaternary"
+            onClick={() => {
+              setInviteToken('')
+              closeDialog()
+              formikProps.resetForm()
+              formikProps.validateForm()
```
MorganeLecurieux, line 99: Still not sure about that.. ?

src/components/settings/members/CreateInviteDialog.tsx
```diff
@@ -0,0 +1,193 @@
...
+                await formikProps.submitForm()
+              }}
+            >
+              {translate('text_63208c711ce25db7814074d9')}
+            </Button>
+          ) : (
+            <Button
+              disabled={!!error}
+              onClick={() => {
+                navigator.clipboard.writeText(`${window.location.origin}/invitation/${inviteToken}`)
```
MorganeLecurieux, line 117: You're reusing this, to avoid mistypes or make it more maintainable, please reuse the same `const`

src/components/settings/members/CreateInviteDialog.tsx
```diff
@@ -0,0 +1,193 @@
...
+                addToast({
+                  severity: 'info',
+                  translateKey: 'text_63208c711ce25db781407536',
+                })
+                setInviteToken('')
```
MorganeLecurieux, line 122: You're doing all this a bunch of times, might be a good idea to make a function out of it

src/components/settings/members/InviteItem.tsx
```diff
@@ -0,0 +1,124 @@
...
+          <Typography variant="body">{email}</Typography>
+        </LeftBlock>
+        <RightBlock>
+          <Chip label={translate('text_63208b630aaf8df6bbfb2665')} />
+          <Tooltip placement="bottom-end" title={translate('text_63208b630aaf8df6bbfb265f')}>
+            <ActionButtonIcon
+              icon="duplicate"
+              variant="quaternary"
+              onClick={() => {
+                navigator.clipboard.writeText(`${window.location.origin}/invitation/${token}`)
```
MorganeLecurieux, line 47: Ok so you're reusing it as well, I think it either should be in the `externalUrl` file or at least export it from one of the file to reuse it in the other.
Right now if the url changes somwhere it's gonna be annoying to make sure it's good everywhere and can lead to bugs

src/pages/settings/Members.tsx
ansmonjol: Yeah but in that case I want the margin to be on top also, and the component API does not allows such.

I would either increase the margin bottom from the above component or update the API but I felt it was a lot to do to for not that much advantages. I know we have this rule of DS spacing but happy to discuss about it and change the implem

src/components/settings/members/CreateInviteDialog.tsx
ansmonjol: yeah it's a bit terrible. I got the same issue than last time where the form is not reset and validation is wrong even if I had the good settings in formik.

What I could do it to move all those "resets" in the onOpen hook then it's in only one place at least

src/components/settings/members/CreateInviteDialog.tsx
ansmonjol: I was planning to interpolate `/invitation/` with the future route const WDYT ? Isn't it enough ? as the token changes each time

src/pages/settings/Members.tsx
ansmonjol: Veery good points, thanks. I updated the code

src/components/settings/members/CreateInviteDialog.tsx
ansmonjol: Will use generatePath with the window domain, and update all invocations

src/components/settings/members/InviteItem.tsx
ansmonjol: Will update according to https://github.com/getlago/lago-front/pull/415#:~:text=ansmonjol-,now,-Will%20use%20generatePath

src/pages/settings/Members.tsx
ansmonjol: We agree to keep it like that

src/components/settings/members/CreateInviteDialog.tsx
ansmonjol: I'll do that in the 2nd PR

PR #429, opened 2022-09-19
Invitation flow

ansmonjol:
This PR adds the invitation flow. 

image


src/pages/Invitation.tsx
```diff
@@ -0,0 +1,296 @@
...
+              )}
+            <form>
+              <Input
+                disabled
+                name="email"
+                label={translate('text_63246f875e2228ab7b63dcdc')}
+                value={email}
+              />
+
+              <PasswordBlock>
```
MorganeLecurieux, line 173: Looks a lot like the signup... Shouldn't we do one component reusable both for the signup and here as it uses the same logic ?

MorganeLecurieux (approved):

src/pages/Invitation.tsx
ansmonjol: We'll do that in another PR

PR #433, opened 2022-09-20
chore: create artifact for no modules and delay jobs that needs it

ansmonjol:
This PR improves the CI pipeline by creating an artifact for node_modules, used by many other jobs.

It aims to makes the CI shorter of ~1min per jobs that uses it

### DIFF
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -1,12 +1,15 @@
 name: Run Codegen
 
 on:
-  pull_request:
-    types: [opened, reopened, review_requested, ready_for_review, synchronize]
+  workflow_run:
+    branches:
+      - '**'
+    workflows: [Build node_modules]
+    types:
+      - completed
 
 jobs:
   codegen:
-    if: github.event.pull_request.draft == false
     name: Run Codegen
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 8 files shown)


.github/workflows/codegen.yml
```diff
@@ -1,12 +1,13 @@
...
 
 on:
-  pull_request:
-    types: [opened, reopened, review_requested, ready_for_review, synchronize]
+  workflow_run:
+    workflows: ["Build node_modules"]
+    types: [completed]
 
 jobs:
   codegen:
```
MorganeLecurieux, line 9: Faudrait quand même garder le fait que ça ne run pas quand on est en draft

.github/workflows/cypress.yml
```diff
@@ -1,12 +1,13 @@
...
 
 on:
-  pull_request:
-    types: [opened, reopened, review_requested, ready_for_review, synchronize]
+  workflow_run:
+    workflows: ["Build node_modules"]
+    types: [completed]
 
 jobs:
   cypress-run:
```
MorganeLecurieux, line 9: Same, keep the "non-draft" check

ansmonjol: I have this condition check in the "Build node_modules" workflow. Should we double this condition ?

PR #436, opened 2022-09-22
fix(webpack): force process env variable to return a boolean

ansmonjol:
## Context

App renders a blank page when a process env is not defined (default behaviour)

When Webpack runs, it removes all usages of this variable leading to syntax issues

## Description

We force the interpretation of an undefined variable to return a boolean. So we always have a value.

Note we'll also put a default value in lago docker-compose

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -30,7 +30,7 @@ module.exports = () => {
         APP_ENV: JSON.stringify(APP_ENV),
         API_URL: JSON.stringify(process.env.API_URL),
         APP_VERSION: JSON.stringify(version),
-        LAGO_DISABLE_SIGNUP: process.env.LAGO_DISABLE_SIGNUP,
+        LAGO_DISABLE_SIGNUP: Boolean(process.env.LAGO_DISABLE_SIGNUP),
       }),
     ],
   }


vincent-pochet (approved):

webpack.common.js
```diff
@@ -30,7 +30,7 @@ module.exports = () => {
         APP_ENV: JSON.stringify(APP_ENV),
         API_URL: JSON.stringify(process.env.API_URL),
         APP_VERSION: JSON.stringify(version),
-        LAGO_DISABLE_SIGNUP: process.env.LAGO_DISABLE_SIGNUP,
+        LAGO_DISABLE_SIGNUP: Boolean(process.env.LAGO_DISABLE_SIGNUP),
```
vincent-pochet, line 33: Just to be sure: how this will behave if the env variable is not set (null)?

ansmonjol: `Boolean(null)` returns false

ansmonjol: Boolean returns false for null, undefined, 0 or empty string

vincent-pochet: Nice!

PR #438, opened 2022-09-22
fix(customer): prevent error by resetting providerCustomerId to empty…

ansmonjol:
## Context

When opening the edit customer dialog, we have most of the time a console warning about data type assigned to an inpoout not being correct

## Description

This issue is due to a manual reset we perform on `providerCustomerId` attribute, being reset to a `null` value instead of an empty string


ansmonjol:
Will be fixed with the very near release of a new feature. Closing in the meantime

PR #457, opened 2022-09-28
feat: 🧮 percent coupon

ansmonjol:
## Current situation 

👉 User can’t make a % discount based on the total invoice of a customer
👉 User can’t make a coupon recurring over billing periods

## Description

This PR adds the possibility to create a coupon and...

- [X]  Define the coupon type
    - [X]  Fixed amount
        - [X]  Define the amount + currency
        - [X]  Define the frequency to
            - [X]  Once
            - [X]  Recurring
                - [X]  Define the number of billing period
        - [X]  Define the end date of the coupon object
            - [X]  Set a date
            - [X]  No check → Forever
    - [X]  Percentage
        - [X]  Define the percentage rate
        - [X]  Define the frequency to
            - [X]  Once
            - [X]  Recurring
                - [X]  Define the number of billing period
        - [X]  Define the end date of the coupon object
            - [X]  Set a date
            - [X]  No check → Forever
    - [X]  See the code snippet adapting based on what I configured on the left side

**As a user I can see Coupons list and …**

- [X]  See the caption changes according the settings of a coupons

**As a user I can edit a coupon and …**

- [X]  display all the settings & edit everything if not linked to a customer
- [X]  display all the settings disabled & edit only the name of the coupon & expiration date

**As a user I can apply a coupon to a customer and …**

- [X]  Edit information of this coupon
    - [X]  Choose coupon
    - [X]  Change amount or Percentage rate
    - [X]  Change frequency
        - [X]  Once
        - [X]  Recurring
            - [X]  Number of billing period
    - [X]  Error if currency is not the same as the one linked to plans


src/components/coupons/CouponCaption.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+interface CouponCaptionProps {
+  coupon: CouponItemFragment
+  variant?: TypographyProps['variant']
+}
+
+export const CouponCaption = ({ coupon, variant = 'caption' }: CouponCaptionProps) => {
+  const { translate } = useInternationalization()
+
+  const getCaption = () => {
```
MorganeLecurieux, line 29: You can `memo` this whole component to avoid re-rendering (`memo`, no `useMemo` :) )

src/hooks/useCreateEditCoupon.ts
```diff
@@ -57,10 +64,29 @@ type UseCreateEditCouponReturn = {
...
+    amountCurrency,
+    expirationDate,
+    percentageRate,
+    frequencyDuration,
+    ...others
+  } = values
 
   return {
-    amountCents: Math.round(Number(amountCents) * 100),
+    amountCents:
```
MorganeLecurieux, line 77: You're doing this both here and in AddCouponCustomerDialog right ?

It could be nice to move it into serializers to avoid errors in case of refacto or BE changes

PR #471, opened 2022-10-03
feat(combobox): Display group in startAdorment

ansmonjol:
## Context

We're planning the development of a new feature.

During the scoping, we figured out that combobox that displays values with your losses their context once a value is selecter.

## Description

This PR adds the groupe value before the value itself in the input. Its placed in the input `startAdorment` and does not affect the value itself, only the input display.

This "feature" is only enabled if you passe a value to `renderGroupHeader`.

Monosnap Lago - Local 2022-10-03 16-35-29

### DIFF
--- a/src/components/form/ComboBox/ComboBox.tsx
+++ b/src/components/form/ComboBox/ComboBox.tsx
@@ -33,6 +33,7 @@ export const ComboBox = ({
   disableClearable = false,
   renderGroupHeader,
   virtualized = true,
+  renderGroupInputStartAdornment,
   onChange,
 }: ComboBoxProps) => {
   const { translate } = useInternationalization()
@@ -73,6 +74,13 @@ export const ComboBox = ({
     matchFrom: allowAddValue ? 'start' : 'any',
     stringify: (option) => option.label || option.value,
   })
+  const startAdornmentValue = useMemo(() => {
+    if (!renderGroupInputStartAdornment || !value) return undefined
+
+    let foundGroup = data.find((item) => item.value === value)?.group
+
+    return foundGroup ? renderGroupInputStartAdornment[foundGroup] : undefined
+  }, [data, renderGroupInputStartAdornment, value])
 
   return (
     <Autocomplete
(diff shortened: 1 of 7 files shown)


src/pages/CreateBillableMetric.tsx
```diff
@@ -244,6 +244,17 @@ const CreateBillableMetric = () => {
                         </ComboboxHeader>
                       ),
                     }}
+                    renderGroupInputStartAdornment={[
+                      {
+                        value: translate('text_6310755befed49627644222b'),
```
MorganeLecurieux, line 249: Hum... Why not just use an object like in `renderGroupHeader` to be consistent ?

```
{
  [groupKey]: "translation",
  ...
}
```

src/components/form/ComboBox/ComboBox.tsx
```diff
@@ -89,6 +90,14 @@ export const ComboBox = ({
             infoText={infoText}
             name={name}
             placeholder={placeholder}
+            startAdornmentValue={
```
MorganeLecurieux, line 93: This could be memoized to avoid recalculation at every render

MorganeLecurieux (changes requested):

src/pages/CreateBillableMetric.tsx
ansmonjol: Indeed, way better ! I updated the code with this approach

src/components/form/ComboBox/ComboBox.tsx
```diff
@@ -73,6 +74,13 @@ export const ComboBox = ({
     matchFrom: allowAddValue ? 'start' : 'any',
     stringify: (option) => option.label || option.value,
   })
+  const startAdornmentValue = useMemo(() => {
+    let foundGroup = data.find((item) => item.value === value)?.group
```
MorganeLecurieux, line 78: Sorry 😅

But to avoid unnecessary calculation, can we :
```
if (!renderGroupInputStartAdornment || !value) return undefined 
```

before any check ?

MorganeLecurieux (approved):

src/components/form/ComboBox/ComboBox.tsx
ansmonjol: sure thing !

ansmonjol: done

PR #472, opened 2022-10-03
feat: 💻 JsonEditor

MorganeLecurieux:
Screenshot 2022-10-03 at 17 51 00


src/components/form/JsonEditor/JsonEditor.tsx
```diff
@@ -0,0 +1,282 @@
...
+  infoText?: string
+  error?: string
+  helperText?: string | ReactNode
+  customInvalidError?: string
+  onBlur?: (props: unknown) => void
+  onChange?: (value: string) => void
+  onError?: (err: keyof typeof JSON_EDITOR_ERROR_ENUM) => void
+}
+
+export const JsonEditor = ({
```
ansmonjol, line 37: No "codeEditor" in the end ?

src/components/form/Checkbox/CheckboxField.tsx
```diff
@@ -1,4 +1,5 @@
 import { FormikProps } from 'formik'
+import _get from 'lodash/get'
```
ansmonjol, line 2: Are the changes in this file related to this PR ?

src/components/form/JsonEditor/JsonEditor.tsx
MorganeLecurieux: No because it's just for Json, CodeEditor would be good if we could configure the language which is not the case right now

src/components/form/Checkbox/CheckboxField.tsx
MorganeLecurieux: No, but I saw it so I changed it

ansmonjol (approved):

PR #523, opened 2022-10-31
feat (gocardless): Handle changes for Gocardless integration

lovrocolic:
This PR handles few Stripe changes but mostly covers Gocardless integration changes:

   a. on Stripe Integration page, we don't have bulk create toggle button anymore (this logic is moved to customer level)

   b. New Gocardless integration page is added

   c. There are few changes on customers: Before updating payment provider fields, user has to select payment provider first. New field `sync in provider` is added on customer level


src/components/customers/AddCustomerDrawer.tsx
```diff
@@ -49,6 +50,14 @@ export const AddCustomerDrawer = forwardRef<DrawerRef, AddCustomerDrawerProps>(
     const { isEdition, onSave } = useCreateEditCustomer({
       customer,
     })
+    const [isDisabled, setIsDisabled] = useState<boolean>(false)
+    const handleChange = (checked: boolean) => {
```
ansmonjol, line 54: Probably you can move this method directly into the checkbox onChange function.

Also `isDisabled` could check on `providerCustomer.syncWithProvider` value directly instead of being a `useState`

lovrocolic: Fixed first only first part (as discussed offline)

ansmonjol (approved):

PR #552, opened 2022-11-08
chore(column width): adapt customer list column size

ansmonjol:
## Context

On small screen, just before the breakpoint, the customer name is not shown in customer list

## Description 

This PR reduce the subscriptions column size that don't needs to be that big, in order to give more space to the customer name column.

|before|after|
|-|-|
|image|image|

### DIFF
--- a/src/components/customers/CustomerItem.tsx
+++ b/src/components/customers/CustomerItem.tsx
@@ -71,7 +71,7 @@ export const CustomerItem = memo(({ rowId, customer }: CustomerItemProps) => {
           </NameBlock>
         </CustomerNameSection>
         <PlanInfosSection>
-          <MediumCell align="right">{activeSubscriptionCount}</MediumCell>
+          <Typography align="right">{activeSubscriptionCount}</Typography>
           <SmallCell align="right">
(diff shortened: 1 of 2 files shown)


src/components/customers/CustomerItem.tsx
```diff
@@ -152,7 +152,7 @@ const Item = styled(ListItemLink)`
 `
 
 const MediumCell = styled(Typography)`
-  width: 200px;
```
MorganeLecurieux, line 155: I wonder if we couldn't let the "subscription" column shrink to have more space in this one ?

MorganeLecurieux (approved):

src/components/customers/CustomerItem.tsx
ansmonjol: Instead of the customer name one ? yep sounds better so the customer name had a min width instead.

Gonna update

ansmonjol: Done ! I've removed the min-width of the middle column, so it only takes the place it needs to take and customer name can be fully displayed

PR #614, opened 2022-12-07
feat (coupons): handle reusable coupons

lovrocolic:
## Context

Handle reusable coupons

## Description

Currently, the same coupon can be applied multiple times on a customer. With new changes, we are able to control if coupon is reusable or not.

New checkbox is added on coupon form; default state is reusable = true


src/hooks/useCreateEditCoupon.ts
```diff
@@ -29,6 +29,7 @@ gql`
     name
     amountCurrency
     code
+    reusable
```
ansmonjol, line 32: Is that change needed ? Seems `reusable` is not used in here

ansmonjol (approved):
Just one comment otherwise LGTM !

src/hooks/useCreateEditCoupon.ts
lovrocolic: Hmm, If I remember correctly it was needed (this is edit coupon fragment) since reusable checkbox is also editable...

ansmonjol: ok probably an issue with the fragment usage in here, that was already done before you jumped into this file.

Will resolve, as a refactor can come later

PR #661, opened 2023-01-02
Premium warning dialog

ansmonjol:
## Context

Users can have different level of access to the app, leading to some features to not be available anymore

## Description

This PR adds the modal that will be displayed to users if they try to access to a premium feature

image


src/components/PremiumWarningDialog.tsx
```diff
@@ -0,0 +1,53 @@
...
+            <Button>{translate('text_63b3155768489ee342482f55')}</Button>
+          </LinkTo>
+        </>
+      )}
+    />
+  )
+})
+
+const BGIcon = styled.div`
+  width: ${theme.spacing(10)};
```
MorganeLecurieux, line 38: Hummm I don't think we should use spacing as width... 
Spacing are a specific system just for spacing. Width can be out of this system (and also let's say we change the system of spacing by putting it in base 2, this will f*** up the width based on it.
Just use the hardcoded value here

src/components/PremiumWarningDialog.tsx
```diff
@@ -0,0 +1,53 @@
...
+  height: ${theme.spacing(10)};
```
MorganeLecurieux, line 39: Same here

src/components/PremiumWarningDialog.tsx
```diff
@@ -0,0 +1,53 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-color: ${theme.palette.secondary[100]};
+  border-radius: ${theme.spacing(3)};
```
MorganeLecurieux, line 44: Same here

src/components/PremiumWarningDialog.tsx
```diff
@@ -0,0 +1,53 @@
...
+        </BGIcon>
+      }
+      title={translate('text_63b3155768489ee342482f4f')}
+      description={translate('text_63b3155768489ee342482f51')}
+      actions={({ closeDialog }) => (
+        <>
+          <Button variant="quaternary" onClick={closeDialog}>
+            {translate('text_62f50d26c989ab03196884ae')}
+          </Button>
+          <LinkTo href="mailto:xyz@yourapplicationdomain.com">
```
MorganeLecurieux, line 28: Don't forget to change this

src/components/designSystem/Dialog.tsx
```diff
@@ -13,34 +13,37 @@ import { useEffect } from 'react'
 import { theme, ButtonGroup } from '~/styles'
 
 import { Typography } from './Typography'
+import { Icon, IconName } from './Icon'
```
MorganeLecurieux, line 16: I'm wondering if this is something we will reuse or if we're changing a base component just to fit an edge case ?

MorganeLecurieux (changes requested):
Some changes needed :)

src/pages/CreatePlan.tsx
```diff
@@ -196,6 +198,7 @@ const CreatePlan = () => {
     formikProps.values.interval,
   ])
 
+  premiumWarningDialogRef.current?.openDialog()
```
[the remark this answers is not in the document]
ansmonjol: Damn, I push this unexpected change. Will remove

src/components/designSystem/Dialog.tsx
ansmonjol: Not sure how to nicely handle that tbh if I don't update the dialog api.

src/pages/CreatePlan.tsx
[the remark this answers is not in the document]
ansmonjol: Removed 🙏

src/components/PremiumWarningDialog.tsx
ansmonjol: Done

src/components/designSystem/Dialog.tsx
ansmonjol: Seen with product, icon is removed for now, we'll see later if we want to use icon on modals

MorganeLecurieux (approved):

PR #669, opened 2023-01-04
feat: display contact lago modal on premium feature access

ansmonjol:
## Context

We provide different access to users depending on their license.

## Description

If you're premium you can access them all, otherwise you're displayed a modal asking you to contact Lago

Note the customer timezone input is disabled if user is not premium, no modal is displayed

image


src/hooks/customer/useIsPremiumUser.tsx
```diff
@@ -0,0 +1,21 @@
...
+gql`
+  query getCurrentUserPremium {
+    currentUser {
+      id
+      premium
+    }
+  }
+`
+
+export const useIsPremiumUser = () => {
```
MorganeLecurieux, line 14: This is a bit too specific IMO, we should have a hook that allow to access the currentUser

src/components/customers/AddCustomerDrawer.tsx
```diff
@@ -178,12 +180,14 @@ export const AddCustomerDrawer = forwardRef<DrawerRef, AddCustomerDrawerProps>(
               formikProps={formikProps}
             />
 
-            {/* <ComboBoxField
```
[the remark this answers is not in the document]
ansmonjol: wrote it in the PR description, this part of the app only needs to be disabled so we're good ! 

image

src/hooks/customer/useIsPremiumUser.tsx
ansmonjol: As discussed on Slack, we'll keep it like this until a bigger refactor planned on both orga and customer data management

MorganeLecurieux (approved):

PR #691, opened 2023-01-11
Search for Plan, AddOns, Coupons and customers

ansmonjol:
## Roadmap Task

👉 https://github.com/getlago/lago/issues/156

## Context

It’s hard for our user to find a specific objects when there’s more than 10 objects in a list view.

## Description

This PR adds a search bar in those pages
- Plans list
- AddOns list
- Coupons list
- Customers list

Better review commit by commit


src/pages/PlansList.tsx
```diff
@@ -43,62 +47,109 @@ const PlansList = () => {
   })
   let index = -1
 
+  useEffect(() => {
```
MorganeLecurieux, line 50: So considering how it's used and that you repeat the logic in all the places with search, I think we need to architecture it differently : 

`TextInputForSearch` doesn't really make sense, maybe just as a UI component and then it should accept the prop "onChange"
Also the `TextInput` already has a `localValue` so I don't think you need an additional one here 🤔

Then all the logic regarding the query should be in `useDebouncedSearch` that should be used where the search is used : 
It should accept a `searchQuery` and `loading` and : 
- Launch the query on first load
- Return a `loading` state that is "debounced" has you're doing it here

ansmonjol: Changes have been pushed, thanks for your message

MorganeLecurieux (approved):

PR #711, opened 2023-01-17
Search cust and invoices

ansmonjol:
## Roadmap Task

👉 https://github.com/getlago/lago/issues/156

## Context

It’s hard for our user to find a specific objects when there’s more than 10 objects in a list view.

## Description

This PR adds a search bar in those pages

- Invoices list
- Draft invoice list
- Coupons list
- Subscriptions in combobox
- Addon in combobox
- Coupons in combobox
- Credit notes in customer details view

Better review commit by commit


src/components/customers/CustomerCreditNotesList.tsx
```diff
@@ -7,6 +7,7 @@ import {
   CurrencyEnum,
   useGetCustomerCreditNotesQuery,
```
MorganeLecurieux, line 8: Oups ^^

src/components/customers/CustomerCreditNotesList.tsx
```diff
@@ -48,10 +57,12 @@ export const CustomerCreditNotesList = ({
...
-  const { data, loading, error, fetchMore } = useGetCustomerCreditNotesQuery({
-    variables: { customerId, limit: 20 },
-    skip: !customerId,
-  })
+  const [getCreditNotes, { data, loading, error, fetchMore, variables }] =
+    useGetCustomerCreditNotesLazyQuery({
+      variables: { customerId, limit: 20 },
+    })
+  const { debouncedSearch, isSearchLoading } = useDebouncedSearch(getCreditNotes, loading)
+  const isLoading = isSearchLoading || loading
```
MorganeLecurieux, line 65: But... I thought we wanted to debounce the `loading`, if you do that it's not debounced anymore 🤔

src/components/customers/CustomerInvoicesTab.tsx
```diff
@@ -54,106 +63,121 @@ export const CustomerInvoicesTab = ({ customerId, customerTimezone }: CustomerIn
...
+      limit: 20,
+      status: InvoiceStatusTypeEnum.Finalized,
+    },
+    notifyOnNetworkStatusChange: true,
   })
+  const { debouncedSearch, isSearchLoading } = useDebouncedSearch(
+    getFinalizedInvoices,
+    loadingFinalized
+  )
+  const isFinalizedLoading = isSearchLoading || loadingFinalized
```
MorganeLecurieux, line 87: Same... If you do this there is no point to debounce the loading for the search

src/components/customers/creditNotes/CreditNotesList.tsx
```diff
@@ -264,7 +289,25 @@ CreditNotesList.displayName = 'CreditNotesList'
...
+  box-shadow: ${theme.shadows[7]};
+
+  div {
+    width: 100%;
+    margin-right: 70px;
+    display: flex;
+    justify-content: flex-end;
+
+    span:first-child {
+      margin-right: 16px;
```
MorganeLecurieux, line 304: use the `theme.spacing` here (same for the margin right before, 70 shouldn't be possible 🤔

src/core/apolloClient/cache.ts
```diff
@@ -67,7 +67,7 @@ export const cache = new InMemoryCache({
           keyArgs: ['invoiceId'],
         },
         customerInvoices: {
```
MorganeLecurieux, line 69: I think you forgot to add that for the other search queries no ?

src/pages/CustomerDraftInvoicesList.tsx
```diff
@@ -85,20 +108,31 @@ const CustomerDraftInvoicesList = () => {
...
+          </Typography>
+          <SearchInput
+            onChange={debouncedSearch}
+            placeholder={translate('text_63c6cac5c1fc58028d0235d9')}
+          />
+        </ListHeader>
+
         <CustomerInvoicesList
-          loading={loading}
+          isLoading={!!isSearchLoading || loading}
```
MorganeLecurieux, line 134: Same question for the loading...

src/pages/InvoicesList.tsx
```diff
@@ -94,6 +103,8 @@ const InvoicesList = () => {
       }),
     },
   })
+  const { debouncedSearch, isSearchLoading } = useDebouncedSearch(getInvoices, loading)
+  const isLoading = isSearchLoading || loading
```
MorganeLecurieux, line 107: 🤷‍♀️

MorganeLecurieux (changes requested):

src/components/customers/CustomerCreditNotesList.tsx
ansmonjol: If the search is loading, then `loading` will never be read (`true || false`)

I kept this for the loading on scroll as we don't want to debounce it. Does it make sense ?

src/core/apolloClient/cache.ts
ansmonjol: I had no issue with others resources

Basically, I mostly want to cache results no matter the search result.

`customerInvoices` case was a bit different: There is 2 query containing ont data that is stored in the same object, but can contains a different value: metadata.totalCount.

The first one is used for the load on scroll, and needs to change depending on the number of results prompted on the screen.
The second one is used to get the total number of invoices (draft in this case), and this value should not change depending on the `searchTerm`.

Another approach could have been to store the initial `totalCount` value in a local state but I preferred the cache approach, avoiding "complexity" in the component itself.

Happy to discuss about it !

src/components/customers/creditNotes/CreditNotesList.tsx
ansmonjol: I messed up with the skeleton here. Took a different approach but still had to have a big value on the margin-right.

Used theme spacing and code is simpler. Let me know what you think

src/components/customers/CustomerCreditNotesList.tsx
ansmonjol: Removed 😬

src/core/apolloClient/cache.ts
MorganeLecurieux: Ok got it
I think is not the best option then as then the caching is ont really happening for this query (each time you have a new search, the items are stored with a new key, apollo does not get the ones already fetched)

So the 2 options are :
- The one you mentionned
- Having a separated query with no search that gets the totalCount only

MorganeLecurieux (approved):

src/components/customers/CustomerCreditNotesList.tsx
ansmonjol: Logic updated

PR #726, opened 2023-01-20
test: reduce Drawer transition duration during tests

ansmonjol:
## Context

Out test suite is taking a lot of time.

Trying to find way to reduce it a bit

## Description

This PR removed the timer when Drawer component opens for it to be instant


ansmonjol:
Not even noticeable, will check on node_modules size first

PR #811, opened 2023-02-13
chore: type countryCodes object

ansmonjol:
Better with types!


src/countryCodes.ts
```diff
@@ -0,0 +1,257 @@
+import { CountryCode } from './generated/graphql'
```
MorganeLecurieux, line 1: Not a big fan to put that at the root
Why did you remove it from public ?
Alternative could be in `core`

src/components/settings/EditOrganizationInformationsDialog.tsx
```diff
@@ -14,20 +14,19 @@ import {
...
 
-const countryData: { value: string; label: string }[] = Object.keys(CountryCodes).map(
-  (countryKey) => {
-    return {
-      value: countryKey,
-      // @ts-ignore
-      label: CountryCodes[countryKey],
-    }
+const countryData: { value: string; label: string }[] = (
+  Object.keys(CountryCodes) as Array<keyof typeof CountryCodes>
```
MorganeLecurieux, line 23: This is used several times, maybe we can export it form the countryCode file
Also, the the type is just an array of strings : `Object.keys(CountryCodes) as string[]`

src/components/settings/EditOrganizationInformationsDialog.tsx
```diff
@@ -14,20 +14,19 @@ import {
...
 const FILE_MAX_SIZE = 800000
 
-const countryData: { value: string; label: string }[] = Object.keys(CountryCodes).map(
-  (countryKey) => {
-    return {
-      value: countryKey,
-      // @ts-ignore
-      label: CountryCodes[countryKey],
-    }
+const countryData: { value: string; label: string }[] = (
```
MorganeLecurieux, line 22: This should be the `BasicComboBoxData` type

ansmonjol: Indeed will look into it

src/countryCodes.ts
ansmonjol: Core sound like a good alternative,

updated!

src/components/settings/EditOrganizationInformationsDialog.tsx
ansmonjol: Done! thanks for pointing this

MorganeLecurieux (approved):

PR #816, opened 2023-02-14
Update invoice payment status

ansmonjol:
## Context

We can update an invoice via API and change its `payment_status` value.

We want to port this feature in the UI

## Description

This PR uses the new mutation that allow updating an invoice paymentStatus.

---

Related to https://github.com/getlago/lago-api/pull/830


src/components/invoices/EditInvoicePaymentStatusDialog.tsx
```diff
@@ -0,0 +1,143 @@
...
+          formikProps.resetForm()
+          formikProps.validateForm()
+        }}
+        actions={({ closeDialog }) => (
+          <>
+            <Button
+              variant="quaternary"
+              onClick={() => {
+                closeDialog()
+                formikProps.resetForm()
```
MorganeLecurieux, line 107: no need for `validateForm` here ?

MorganeLecurieux (approved):

src/components/invoices/EditInvoicePaymentStatusDialog.tsx
ansmonjol: Nope, modal is init with a given prop, so no matter what was there before. Even tho we remove the data and close, value is set onOpen

We also do not show errors bellow the input and it's not possible to submit without value

PR #851, opened 2023-02-22
misc(test): reduce animation during tests

ansmonjol:
## Context

The world need blazingly-fast® tests

## Description

This PR removes the animation duration in tests


MorganeLecurieux:
I’m not really convinced by the approche...
We will never really want to tests the transitions, so we’d better disable all of them at once instead of doing it one by one
Last step of [this article](https://glebbahmutov.com/blog/css-animations/) is an option

PR #862, opened 2023-02-28
feat(customer): Add, edit, remove and display metadata

ansmonjol:
## Context

We only have a limited number of fields on the customer object. You cannot write new fields for a customer.

## Description

This PR allow to add/edit/delete metadata on customers


src/components/customers/AddCustomerDrawer.tsx
```diff
@@ -1,10 +1,19 @@
-import { forwardRef, useEffect, useState, RefObject } from 'react'
-import styled from 'styled-components'
+import React, { forwardRef, useEffect, useState, RefObject } from 'react'
+import styled, { css } from 'styled-components'
 import { useFormik } from 'formik'
-import { object, string } from 'yup'
+import { array, object, string } from 'yup'
+import _, { FieldWithPossiblyUndefined } from 'lodash'
```
MorganeLecurieux, line 5: Never import `_`
It prevents the tree shaking and will import the whole lodash lib while you only need one or 2 functions from it

ansmonjol: arf I thought it would do it anyway 😓 Thanks, will update

src/components/customers/AddCustomerDrawer.tsx
```diff
@@ -67,14 +82,73 @@ export const AddCustomerDrawer = forwardRef<DrawerRef, AddCustomerDrawerProps>(
...
+                    ?.map((m: { key: string }) => m.key)
+                    ?.lastIndexOf(value)
+                ) {
+                  return createError({
+                    path,
+                    message: MetadataErrorsEnum.uniqueness,
+                  })
+                }
+
+                if (value.length > 20) {
```
MorganeLecurieux, line 114: Use a constant maybe for 20, that way you can give it a understandable name

src/components/customers/AddCustomerDrawer.tsx
```diff
@@ -67,14 +82,73 @@ export const AddCustomerDrawer = forwardRef<DrawerRef, AddCustomerDrawerProps>(
...
+                  })
+                }
+
+                return true
+              },
+            }),
+            value: string().test({
+              test: (value, { createError, path }) => {
+                if (!value) return false
+                if (value.length > 40) {
```
MorganeLecurieux, line 127: Use a constant maybe for 40, that way you can give it a understandable name (because right now we don't know what it stands for)

src/components/customers/AddCustomerDrawer.tsx
```diff
@@ -67,14 +82,73 @@ export const AddCustomerDrawer = forwardRef<DrawerRef, AddCustomerDrawerProps>(
...
+            }),
+          })
+        ),
       }),
       validateOnMount: true,
       enableReinitialize: true,
       onSubmit: async (values, formikBag) => {
+        values?.metadata?.forEach(
+          (
+            element: CustomerMetadataInput & {
```
MorganeLecurieux, line 145: You're using this type several time, I guess you could make a reusable type out of it

src/components/customers/AddCustomerDrawer.tsx
```diff
@@ -67,14 +82,73 @@ export const AddCustomerDrawer = forwardRef<DrawerRef, AddCustomerDrawerProps>(
...
+              localId?: string
+            }
+          ) => {
+            delete element.localId
+          }
+        )
         const answer = await onSave(values)
```
MorganeLecurieux, line 152: By doing this you modify the formikProps.values datas without "informing" which can lead to side effects (especially if the mutation does not work, your list of elements might be f****ed)
To avoid that, you could just : 
```tsx
const answer = await onSave(values?.metadata?.map(({ lodalId, ...metadata }) => metadata))
```
That way you don't modify the initial object

src/components/customers/AddCustomerDrawer.tsx
```diff
@@ -348,7 +422,152 @@ export const AddCustomerDrawer = forwardRef<DrawerRef, AddCustomerDrawerProps>(
...
+                      (
+                        m: CustomerMetadataInput & {
+                          localId?: string
+                        },
+                        i
+                      ) => {
+                        const metadataItemKeyError: FieldWithPossiblyUndefined<
+                          string | undefined,
+                          `${number}`
+                        > = _.get(formikProps.errors, `metadata.${i}.key`)
```
MorganeLecurieux, line 464: Replace all those by just `get`

src/components/customers/AddCustomerDrawer.tsx
```diff
@@ -348,7 +422,152 @@ export const AddCustomerDrawer = forwardRef<DrawerRef, AddCustomerDrawerProps>(
...
+                        )
+                      }
+                    )}
+                  </MetadataGrid>
+                </div>
+              )}
+              <Button
+                startIcon="plus"
+                variant="quaternary"
+                disabled={(formikProps?.values?.metadata?.length || 0) >= 5}
```
MorganeLecurieux, line 554: Same, avoid to use hard coded values, naming them allow to have context

src/components/customers/AddCustomerDrawer.tsx
```diff
@@ -376,12 +595,33 @@ const HelperText = styled(Typography)`
...
 
   > *:not(:last-child) {
     margin-bottom: ${theme.spacing(6)};
   }
 `
 
+const MetadataGrid = styled.div<{ $isHeader?: boolean }>`
+  display: grid;
+  grid-template-columns: 200px 1fr 60px 24px;
+  gap: 24px 12px;
```
MorganeLecurieux, line 609: Gap are spacing, use `theme.spacing`

src/core/apolloClient/cache.ts
```diff
@@ -66,6 +66,9 @@ export const cache = new InMemoryCache({
         customerCreditNotes: {
           keyArgs: ['invoiceId'],
         },
+        customerMetadata: {
```
MorganeLecurieux, line 69: I'm confused... Why do you need to do that ?

ansmonjol: agree, did that cause of a cache warning asking to do so somehow.

Will try again to understand what could cause it

src/components/customers/AddCustomerDrawer.tsx
ansmonjol: Fixed

src/components/customers/AddCustomerDrawer.tsx
ansmonjol: Fixed

src/components/customers/AddCustomerDrawer.tsx
ansmonjol: Fixed

MorganeLecurieux (approved):

src/core/apolloClient/cache.ts
ansmonjol: Ok so here is what happens

## Context

We have an apollo warning when deleting metadata and saving a customer with those changes.

Reason is that apollo does not know how to reconcile this array of objects that have theyr own type.
Specially because when deleting the items from the UI, We delete them from the Formik values but not from the cache.
So when we udpate the customer, we have the current cached version of the customer containing metadata that does not exists anymore.

## Explorations

I tried to 
- manually evict deleted elements
- define a customer merge function for metadata in `cache.ts`
- remove the manual udpate cache method and try toi mix fragment definition to attempt to make apollo understand by itself which object to update

None of those worked. The interesting about the last item is that 
- The customerDetails and Customer types are somehow mixed in the app. We should only have one type imo
- This part of the app is pretty message in term of fragment definition. There is a wrong pattern where data is asked in parent and spread over all children, and also a hook is taking care or defining expected data on mutation result, but it looks to not match ideally with the fragments for the list / details of customer.

## Next steps

We need to
1.  clean types definition on BE side and reflect this change on FE
2. clear fragments definition and data fetching hierarchy in the app

I'll create a ticket for this on FE repo as in our Notion kanban and take care of both BE and FE 

Also, I'll resolve this conversation and keep the change as it is in this PR as it's the less ugly solution I would say

PR #912, opened 2023-03-20
chore: 👩🏽‍🎤 Allow variant and contextual translations

MorganeLecurieux:
- [x] Add `useContextualLocale` hook to allow change of a local only in one component
- [x] Update the existing translation hook to avoid rewriting code
- [x] Add script to import all the variants translations

🚨 As I can't import the locale config in the script, this has to be manually updated (ie if we add a new variant, we need to add that variant in the script)


ansmonjol (approved):
🔥

PR #938, opened 2023-04-07
Fix timezone

ansmonjol:
## Context

We have an issue on the timezone.
First, the timezone config set on login, was lost on page reload.
Second, we noticed we needed to send the subscription date in UTC to the backend .

## Description

The config is now done during the user info fetch on each page boot.

Also, we pre-format the subscription date before sending it to BE


src/hooks/customer/useAddSubscription.tsx
```diff
@@ -194,7 +194,7 @@ export const useAddSubscription: UseAddSubscription = ({
           input: {
             customerId,
             ...(!existingSubscription
-              ? { subscriptionAt: subsDate }
+              ? { subscriptionAt: DateTime.fromISO(subsDate).startOf('day').toUTC().toISO() }
```
jdenquin, line 197: is the `toISO()` needed since you already do a `fromISO()` ?

jdenquin (approved):

PR #1148, opened 2023-08-22
feat (localization): Add FE translations for italian

lovrocolic:
## Context

Add possibility to set italian as default locale for documemts. Also, add italian translations for email preview section


ansmonjol (approved):

src/components/customerPortal/PortalInvoiceListItem.tsx
```diff
@@ -49,6 +49,7 @@ interface PortalInvoiceListItemProps {
   invoice: PortalInvoiceListItemFragment
   translate: Function
   className?: string
+  documentLocale: string
```
ansmonjol, line 52: Should be of type LocaleEnum

src/pages/auth/PortalInit.tsx
```diff
@@ -43,11 +43,11 @@ const PortalInit = () => {
     nextFetchPolicy: 'network-only',
     skip: !isPortalAuthenticated || !token,
   })
-  const { translateWithContextualLocal: translate } = useContextualLocale(
-    (data?.customerPortalUser?.billingConfiguration?.documentLocale as Locale) ||
-      (data?.customerPortalOrganization?.billingConfiguration?.documentLocale as Locale) ||
-      'en'
-  )
+
+  let documentLocale = (data?.customerPortalUser?.billingConfiguration?.documentLocale as Locale) ||
```
ansmonjol: You can set it as a const. It won't be reassigned

src/components/customerPortal/PortalInvoiceListItem.tsx
ansmonjol: if it complain about the 'en' being a non valid value, we need to update the LocaleEnum definition to be an object `as const`

Let me know if that's happening

lovrocolic: changed

PR #1210, opened 2023-09-12
feat(bm): add weighted sum aggregation type

ansmonjol:
👉 https://getlago.canny.io/feature-requests/p/support-gauge-model-aggregation-type

## Context

We want to add a new BM type in the app

## Description

This PR does add the weighted sum BM type to the app


vincent-pochet (approved):
:rocket:

PR #1372, opened 2023-12-28
Prevent Credit Note details from always displaying hardcoded USD

edwardmp:
A hardcoded USD currency is shown because accountCurrency is never fetched using GraphQL, this PR fixes this bug.

## Context

Even if my credit note is in EUR, USD is shown for the item

## Description

The issue is that the `amountCurrency` is used in the code to render the item, however it isn't fetched using GraphQL, defaulting to USD.
See screenshot.

Screenshot 2023-12-28 at 18 16 44

### DIFF
--- a/src/pages/CreditNoteDetails.tsx
+++ b/src/pages/CreditNoteDetails.tsx
@@ -85,6 +85,7 @@ gql`
       }
       items {
         amountCents
+        amountCurrency
         fee {
           id
           amountCents


ansmonjol:
Hey @edwardmp, thanks a lot for reporting this and fixing it!

I've re-opened a PR with your changes on our repo so CI can pass.

I've marked you as co-author of the commit of course.

https://github.com/getlago/lago-front/pull/1373

PR #1374, opened 2024-01-04
Show payment token in Customer Info

TronTram:
## What
- Show `payment token` in Customer Info
image


ansmonjol:
Hi @TronTram, thanks for your submission.
I have few remarks about this PR, so next one can be even better!

- I feel this PR does more than simply showing a payment token on the customer infos. Should it be about adding the Pinet PSP instead? May be worth adding more context in the description and reword title
- About this Pinet provider, I have not been able to find anything related. Can you point to a doc or something?
- I assume some BE work should be performed too. Are you working on it maybe? I can see you closed your initial PR.

In addition to this, we're currently working on [allowing multiple PSP connection](https://getlago.canny.io/feature-requests/p/connect-more-than-1-payment-provider-account-to-lago).
This work will lead to a complete refactor of the pages you updated/created in this PR and all related Backend parts that may need to be updated for your needs.

Regarding this last point and the list above, I'll close this PR.

If you feel any needs about adding a new PSP, I suggest you to create a new feature request in [our Canny board](https://getlago.canny.io/feature-requests).

PR #1413, opened 2024-02-09
misc: add missing KWD currency

lovrocolic:
## Context

This PR adds missing KWD currency

### DIFF
--- a/src/generated/graphql.tsx
+++ b/src/generated/graphql.tsx
@@ -1276,6 +1276,8 @@ export enum CurrencyEnum {
   Kmf = 'KMF',
   /** South Korean Won */
   Krw = 'KRW',
+  /** Kuwaiti Dinar */
+  Kwd = 'KWD',
   /** Cayman Islands Dollar */
   Kyd = 'KYD',
   /** Kazakhstani Tenge */


jdenquin (approved):

ansmonjol:
This currency can have up to 3 decimal precision 

You should add it to the list of 3  precision currencies in src/core/serializers/serializeAmount.ts

ansmonjol (changes requested):
This currency can have up to 3 decimal precision

You should add it to the list of 3 precision currencies in src/core/serializers/serializeAmount.ts

michaelponrajah:
I think it's already mentioned [here](https://github.com/getlago/lago-front/blob/c8091cace463c13460cc888e8798e01a96902281/src/core/serializers/serializeAmount.ts#L4)

michaelponrajah (approved):

lovrocolic:
> I think `KWD` is already mentioned [here](https://github.com/getlago/lago-front/blob/c8091cace463c13460cc888e8798e01a96902281/src/core/serializers/serializeAmount.ts#L4), line 4

yep, KWD has been included in the list from the start

PR #1452, opened 2024-03-30
Fix back navigation after close tax add/edit form

josemoyab:
## Context

After closing the create/edit tax form in the settings, I noticed that the back button wasn't functioning as intended. Instead of navigating the user back to the Home page, it was directing them back to the tax form.

## Description

This PR modifies the back redirection URLs on the settings layout to exclude the create and update routes, ensuring that the back button now correctly redirects users to the Home page.


ansmonjol (approved):
Very good catch! Thanks for the PR 🙏

ansmonjol:
CI is not properly configured yet, I'll pack your 2 PR and mark you as co-author

PR #1473, opened 2024-04-25
chore: refactor orga types with new definition

ansmonjol:
Retrieved organizations returns all their attributes, no matter if they are fetched directly (get current organization) or alongside with another attribute (current user -> organization)

There is some situations where having such can expose too many informations on non-accessible data 

We decided to have a new type of organization, that allow to fetch any kind of attributes (for get current organization)
And then another one for cross data fetching (for current user -> organization)


keellyp (approved):
LGTM!

PR #1491, opened 2024-05-06
bug(dates) remove hardcoded dot in date format

ansmonjol:
Dates format almost all contain this hardcoded dot.

It lead to wring notation for May, cause it displays `May.`.

The plan is the following:
1. Update all formatting to remove this hardcoded dit (this PR)
2. Later, have a scaling task to centralise all formatting, improve the system and change the method to format. I would love to use `toLocaleString` with `DATE_MED` preset instead of `.toFormat` ([see doc](https://github.com/moment/luxon/blob/master/docs/formatting.md#presets))

Fixes DLVRY-295

### DIFF
--- a/src/components/TimezoneDate.tsx
+++ b/src/components/TimezoneDate.tsx
@@ -23,7 +23,7 @@ interface TimezoneDateProps {
 }
 
 export const TimezoneDate = ({
-  mainDateFormat = 'LLL. dd, yyyy',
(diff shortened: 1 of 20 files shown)


keellyp (approved):

ansmonjol:
More complicated that this.

We should have such logic. Luxon does not seem to have this implemented. Gonna have a deeper look

PR #1496, opened 2024-05-10
feat: handle custom aggregation

keellyp:
## Context

This PR should handle custom aggregation type


src/core/serializers/serializePlanInput.ts
```diff
@@ -112,6 +112,9 @@ const serializeProperties = (properties: Properties, chargeModel: ChargeModelEnu
             : undefined,
         }
       : { perTransactionMinAmount: undefined, perTransactionMaxAmount: undefined }),
+    ...(chargeModel === ChargeModelEnum.Custom
+      ? { customProperties: JSON.parse(properties?.customProperties) }
```
ansmonjol, line 116: There is an associated test for this file. Maybe it's worth checking this transform?

ansmonjol (approved):

src/components/designSystem/Drawer.tsx
```diff
@@ -119,7 +129,8 @@ const Header = styled.div`
   }
 `
 
-const Content = styled.div`
+const Content = styled.div<{ $fullContentHeight?: boolean }>`
+  height: ${({ $fullContentHeight }) => ($fullContentHeight ? '100%' : ' ')};
```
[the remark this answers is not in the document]
keellyp: @ansmonjol Not always, when the content inside the drawer is longer than the viewport height, we don't want to set the height to 100% otherwise it's breaking the container padding-bottom

src/core/serializers/serializePlanInput.ts
keellyp: @ansmonjol Oups my bad! See 833e8ef6333f53bb2e29d2d8077b608091050e83

PR #1498, opened 2024-05-14
chore(inspectTranslations) add flag to replace TODO: copy

ansmonjol:
While implementing a feature, we often take care of the copy management as a last step.

We often write them in our templates as "TODO: Final Copy wording here", so prefixed with TODO: 

This PR allow to flag the inspectTranslations script to find and replace those copy, and replace them with the value present in base.json, for it's corresponding wording.

NOTE: This script is not fully working yet, and only replace one file.


keellyp (approved):
So far it looks good to me, but can you provide an example on how to run the script because i think we don't have the same workflow when updating our keys 😄

PR #1511, opened 2024-05-21
misc: improve replacement regex in script

ansmonjol:
It does
- Make sure the whole string is replaced with the expected key
- Make sure the string to replace is exactly the same as the one targeted (case sentitive)


keellyp (approved):
nice!!

PR #1513, opened 2024-05-21
feat: add okta login method

keellyp:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/{{FEATURE_SLUG}}

## Context

This PR should handle Okta SSO for login and invitation mode


src/layouts/Settings.tsx
```diff
@@ -110,6 +111,10 @@ const Settings = () => {
                 link: INTEGRATIONS_ROUTE,
                 hidden: !hasPermissions(['organizationIntegrationsView']),
               },
+              {
+                title: translate('text_664c732c264d7eed1c74fd96'),
+                link: AUTHENTICATION_ROUTE,
```
keellyp, line 116: Since you merged rbac logic, should I add permission check here too @ansmonjol ?

ansmonjol: Good point, you probably need to check with product. I'm not sure such permissions exists today. Feel it's supposed to exist tho.

src/pages/auth/LoginOkta.tsx
```diff
@@ -0,0 +1,175 @@
...
+      context: { silentErrorCodes: [LagoApiError.UnprocessableEntity] },
+      fetchPolicy: 'network-only',
+    })
+
+  useEffect(() => {
+    if (lagoErrorCode) {
+      setErrorAlert(lagoErrorCode as LagoApiError)
+
+      // Remove the error code from the URL, so it disappears on page reload
+      history.replaceState({}, '', window.location.pathname)
```
ansmonjol, line 57: Just observation here. This code is also used in Google SSO code but while writing it I felt it's a bit of a hack and the url could contain other informations that are important for the user experience (got one case iun mind with future SalesForce iFrame usage)

Anyway, if you feel this should be changed or handled differently feel free to do so! Let's discuss about it together if you want 🙌

ansmonjol (approved):

src/pages/auth/LoginOkta.tsx
keellyp: Yep I agree but I thought that clean the url before submitting again would prevent "dirty" url it case if the user is encountering errors multiple times. 
Sometimes it can break with the redirect_uri so I've no strong opinion on this one 🤔

ansmonjol (approved):

PR #1527, opened 2024-05-31
refactor: use builder to write our code snippets

keellyp:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/{{FEATURE_SLUG}}

## Context

While working on screens where we need to update the code snippets, it can sometimes be hard to make sure the code rendered is valid with all the conditions that applies to our forms. 

## Description

This PR should create a new snippet builder tool that would help us write our code snippets. 
This is still WIP but lmk you think!

### DIFF
--- a/src/components/CodeSnippet.tsx
+++ b/src/components/CodeSnippet.tsx
@@ -63,7 +63,7 @@ export const CodeSnippet = memo(
                 variant="secondary"
                 startIcon="duplicate"
                 onClick={() => {
-                  copyToClipboard(code)
+                  copyToClipboard(code, { ignoreComment: true })
 
                   addToast({
                     severity: 'info',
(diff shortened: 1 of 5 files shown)


src/components/CodeSnippet.tsx
```diff
@@ -63,7 +63,7 @@ export const CodeSnippet = memo(
                 variant="secondary"
                 startIcon="duplicate"
                 onClick={() => {
-                  copyToClipboard(code)
+                  copyToClipboard(code, true)
```
ansmonjol, line 66: Suggestion: maybe make the the option explicit.
Either name all attributes or make the last one part of an option object so we can
`copyToClipboard(code, { ignoreComment: true })` or something like that

ansmonjol (approved):

PR #1546, opened 2024-06-12
feat: update status component

keellyp:
## Context

To prepare my next PR, I'm updating the status component

## Description

`<Status/>` component should be responsible of **its own style and label** depending in `type: danger | success | warning | default...`. Label should be restricted by its type, meaning that we should not be able to set this for instance: 

```javascript
<Status type="success" label="failed" /> // it will throw a ts error
``` 

| Before | After |
|--------|-------|
| Capture d’écran 2024-06-12 à 10 23 59 | Capture d’écran 2024-06-12 à 10 07 44 |

QA-app: https://design-status-migration-app.staging.getlago.com/


ansmonjol (approved):
Yey let's gooo

PR #1557, opened 2024-06-19
feat: add xero integration and map items

ansmonjol:
This feature mainly add the New Zero integration and allows to map temps in it's settings

It also includes some fixes and improvements on others integrations.

Note the Xero integration is hidden behind a FE only FF

There is a remaining TODO, i'm working on it but it's UI related issue that is not visible for the moment anyway


keellyp (approved):
LGTM 👍

PR #1575, opened 2024-06-30
misc(currency) add IRR

AnonC0DER:
Adding IRR (Iranian Rial) currency

### DIFF
--- a/src/generated/graphql.tsx
+++ b/src/generated/graphql.tsx
@@ -1475,6 +1475,8 @@ export enum CurrencyEnum {
   Inr = 'INR',
   /** Icelandic Króna */
   Isk = 'ISK',
+  /** Iranina Rial */
+  Irr = 'IRR',
   /** Jamaican Dollar */
   Jmd = 'JMD',
   /** Jordanian Dinar */


ansmonjol:
Hey @AnonC0DER thanks for adding this new currency.

I made few search, but can you confirm this currency does have a 2 digit precision for the cents?
We have a particular handling for currencies with different digit precision and I would love to make sure we're ok for this one.

AnonC0DER:
Hello @ansmonjol, thank you for checking.

No, all amounts in IRR should be treated as whole numbers without any decimal places. 
I wasn't aware there is a specific area for handling these types of currencies. 
Could you direct me to it so I can modify the PR?

ansmonjol:
Ok I did check those infos

[This list](https://developer.cybersource.com/docs/cybs/en-us/currency-codes/reference/all/na/currency-codes/currency-codes.html) that present IRR with 2 decimal, but I agree I could find other resources that says 0 decimal too 😕 

But most important, [this is the specification of the lib we use on Backend side](https://github.com/RubyMoney/money/blob/37e183a883b1629dee30570e3482c35e591530be/config/currency_iso.json#L1018) for all currencies, and IRR shows with `"subunit_to_unit": 100,`, just like `EUR`.
JPY that have no decimal is defined as `"subunit_to_unit": 1,` for example.

Even if that is a specification mistake, we'll have to follow this package definition and handle it as a 2 decimal currency, otherwise it would add buggy behaviour in the app for you (sending amount as 0 decimal but handled as 2 decimal on Backend)

In such case, for 2 decimal currencies, there is no other change to add to your PR! 
FTR, [this is the place we define currencies precision](https://github.com/getlago/lago-front/blob/main/src/core/serializers/serializeAmount.ts#L5). Very custom and specific implem, don't feel bad not finding it 😄 

Ok, sorry for this long text but we should be all good to merge your PR haha.

Probably the PR's CI will fail because of GH token error, but in such case I'll re-open it and tag you as co-author if you don't mind.

AnonC0DER:
Great! 
Thank you for the quick response.

ansmonjol:
Re-opened [here](https://github.com/getlago/lago-front/pull/1577)

Thanks for your contribution @AnonC0DER

PR #1587, opened 2024-07-03
feat: add succeededAt date

keellyp:
## Roadmap Task

LAGO-77

## Description

Add succeededAt date in invoice and credit note form

QA: https://recurring-app.staging.getlago.com/login ✅


src/components/invoices/details/InvoiceDetailsTable.tsx
```diff
@@ -178,10 +182,16 @@ export const InvoiceDetailsTable = memo(
                   displayName={
                     invoice.invoiceType === InvoiceTypeEnum.AddOn
                       ? translate('text_6388baa2e514213fed583611', { name: fee.itemName })
-                      : invoice.invoiceType === InvoiceTypeEnum.OneOff
+                      : invoice.invoiceType === InvoiceTypeEnum.OneOff ||
+                          invoice.invoiceType === InvoiceTypeEnum.AdvanceCharges
                         ? fee.invoiceDisplayName || fee.itemName
                         : translate('text_637ccf8133d2c9a7d11ce6e1')
                   }
+                  succeededDate={
```
ansmonjol, line 190: What about naming this attribute in a more generic way and use something like "description"?

It won't match much with "displayName" aboce, but probably we'll re-use this attribute to display something else than a date later

Nitpicking so feel free to resolve if you want!

ansmonjol (approved):

src/components/invoices/details/InvoiceDetailsTable.tsx
keellyp: @ansmonjol Actually the "description" in itself is computed here https://github.com/getlago/lago-front/blob/b50da2ac57fe0f370231952797491fe959b24e56/src/components/invoices/details/InvoiceDetailsTableBodyLine.tsx#L213-L217 with the subLabel being declared in the component itself rather than passed like a props from the parent. So I thought that `description` would not be exact actually 🤔

PR #1607, opened 2024-07-15
feat: release xero integration

ansmonjol:
This PR makes the xero integration visible in the UI, so customers can use it.

It keep the CN sync hidden tho, we wait for a fix before making it enabled


keellyp (approved):

ansmonjol:
Done here instead https://github.com/getlago/lago-front/pull/1621

PR #1610, opened 2024-07-16
misc: update the xero account fetch

ansmonjol:
Account fetch is now a mutation


src/components/settings/integrations/XeroIntegrationMapItemDialog.tsx
```diff
@@ -567,6 +543,12 @@ export const XeroIntegrationMapItemDialog = forwardRef<XeroIntegrationMapItemDia
           formikProps.resetForm()
           formikProps.validateForm()
         }}
+        onOpen={() => {
+          // Have to delay the ececution of the query, as the dialog props are not present immediatly after the dialog is opened
+          setTimeout(() => {
+            getXeroIntegrationItems()
+          }, 1)
```
keellyp, line 550: Using a asynchronous function does not work?

ansmonjol: This method is already an async one, it's from a lazy query builder.

I had a situation where on mount, this modal data (specially the data type) is not the one expected, as it's used by 2 different type here, and the type is used as param in the query.

I cannot really check for type presence as it can already be there but having a wrong value in the state.

The only workaround I found is this one: delay the execution on next clock tick, cause there the type have been updated once received in props and this callback stack item can now be executed with the correct context.

If you see a better approach let me know! I think I could trigger this in the `useImperativehandle` Method, the first one to have the correct data before updating the state
But at the same time I preferred it to keep not having anything related with data fetching
WDYT?

keellyp: Hum I see, I haven't an other solution in mind to propose at the moment but thanks for the context! Let's use this one!

keellyp (approved):

PR #1627, opened 2024-07-26
feat: new table component

keellyp:
## Roadmap Task

LAGO-145

## Context

New table component that is not used in the app for the moment

### DIFF
--- a/src/core/utils/responsiveProps.ts
+++ b/src/core/utils/responsiveProps.ts
@@ -2,7 +2,7 @@ import { css } from 'styled-components'
 
 import { theme } from '~/styles'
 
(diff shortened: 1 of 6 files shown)


src/core/utils/responsiveProps.ts
```diff
@@ -0,0 +1,41 @@
+import { css } from 'styled-components'
+
+import { theme } from '~/styles'
+
+type Breakpoint = 'xs' | 'sm' | 'md' | 'lg' | 'xl'
```
ansmonjol, line 5: Can it be `keyof theme.breakpoints.values` ?

ansmonjol (approved):

PR #1631, opened 2024-07-31
feat: add settings for no invoice top-up

keellyp:
## Roadmap Task

LAGO-107

## Description

- Add switch option in top-up dialog
- Add switch option in wallet settings


src/components/wallets/TopupWalletDialog.tsx
```diff
@@ -82,6 +84,7 @@ export const TopupWalletDialog = forwardRef<DialogRef, TopupWalletDialogProps>(
               walletId: wallet.id,
               grantedCredits: grantedCredits === '' ? '0' : String(grantedCredits),
               paidCredits: paidCredits === '' ? '0' : String(paidCredits),
+              invoiceRequiresSuccessfulPayment: formikProps.values.invoiceRequiresSuccessfulPayment,
```
ansmonjol, line 87: Why not getting it from the props but directly from formik object?

ansmonjol (approved):
Added one comment otherwise lgtm!

src/components/wallets/TopupWalletDialog.tsx
keellyp: Nice catch!! Fixing it

PR #1640, opened 2024-08-06
feat: export invoices

ansmonjol:
We built a filter logic in the invoice page. We can now export this filtered list

It's a mutation that will take the current filters, and request an export to the BE using them.

They should receive the mail on the currently connected user's email

Fixes LAGO-143


src/components/invoices/ExportInvoicesDialog.tsx
```diff
@@ -0,0 +1,216 @@
...
+    })
+
+    const formikProps = useFormik<Omit<CreateDataExportsInvoicesInput, 'filters'>>({
+      initialValues: {
+        format: DataExportFormatTypeEnum.Csv,
+        resourceType: ExportTypeEnum.Invoices,
+      },
+      validationSchema: object().shape({
+        format: string().required(''),
+        resourceType: string().required(''),
```
keellyp, line 64: I think you can do something like this to be more restrictive: 

```suggestion
        resourceType: string().valid(...Object.values(ExportTypeEnum)).required('')
```

src/styles/muiTheme.ts
```diff
@@ -57,6 +57,7 @@ export const theme = createTheme({
     MuiPopper: {
       styleOverrides: {
         root: {
+          zIndex: '1800 !important',
```
keellyp, line 60: Are you sure that it will not cause the same issue as we encountered yesterday? 👀

keellyp (approved):
Few nitpick comments but lgtm overall!

src/styles/muiTheme.ts
ansmonjol: Good catch, it would! I removed it in another PR and rebase bru he's it back, will remove

src/components/invoices/ExportInvoicesDialog.tsx
ansmonjol: `resourceType: string().oneOf(Object.values(ExportTypeEnum)).required(''),` did the trick.

Good idea for filters too, I'll change separately thanks!

PR #1642, opened 2024-08-07
fix: use paymentDisputeLostAt instead of paymentDisputeLosable

keellyp:
## Roadmap Task

issue-413

## Context

we used the wrong attribute to determine if the payment was disputed or not

## Description

use paymentDisputeLostAt instead of paymentDisputeLosable

### DIFF
--- a/src/components/customerPortal/PortalInvoicesList.tsx
+++ b/src/components/customerPortal/PortalInvoicesList.tsx
@@ -35,7 +35,7 @@ gql`
     id
     paymentStatus
     paymentOverdue
-    paymentDisputeLosable
+    paymentDisputeLostAt
     number
     issuingDate
     totalAmountCents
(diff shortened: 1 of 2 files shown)


src/components/customerPortal/PortalInvoicesList.tsx
```diff
@@ -76,11 +76,12 @@ gql`
...
   paymentStatus,
   paymentOverdue,
-  paymentDisputeLosable,
+  paymentDisputeLostAt,
 }: {
   paymentStatus: InvoicePaymentStatusTypeEnum
   paymentOverdue: boolean
-  paymentDisputeLosable: boolean
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  paymentDisputeLostAt: any | null
```
ansmonjol, line 84: Isn't it string | null?

ansmonjol: Oh yeah the type from the generated is any 😢

ansmonjol (approved):

src/components/customerPortal/PortalInvoicesList.tsx
keellyp: I guess typing as string works!

PR #1651, opened 2024-08-12
feat: request payment page

keellyp:
## Roadmap Task

LAGO-139

## Context

Related to overdue balance feature

## Description

This PR should add a new route under customer router : `/request-overdue-payment`


src/pages/CustomerRequestOverduePayment/components/EmailPreview.tsx
```diff
@@ -0,0 +1,246 @@
...
+`
+
+const Text = styled(Typography)`
+  font-family: Helvetica, sans-serif;
+  font-size: 16px;
+  font-weight: 400;
+  line-height: 24px;
+`
+
+const CompanyName = styled(Text)`
```
ansmonjol, line 219: Why not using the Typography component here?

src/pages/CustomerRequestOverduePayment/validateEmails.ts
```diff
@@ -0,0 +1,22 @@
+import { string } from 'yup'
+
+// Matches a comma surrounded by optional whitespace
+const SEPARATOR_REGEX = /\s*,\s*/
```
ansmonjol, line 4: Is there a specific case that requires this regex?

At first I thought performing a `split(',').map(e => e.trim())` would be enough but maybe I was wrong!
And as this would needs an extra filter to check for the email presence we could replace all of it by a reduce, that will trim and push value only of present.

Anyway, don't need to change this logic if it works! Just curiosity

src/pages/CustomerRequestOverduePayment/validateEmails.ts
```diff
@@ -0,0 +1,22 @@
...
+  for (let email of emailArray) {
+    if (!string().email().isValidSync(email)) {
+      return false
+    }
+  }
+  return true
+}
+
+export const serializeEmails = (emails: string): string => {
+  return emails.replace(' ', '')
```
ansmonjol, line 21: Shouldn't that be a `replaceAll` ?

src/pages/CustomerRequestOverduePayment/components/EmailPreview.tsx
keellyp: `Text` is a custom component from `Typography` defined on line 212

src/pages/CustomerRequestOverduePayment/validateEmails.ts
keellyp: Nice catch!!

src/pages/CustomerRequestOverduePayment/validateEmails.ts
keellyp: I've totally overcomplicated my function. This solution works well haha. Updating it!

src/pages/CustomerRequestOverduePayment/components/EmailPreview.tsx
ansmonjol: Sure, but what's the advantage of using Text over out internal Typography?

keellyp: Cause for the email preview we don't use the same font family...

PR #1667, opened 2024-08-14
fix: make take browser compatible

keellyp:
## Context

I've realized that we had some css issues when using Safari and Firefox. 
This PR aims to fix them.

## Description

- Remove unused `isFullWidth` props
- Add wrapping div with `width: 0` and `min-width: 100%` to fix to overflow scroll issue 

| **Before** | **After** |
| ---------- | ---------- |
| Capture d’écran 2024-08-14 à 17 29 07 | Capture d’écran 2024-08-14 à 17 25 17 | 

- Use `@supports (animation-timeline: scroll(inline))` when it's possible, otherwise don't render border-right on action column.

| **Before** | **After** |
| ---------- | ---------- |
| Capture d’écran 2024-08-14 à 17 29 40 | Capture d’écran 2024-08-14 à 17 25 14 | 

- Use `transform: 'translateZ(0)'` to fix Safari issue with the scrollbar when the action column is sticky

| **Before** | **After** |
| ---------- | ---------- |
| Capture d’écran 2024-08-14 à 18 03 55 | Capture d’écran 2024-08-14 à 18 11 57 | 

- Use `undefined` instead of `unset` to remove issue when hovering rows

| **Before** | **After** |
| ---------- | ---------- |
| Capture d’écran 2024-08-14 à 18 21 50 | Capture d’écran 2024-08-14 à 18 22 09 |

### DIFF
--- a/src/components/designSystem/Table/Table.tsx
+++ b/src/components/designSystem/Table/Table.tsx
@@ -214,7 +214,7 @@ export const Table = <T extends DataItem>({
       $rowSize={rowSize}
     >
       <TableHead>
-        <TableRow>
+          <tr>
           <>
             {columns.map((column, i) => (
               <TableCell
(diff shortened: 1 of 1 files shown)


ansmonjol (approved):
Awesome!

PR #1696, opened 2024-08-27
feat: progressive billing (plan + subscription)

keellyp:
## Roadmap Task

LAGO-222

## Context

Progressive billing feature

## Description

Add the possibility to add progressive billing in plan

Note: I recommend reviewing by files changed instead of commit by commit as I've refactored things during development


src/components/invoices/details/InvoiceDetailsTableFooter.tsx
```diff
@@ -51,6 +52,32 @@ export const InvoiceDetailsTableFooter = memo(
...
+              <tr>
+                <td></td>
+                <td colSpan={colSpan}>
+                  <Typography variant="bodyHl" color="grey600">
+                    {translate('text_1724936123802q74m2xxak3o')}
+                  </Typography>
+                </td>
+                <td>
+                  <Typography variant="body" color="success600">
+                    {' '}
```
ansmonjol, line 65: Is that intended?

src/components/plans/details/PlanDetailsAdvancedSettingsSection.tsx
```diff
@@ -79,8 +79,7 @@ const PlanDetailsAdvancedSettingsSection = ({
                             currency: currency,
                           },
                         ),
-                        threshold.thresholdDisplayName ||
-                          `${translate('text_1724179887723pt34ivcecdt')} ${i + 1}`,
+                        threshold.thresholdDisplayName || '',
```
ansmonjol, line 82: I thought we should display a placeholder name here, did that changed in the meantime?

src/core/serializers/serializePlanInput.ts
```diff
@@ -144,6 +146,24 @@ export const serializePlanInput = (values: PlanFormInput) => {
             taxes: undefined,
           }
         : {},
+    usageThresholds: [
+      ...(nonRecurringUsageThresholds ?? []).map((threshold) => ({
+        ...threshold,
+        thresholdDisplayName: threshold.thresholdDisplayName ?? null,
```
ansmonjol, line 152: hum, If the user tries to edit the name, and then remove the value (at the same moment), the value will be an empty string if I'm correct
And there you will then send this empty string to the BE.
How would that be then on fetch? Will we receive `null` or `''`?
If it's null, maybe we should use `||` instead WDYT?

src/components/plans/details/PlanDetailsAdvancedSettingsSection.tsx
keellyp: Ye it did (cf. https://linear.app/getlago/issue/LAGO-256/%5Bqa-8%5D-remove-predefined-name-when-threshold-name-is-empty)

src/core/serializers/serializePlanInput.ts
```diff
@@ -144,6 +146,24 @@ export const serializePlanInput = (values: PlanFormInput) => {
...
+        amountCents: Number(serializeAmount(threshold.amountCents, values.amountCurrency)),
+      })),
+      ...(recurringUsageThreshold
+        ? [
+            {
+              ...recurringUsageThreshold,
+              thresholdDisplayName: recurringUsageThreshold.thresholdDisplayName ?? null,
```
ansmonjol, line 159: same as above here

src/components/invoices/details/InvoiceDetailsTableFooter.tsx
keellyp: Nop, fixing it!

src/components/plans/details/PlanDetailsAdvancedSettingsSection.tsx
ansmonjol: ok!

ansmonjol (approved):

PR #1712, opened 2024-09-04
misc: make sure redirection is not lost after outside navigation

ansmonjol:
Fixes ISSUE-422

## Context

If a user follows a link to Lago and is not redirected, they are asked to login.
Once logged in successfully, their reach their expected destination.

However, if a user logs in using a third party (like google SSO), this redirection logic breaks.
The reason is that during the thirdparty login process, they are redirected out the app and the "memory" of the destination page is lost.

## Description

This PR addresses this situation and bring a fix using the browser's local storage.

When ever a user is 
- kicked out the app after following a link (needs to login)
- Logs out the app (couldn't bypass this, this event looks just like a kick out)

they now "register" their last visited page (or destination in the first case) into their local storage.

This data is not lost when they use a thirdparty login (unless they use private navigation probably)

--

This is the best approach I found regarding those criteria
- Speed of implementation
- Number of files to change
- Complexity

It's not ideal, not the best, but probably the fastest approach regarding our current implemented logic.

I opened a ticket to refactor this approach, but mostly to change our route event and history system.

### DIFF
--- a/src/core/apolloClient/cacheUtils.ts
+++ b/src/core/apolloClient/cacheUtils.ts
@@ -39,6 +39,10 @@ export const setItemFromLS = (key: string, value: unknown) => {
   return localStorage.setItem(key, stringify)
 }
 
+export const removeItemFromLS = (key: string) => {
+  return localStorage.removeItem(key)
+}
+
 // --------------------- Auth utils ---------------------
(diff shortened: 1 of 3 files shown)


keellyp (approved):
LGTM! As you said, not ideal but it's a nice workaround for the moment

PR #1715, opened 2024-09-04
misc: change default size for tables to 48

ansmonjol:


ansmonjol:
Update have been done in another PR

PR #1719, opened 2024-09-05
chore: enable back cypress

keellyp:
## Context

This PR should enable back E2E tests with Cypress latest version.

## Description

- [x] Valid tests
- [x] Enable CI 
- [x] Enable screenshots and videos uploads for failing tests

Fixes [ISSUE-397](https://linear.app/getlago/issue/ISSUE-397/[project]-enable-back-e2e-tests)


cypress/e2e/00-auth/t0-signup.cy.ts
```diff
@@ -5,21 +5,18 @@ describe('Sign up page test', () => {
...
-  )
+  it('should create a new user and redirect to the home after signup', () => {
+    cy.visit('sign-up')
+
+    cy.get('input[name="organizationName"]').type('Company name')
+    cy.get('input[name="email"]').type(userEmail)
+    cy.get('input[name="password"]').type(userPassword)
+    cy.get('[data-test="submit-button"]').click()
+
+    // TODO: Why is it redirecting to customer page and why is it forbidden?
```
ansmonjol, line 16: Should this be removed? I guess what you describe here is related to permissions, but it should not happen with a newly created account 🤔

cypress/support/reusableConstants.ts
```diff
@@ -1,6 +1,7 @@
 export const planWithChargesName = `a plan with charges`
+export const planWithChargeCode = `a_plan_with_chargesnew_code_plan_with_charge`
```
ansmonjol, line 2: Any chance it's a wrong paste in the string and the value should be shorter?

Seems we pasted next to an existing value trying to replace it, imo 😬

ansmonjol (approved):
Few minor comments, approved in the meantime so I don't block you!

cypress/support/reusableConstants.ts
keellyp: nice catch

PR #1723, opened 2024-09-09
feat(InlineBarsChart): add new tooltip on hover API

ansmonjol:
Fixes LAGO-300

This PR does add a new API to this component: `tooltipsData`

Once passed, it can show a tooltip on hover, with a specific given message

`tooltipsData` keeps the same data shape as the `data` in this component, which is [dictated by Rechart](https://recharts.org/en-US/api/Bar#data)


stephenlago99 (approved):
Looks good!

PR #1729, opened 2024-09-10
feat: intl support

keellyp:
## Context

Add a new flag in translation script to allow international keys

## Description

```bash
// Only update english language
yarn translations:add {number}

// Updates all files in translation folder with the same key
yarn translations:add {number} --intl
```


ansmonjol (approved):
Love it, thanks for taking care of this

PR #1734, opened 2024-09-12
feat: Differentiate between company and individuals

stephenlago99:
Since the users now have the ability to add a `customerType` to Customers, a new dropdown was added in the Customer creation modal.

Customers also have a fisrtname & lastname now, so we've changed most of the places that render the customer name to also take them into account.


src/components/customers/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  if (firstname || lastname) {
+    return getInitials(`${firstname} ${lastname}`.trim())
+  }
+
+  return '-'
+}
+
+export const computeCustomerName = (
```
ansmonjol, line 22: IIRC the backend was planning to return us this value in a new customer's attribute called `displayName`
See their dive-in: https://www.notion.so/getlago/BE-Differentiate-companies-vs-individuals-for-Lago-customers-abc18d1e050c48b5b86b1a2a3c172565?pvs=4#7b3983ab3632430da86f91b5a25508a8

brunomiguelpinto: yes displayName should be available on the graphql

stephenlago99: Yup, added a new commit. Used `displayName` instead of computing the name on the frontend. I've also kept the function, so if in the future we would need to change anything related to customer names, it would be way easier.

keellyp (approved):

PR #1767, opened 2024-09-27
feat: new hubspot integration

keellyp:
## Context

Hubspot integration

## Description

This PR adds hubspot integration in the settings


tailwind.config.ts
```diff
@@ -111,7 +111,17 @@ const config = {
       xl: '0px 16px 24px 0px rgba(25, 33, 46, 0.10)',
     },
     fontFamily: {
-      sans: ['Inter, Arial , Verdana , Helvetica , sans-serif'],
+      sans: ['Inter, Arial, Verdana, Helvetica, sans-serif'],
+    },
+    container: {
```
ansmonjol, line 116: Why having this as a config rather than having twice the same classname in the files?

Asking as I'm afraid those kind of global classes may not be re-used that much and the day we refactor this component or delete their usage, the config may not be updated.

Not hard pushing, your choice

ansmonjol (approved):
Just one comment, feel free to handle it or not

PR #1768, opened 2024-09-30
chore: some Tailwind component migration

ansmonjol:


src/components/billableMetrics/BillableMetricItem.tsx
```diff
@@ -64,21 +63,23 @@ export const BillableMetricItem = memo(
...
-            </NameBlock>
-          </BillableMetricName>
-          <CellSmall align="right">{formatTimeOrgaTZ(createdAt)}</CellSmall>
-          <ButtonMock />
+            </div>
+          </div>
+          <Typography className="mr-6 w-28" align="right">
+            {formatTimeOrgaTZ(createdAt)}
+          </Typography>
+          <div className="w-10" />
```
keellyp, line 82: Why do we need this empty div?

keellyp (approved):

src/components/billableMetrics/BillableMetricItem.tsx
ansmonjol: The button is placed with an absolute position based on the parent relative, parent being the whole line container.

To make the items well alligned we need to "reserve" some place for the button to be placed at the end of the line, pushing away other line infor (name, code, etc)

ansmonjol: This is the old way we have to create those buttons, a event prevent default would do the job 🤷

PR #1784, opened 2024-10-02
feat: Customer Portal v2

stephenlago99:
### Context

This pull request contains the new version of the Customer Portal.

Goals for the new version of the Customer Portal:

- Make it accessible to everyone (goal solved in another PR that was already merged)
- Expose subscription and usage data (UsagePage, UsageSection)
- Expose customer information and edit customer information (Customer information section already existed, so it was only polished and a page with a form was created)
- Expose wallet information and top-up workflow (WalletPage, WalletSection)

It also includes fixes after the first round of QA testing (commits starting with `LAGO-`).

### Description

Since this implementation required a refactoring of the existing CustomerPortal, this couldn't be split in multiple merge requests. There are new components that are used in multiple places, and separating it would have made the development process harder.

#### New components

Since the old Customer portal was missing some sections, some new components were developed for this:
- Wallet section / Wallet page
- Usage section / Usage page
- Customer information page (for editing the customer information)

To aid code share between them, some helper components were also created:
- PageTitle
- SectionContainer
- SectionError
- SectionLoading
- SectionTitle
- TextButton

The design also facilitated the creation of reusable components, since design bits were re-used between sections and pages.

##### SearchInput

A new `className` param was added to aid tailwind styling.

##### Subscription usage table / graph

`src/components/subscriptions/SubscriptionCurrentUsageTable.tsx`

`src/components/subscriptions/SubscriptionUsageLifetimeGraph.tsx`

Since the Usage part required us to display the usage table / graph, the existing components were refactored and de-coupled from the graphql queries.

##### New routes

The Customer Portal now includes a way to navigate between pages (in the previous implementation, we didn't have separate pages).

Besides the existing `CUSTOMER_PORTAL_ROUTE`, some new routes were added:

For general pages:

`export const CUSTOMER_PORTAL_ROUTE_PAGE = '/customer-portal/:token/:page'`

For pages with items (like the usage section):

`export const CUSTOMER_PORTAL_ROUTE_PAGE_ITEMID = '/customer-portal/:token/:page/:itemId'`


src/components/customerPortal/PortalCustomerInfos.tsx
```diff
@@ -22,198 +32,209 @@ gql`
...
+  addressB?: CustomerAddressInput | null
+}) =>
+  addressA &&
+  addressB &&
+  addressA.addressLine1 === addressB.addressLine1 &&
+  addressA.addressLine2 === addressB.addressLine2 &&
+  addressA.state === addressB.state &&
+  addressA.country === addressB.country &&
+  addressA.city === addressB.city &&
+  addressA.zipcode === addressB.zipcode
```
keellyp, line 81: I'd probably opt for a utility function that is more agnostic about the address keys, in case we decide to add more fields in the future.

Something like this might work—what do you think?

```suggestion
  return (Object.keys(addressA) as (keyof CustomerAddressInput)[]).every(
    (key) => addressA[key] === addressB[key]
  );
```

src/components/customerPortal/common/CustomerPortalSidebar.tsx
```diff
@@ -0,0 +1,90 @@
...
+  margin-right: ${theme.spacing(3)};
+  border-radius: 8px;
+
+  > img {
+    height: 100%;
+    width: 100%;
+    object-fit: cover;
+    border-radius: inherit;
+  }
+`
```
keellyp, line 88: Can we use tailwind instead?

src/components/customerPortal/common/SectionContainer.tsx
```diff
@@ -0,0 +1,7 @@
+type SectionContainerProps = {
+  children: React.ReactNode
+}
+
+const SectionContainer = ({ children }: SectionContainerProps) => <div>{children}</div>
```
keellyp, line 5: Why do we need this container?

src/components/customerPortal/customerInformation/CustomerInformationPage.tsx
```diff
@@ -0,0 +1,330 @@
...
+    onCompleted(res) {
+      if (res) {
+        onSuccess?.()
+      }
+    },
+  })
+
+  const formikProps = useFormik<Partial<UpdateCustomerInput>>({
+    initialValues: {
+      // @ts-ignore
```
keellyp, line 57: What did you ignored?

src/components/customerPortal/PortalInvoicesList.tsx
```diff
@@ -68,10 +76,50 @@ gql`
     }
   }
 
+  query getCustomerPortalInvoicesCollection($expireCache: Boolean) {
+    customerPortalInvoiceCollections(expireCache: $expireCache) {
```
ansmonjol, line 80: Broader topic, I'm wondering if this endpoint could be "melted" into the `customerPortalInvoices` used above.
That would require some BE rework and probably be skipped, but here is my path
- I feel we query the same resource in both endpoints
- `expireCache` seems not to be used here
- `amountCents` is probably the total amount cents of all invoices added. Wondering if that could be added to the metadata
- `invoicesCount` is probably the `customerPortalInvoices`' `metadata.totalCount`, or as we han have filters in `customerPortalInvoices` it should be another metadata
- the `currency` got there should be the customer's one, fetched bellow in `getCustomerPortalUserCurrency` query
Then this query could be replaced.

That's form my understanding I may miss something tho, let me know!

src/components/customerPortal/PortalInvoicesList.tsx
```diff
@@ -68,10 +76,50 @@ gql`
...
+    customerPortalOverdueBalances(expireCache: $expireCache) {
+      collection {
+        amountCents
+        currency
+        lagoInvoiceIds
+      }
+    }
+  }
+
+  query getCustomerPortalUserCurrency {
```
ansmonjol, line 99: About that `customerPortalUser` query. 
I can see it's performed twice in your PR, and probably more in the current code or next to come. (just spotted also in `src/components/customerPortal/PortalCustomerInfos.tsx`)

That sound like a data we may want to share across all the customer portal and that won't necessarily change once we load the page.
Did you consider putting that customer fetch logic in a reusable hook, method or any place that could allow fetching those informations only once and re-use it elsewhere?

src/components/customerPortal/usage/UsageSubscriptionItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+          {subscription?.plan?.interval && (
+            <p className="text-base font-normal lowercase text-grey-700">
+              {translate(planIntervalLabel(subscription.plan.interval))}
+            </p>
+          )}
+        </div>
+      )}
+
+      <p className="text-sm font-normal leading-6 text-grey-600">
+        {translate('text_1728377747178bfroky3hn30')}{' '}
```
ansmonjol, line 98: Probably this `{' '}` and the 4 other occurrences can be removed

src/components/customerPortal/PortalCustomerInfos.tsx
```diff
@@ -22,198 +32,209 @@ gql`
...
+        zipcode
+      }
     }
   }
 `
 
-interface PortalCustomerInfosProps {
-  translate: Function
+const FieldTitle = ({ title }: { title: string }) => (
+  <p className="text-sm font-normal leading-6 text-grey-600">{title}</p>
```
ansmonjol, line 49: Why not using the Typography component we have in our design system?
IIRC there is an attribut you could pass to make it use a `p` and not a `div` as it does by default

Maybe it raise the question of the usage of such typo component as we have those tailwind classes now, but I would rather make a large migration later than starting to introduce some different implementation WDYT?

src/components/customerPortal/usage/UsageSubscriptionItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { useInternationalization } from '~/hooks/core/useInternationalization'
+
+type UsageSubscriptionItemProps = {
+  subscription?: SubscriptionForPortalUsageFragment | null
+  applicableTimezone?: TimezoneEnum | null
+  loading?: boolean
+  children?: React.ReactNode
+}
+
+const planIntervalLabel = (interval?: string | null) => {
```
ansmonjol, line 15: Nitpicking, seems interval will always be present as you have a condition above invocation.

Also, maybe a lookup object could help being more concise as you did for `INVOICE_TYPE_TRANSLATION_MAP`!

src/core/router/CustomerPortalRoutes.tsx
```diff
@@ -8,10 +8,20 @@ const PortalInit = lazyLoad(
 
 // ----------- Routes -----------
 export const CUSTOMER_PORTAL_ROUTE = '/customer-portal/:token'
+export const CUSTOMER_PORTAL_ROUTE_PAGE = '/customer-portal/:token/:page'
```
ansmonjol, line 11: Why do we prefer having a list of page and trigger their render in the DOM with `if page === 'xxx' => render xxx` rather than having a route definition for each pages?

src/components/customerPortal/PortalInvoicesList.tsx
stephenlago99: Good suggestion, but currently each query has some logic embedded below (the loading / error states etc), so I would keep this as a future improvement (they fetch different data anyways, but your solution would indeed reduce the number of calls).

src/components/customerPortal/PortalCustomerInfos.tsx
stephenlago99: Good suggestion, but I would also keep this as a future improvement. Currently <Typography> brings lots of functionality behind the scenes and since all the typography in the portal was mapped to the Figma design, I'm afraid that this might include un-wanted changes.

src/components/customerPortal/PortalInvoicesList.tsx
ansmonjol: ok! Yeah I think it would be a real benefit reducing some api calls as we're adding more and more around

src/components/customerPortal/PortalInvoicesList.tsx
ansmonjol: We decided to tackle this later

src/core/router/CustomerPortalRoutes.tsx
stephenlago99: Done, I have removed the extra routes, included an Outlet and did it through react-router

src/components/customerPortal/usage/UsageSubscriptionItem.tsx
stephenlago99: Good point, removed the optional part and created a map.

src/components/customerPortal/usage/UsageSubscriptionItem.tsx
stephenlago99: Replaced some with "&nbsp;" (the structure there was too tricky to insert in a translation) and inserted 2 of them in translations directly.

src/components/customerPortal/usage/UsageSection.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  const subscription = portalUsageData?.customerPortalSubscriptions?.collection
+
+  const applicableTimezone =
+    portalUsageData?.customerPortalSubscriptions?.collection?.[0]?.customer?.applicableTimezone
+
+  const isLoading = portalUsageLoading
+  const isError = portalUsageError
+
+  const refreshSection = () => {
```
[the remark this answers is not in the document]
stephenlago99: Removed.

src/components/customerPortal/PortalInvoicesList.tsx
```diff
@@ -139,48 +230,91 @@ const PortalInvoicesList = ({ translate, documentLocale }: PortalCustomerInvoice
...
 
-  const { debouncedSearch, isLoading } = useDebouncedSearch(getInvoices, loading)
+  const { debouncedSearch, isLoading: searchIsLoading } = useDebouncedSearch(getInvoices, loading)
   const { metadata, collection } = data?.customerPortalInvoices || {}
   const hasSearchTerm = !!variables?.searchTerm
   const hasNoInvoices = !loading && !error && !metadata?.totalCount && !hasSearchTerm
 
+  const { currentPage = 0, totalPages = 0 } = metadata || {}
+
+  const refreshSection = () => {
```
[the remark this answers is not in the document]
stephenlago99: Removed.

To give some additional context on why this was introduced in the first place: if we pass the `refetch` from Apollo directly to an onClick handler, it results in a 'cannot convert circular structure' error (https://stackoverflow.com/questions/41668454/circular-error-on-apollo-client-when-trying-to-refetch). Not sure why, but it would be interesting to debug.

src/components/customerPortal/customerInformation/CustomerInformationPage.tsx
stephenlago99: Removed.

src/components/customerPortal/common/SectionContainer.tsx
stephenlago99: Removed.

src/components/customerPortal/common/CustomerPortalSidebar.tsx
stephenlago99: Removed all styled components and used tailwind.

src/components/customerPortal/PortalCustomerInfos.tsx
stephenlago99: Good point, made it like this, I just had to type out the keys, because I was passing `customerPortalUser` directly.

src/components/customerPortal/PortalInvoicesList.tsx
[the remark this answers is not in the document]
ansmonjol: hum interesting, and might be in their side or I don't get it.
Yeah ok both would have worked then, i guess

src/components/customerPortal/PortalCustomerInfos.tsx
ansmonjol: Awesome

src/core/router/CustomerPortalRoutes.tsx
ansmonjol: Yeah I think it's way better this way! 🙌

translations/base.json
```diff
@@ -2476,5 +2466,48 @@
...
+  "text_1728470529877eb3qoinwqa5": "Balance updated on {{date}}",
+  "text_1728472381019nimbf36pk6t": "Type",
+  "text_1728472697691k6k2e9m5ibb": "Subscription",
+  "text_1728472697691x8zhuzkp74r": "Credits",
+  "text_1728472697691t126b808cm9": "One off invoice",
+  "text_1728472697691y39tdxgyrcg": "Advance charges",
+  "text_1728472697691pv1eqroobte": "Progressive billing",
+  "text_1728497501618g7qbdad97r5": "Type",
+  "text_1728498418253nyv3qmz9k5k": "Top-Up Wallet",
+  "text_1728546284339z3fs0oqdejs": "Something went wrong, token is invalid or has expired"
```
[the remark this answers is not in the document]
stephenlago99: Done, translated all the labels in all the languages.

ansmonjol (approved):
Thank for handling all our comments! 🙌

keellyp (approved):
LGTM so far! Huge PR congrats for the work!

PR #1795, opened 2024-10-09
Misc: some more TW migration

ansmonjol:


src/components/designSystem/graphs/InlineBarsChart.tsx
```diff
@@ -1,8 +1,8 @@
 import { useEffect, useState } from 'react'
 import { Bar, BarChart, BarProps, ResponsiveContainer, XAxis, YAxis } from 'recharts'
-import styled from 'styled-components'
 
-import { theme } from '~/styles'
+import { ChartWrapper } from '~/components/layouts/Charts'
+import { palette } from '~/styles'
```
keellyp, line 5: this file is referencing the palette used by MUI theme, i think it's fine but i wonder if we shouldn't stick with tailwind palette instead and have only one truth source 🤔

keellyp (approved):

src/components/designSystem/graphs/InlineBarsChart.tsx
ansmonjol: There is a class for that indeed! Removed

PR #1803, opened 2024-10-17
misc: migrate Icon component to use Tailwind

ansmonjol:
Main takeaway, we now point directly to the tag svg instead of custom `svg-icon` class.
Should be pretty save and looking around seems to be OK.

Didn't wanted to create a TW utility to point to a specific element.
We'll udpate the MUI config later


src/components/designSystem/Icon/Icon.tsx
```diff
@@ -24,49 +22,46 @@ enum IconAnimationEnum {
...
+}
 interface IconProps {
   name: IconName
-  size?: 'small' | 'medium' | 'large'
+  size?: keyof typeof IconSizeEnum
   color?: IconColor
   className?: string
   animation?: keyof typeof IconAnimationEnum
   onClick?: () => {} | void | Promise<void>
 }
```
keellyp, line 37: Instead creating new types, you can directly extract types from cva like this:
```suggestion
import { cva, VariantProps } from 'class-variance-authority'

export type IconVariantProps = VariantProps<typeof iconStyles>;

interface IconProps extends IconVariantProps {
  name: IconName
  className?: string
  onClick?: () => {} | void | Promise<void>
}
```

See https://cva.style/docs/getting-started/typescript

src/components/designSystem/Icon/Icon.tsx
```diff
@@ -79,53 +74,13 @@ export const Icon = ({
...
-  height: ${(props: { $size: keyof typeof IconSizeEnum }) => IconSizeEnum[props.$size]};
-  color: ${(props) => props.$color};
-  cursor: ${({ $canClick }) => ($canClick ? 'pointer' : 'initial')};
-
-  &.icon-animation--${IconAnimationEnum.spin} {
-    animation: spin 1s linear infinite;
-  }
-
-  &.icon-animation--${IconAnimationEnum.pulse} {
-    animation: pulse 1.5s ease-in-out 0.5s infinite;
```
keellyp, line 107: It seems that pulse animation from tailwind doesn't use the same easing and delays: https://tailwindcss.com/docs/animation#pulse. Maybe we have to update our config

keellyp (approved):

src/components/designSystem/Icon/Icon.tsx
ansmonjol: yep, thought no one would notice haha

We don't use this API today so I decided to keep with the default Tailwind config 😬

src/components/designSystem/Icon/Icon.tsx
ansmonjol: Sick! love it, updated the code

PR #1805, opened 2024-10-18
misc: update all filters related components to use tailwind

ansmonjol:
Main takeaway: I've updated the API of `<QuickFilters>` and `<Filters>` API as some attributes were not used at all. It helped making the implementation easier and if needed we can update them later to add that back

### DIFF
--- a/src/components/designSystem/Filters/ActiveFiltersList.tsx
+++ b/src/components/designSystem/Filters/ActiveFiltersList.tsx
@@ -1,10 +1,8 @@
 import { Typography } from '@mui/material'
 import { useMemo } from 'react'
 import { useSearchParams } from 'react-router-dom'
-import styled from 'styled-components'
 
 import { useInternationalization } from '~/hooks/core/useInternationalization'
-import { theme } from '~/styles'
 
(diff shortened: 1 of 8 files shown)


src/components/designSystem/Filters/ActiveFiltersList.tsx
```diff
@@ -53,22 +51,15 @@ export const ActiveFiltersList = ({ filters }: ActiveFiltersListProps) => {
   return (
     <>
       {activeFilters.map(({ label, value }, index) => (
-        <ActiveFilterChip key={`active-filter-${index}`}>
+        <div
+          className="box-border flex h-8 items-center rounded-[100px] bg-grey-100 px-3 py-0"
```
keellyp, line 55: I don't think box-border is useful here as it's applied by default by tailwind everywhere

keellyp (approved):

src/components/designSystem/Filters/ActiveFiltersList.tsx
ansmonjol: indeed! Removed them in all filters components and pushed

PR #1814, opened 2024-10-22
fix: remove duplicated keys in international translations

keellyp:
## Context

Clean translation file.

## Description

We had duplicated keys in the international files. This PR should clean them.


stephenlago99 (approved):
Looks good! 👏

PR #1815, opened 2024-10-22
feat: preview email in dunning creation form

keellyp:
## Context

This is related to automatic dunning initiative

## Description

It should implement the preview email in the create dunning page.
We use the same dunning email as for the manual payment feature but with fake data.

It can be safely merge as this is not visible in the app directly but only through `/settings/dunnings/create` URL

Fixes LAGO-403


ansmonjol (approved):
Need to address warning and linter errors but LGTM!

PR #1825, opened 2024-10-24
feat: Custom expressions for billing metrics

stephenlago99:
## Context

This pull request contains the implementations for:

- A new drawer for editing Custom expressions for Billable metrics.
- A new section inside billable metrics that allows you to switch between Unique field / Custom expression

The changes were done based on the frontend dive-in: `[FE] [Dive-in] Billable metric expressions`.

## Description

👉 `CustomExpressionDrawer`:
- Has two editors, enables the user to type & verify & save an expression
- The expression verification uses a new `lago-expression` package

👉 `JsonEditor`:
- Refactored to support passing new text modes (this is useful for not validating expressions as JSON). The thought process behind this change can be found in the dive-in.
- Refactored to support some new properties (custom errors, custom validate)

👉 `Drawer`:
- Refactored to support passing a classname to the StickyBottomBar (in order to be able to pass a z-index, as the editors had a z-index of 4)

👉 `lago-expression`:
- Currently it is embedded in our application, since the npm release process seemed trickier for a wasm package. When it will be released, we will swap it and remove it from our project.


src/pages/CreateBillableMetric.tsx
```diff
@@ -150,6 +166,15 @@ const CreateBillableMetric = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [formikProps.values.aggregationType, formikProps.values.fieldName])
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
stephenlago99: Good point! I'll move the logic inside formik.

src/pages/CreateBillableMetric.tsx
```diff
@@ -136,6 +144,14 @@ const CreateBillableMetric = () => {
     !!formikProps.initialValues.description,
   )
 
+  const [aggregateOnTab, setAggregateOnTab] = useState<'field' | 'custom-expression'>('field')
```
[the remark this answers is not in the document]
stephenlago99: ✅

src/pages/CreateBillableMetric.tsx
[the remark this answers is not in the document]
stephenlago99: ✅

src/lago-expression/expression_js.d.ts
```diff
@@ -0,0 +1,19 @@
+/* tslint:disable */
+/* eslint-disable */
```
ansmonjol, line 2: Are those disabling auto-generated or you added them manually? Just curious if we can inline them at the correct place if only one line is concerned, of if it's a global failure

stephenlago99: I didn't add anything by hand in the `lago-expression` folder. It's all generated by `wasm-pack`.

src/pages/CreateBillableMetric.tsx
```diff
@@ -129,13 +156,33 @@ const CreateBillableMetric = () => {
     }),
     enableReinitialize: true,
     validateOnMount: true,
-    onSubmit: onSave,
+    onSubmit: (values) => {
+      if (values.aggregateOnTab === AggregateOnTab.CustomExpression) {
+        return onSave(_omit(values, ['aggregateOnTab']))
```
ansmonjol, line 161: Refactor suggestion: 
Have an approach "if this, then modify values" then have a simple return given the formated values at the end.
Why? To make is more predictable and easier to encapsulate in a serialiser method we can extract and save later if the code starts to have more complexity.
The would lead to something like
```
const formatedValues = _omit(values, ['aggregateOnTab'])
if (values.aggregateOnTab !== AggregateOnTab.CustomExpression)
  // and maybe add a comment about expression only being used for custom expressions
  formatedValues.expression = null

return onSave(formatedValues)
```


Just a suggestion let me know of that make sense!

webpack.common.js
```diff
@@ -36,5 +36,8 @@ module.exports = () => {
         SENTRY_DSN: JSON.stringify(process.env.SENTRY_DSN),
       }),
     ],
+    experiments: {
+      asyncWebAssembly: true,
```
ansmonjol, line 40: Have you tried on staging if that worked?

ansmonjol (approved):

src/pages/CreateBillableMetric.tsx
stephenlago99: Interesting!

First, in my opinion, the idea of comments describing code isn't optimal. The effects and the reasoning behind our code should be understandable without comments. If we rely on comments to help developers understand the flow, we introduce technical debt since those comments have to be maintained.

Secondly, mutating values in the middle of functions could lead to [unwanted side-effects](https://sapegin.me/blog/avoid-mutation). Your solution, in the context of the current function, doesn't seem troublesome since the function is small. But if we keep mutations like that as a pattern in our codebase, functions that are bigger could become hard-to-follow / hard-to-debug. Indeed, the `onSubmit` could benefit from some refactoring, but it should be done in an immutable way (extract different branches to functions). If you think we should do it now, we could, but I would say that it's a bit of an early optimization.

Thanks!

webpack.common.js
stephenlago99: Yes, the version deployed to staging worked.

src/lago-expression/expression_js.d.ts
ansmonjol: ok then!

src/pages/CreateBillableMetric.tsx
ansmonjol: up to you!

webpack.common.js
ansmonjol: niiice

PR #1830, opened 2024-10-28
misc: fix misleading copy

ansmonjol:
Fixes ISSUE-499


stephenlago99 (approved):
It makes way more sense!

PR #1896, opened 2024-12-02
(refactor): Refactor drawer to tailwind

stephenlago99:
## Context

As part of our ongoing journey to migrate from styled components to tailwind, the Drawer component also needed refactoring.

## Description

This was done on a commit-per-component basis in order to aid the code reviewer (so the reviewer can easily spot any obvious differences or missing styles).

Fixes LAGO-424

### DIFF
--- a/src/components/designSystem/Drawer.tsx
+++ b/src/components/designSystem/Drawer.tsx
@@ -71,8 +71,7 @@ export const Drawer = forwardRef<DrawerRef, DrawerProps>(
     return (
       <>
         {!!opener && cloneElement(opener, { onClick: () => setIsOpen((prev) => !prev) })}
-        <StyledDrawer
-          $hasStickyBottomBar={!!stickyBottomBar}
+        <MuiDrawer
           open={isOpen}
           anchor={anchor}
           elevation={4}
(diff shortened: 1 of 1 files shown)


src/components/designSystem/Drawer.tsx
```diff
@@ -130,11 +131,16 @@ export const Drawer = forwardRef<DrawerRef, DrawerProps>(
           </Content>
 
           {!!stickyBottomBar && (
-            <StickyBottomBar className={stickyBottomBarClassName}>
+            <div
+              className={tw(
+                'border-t-grey200 sticky bottom-0 box-border border-t border-solid bg-white p-4 text-right md:px-12 md:py-4',
```
ansmonjol, line 150: Not sure what classes causes the issue but the sticky footer renders with a wrong border color.

Almost certain `border-t-grey200` is not a valid class but i'm not sure why the linter does not complain.

WDYT of changing all border rules with a `shadow-t` one instead? That also sets the color

stephenlago99: Indeed, it should have been `border-t-gray-200`, but as you've suggested, I've removed all the `border` classes and replaced them with `shadow-t`. Looks better now, good catch!

image

ansmonjol (approved):

PR #1897, opened 2024-12-02
misc: make sure we control filter values displayed

ansmonjol:
## Context

We display some invoice status filter values that are internal states for Lago, hence not being relevant for customers

## Description

This PR make sure we only display what we want to, also passes on other filters to use proper translations for the filter's label

Fixes ISSUE-572


stephenlago99 (approved):
Looks good!

Just a thought I had while going through the changes, not sure if it would be worth it to extract the values in separate objects (like we do for translations). Maybe not in this case, since the enum values aren't reused.

ansmonjol:
> Looks good!
> 
> Just a thought I had while going through the changes, not sure if it would be worth it to extract the values in separate objects (like we do for translations). Maybe not in this case, since the enum values aren't reused.

We could indeed, most of them are a re-usage of existing ones but sometimes they are uses in title and not only in "usage values" so might be overkill here.
Always possible later!

PR #1902, opened 2024-12-03
feat: migrate code snippet component

keellyp:
## Context

Related to tailwind migration

## Description

I've decided to use .css file as we need a lot of customization in the snippet component (to handle different language). Let met know what you think!


src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    useEffect(() => {
+      if (codeRef && codeRef.current) {
+        Prism.highlightElement(codeRef.current)
+      }
+    })
+
+    return (
+      <div className={tw('code-snippet', 'relative h-full', className)}>
+        {loading ? null : (
```
stephenlago99, line 48: What I've noticed is that if we try to swap similar logic to:

```
{!loading && (
  ...
)}

{loading && (
...
)}
```

we get less indentation & the code starts to look less a pyramid. It also makes it easier to debug since following a couple of `?` can get tricky.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,96 @@
...
+    code,
+    language = 'javascript',
+    canCopy = true,
+    displayHead = true,
+  }: CodeSnippetProps) => {
+    const codeRef = useRef(null)
+    const { translate } = useInternationalization()
+
+    useEffect(() => {
+      if (codeRef && codeRef.current) {
```
stephenlago99, line 41: Can be replaced with `codeRef?.current`

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,96 @@
...
+            {displayHead && (
+              <div className="flex h-nav items-center px-8 shadow-b">
+                <Typography variant="bodyHl">
+                  {translate('text_623b42ff8ee4e000ba87d0b2')}
+                </Typography>
+              </div>
+            )}
+            <pre
+              className={tw(
+                // https://prismjs.com/plugins/line-numbers/
```
stephenlago99, line 59: Can this comment be removed?

keellyp: I've added this comment to explain why this classname (not a tailwind one) is mandatory or it'll break the implementation

stephenlago99 (approved):

PR #1904, opened 2024-12-03
misc(Table): migrate to Tailwind

ansmonjol:
## Context

As part of our ongoing journey to migrate from styled components to tailwind, the Table component also needed refactoring.

## Description

Few points to note about this migration:
- All styles components are now JSX components written on top of Table definition, in the same file
- Styles are defined in 3 separated manner
  - `className` for "default" and toggled styles
  - `styles` for "dynamic" styles mapped on variables
  - `sx` for "complex" selector (specific classes, pseudo elements, ...)
- Component class names `lago-*` are not in variables. Even tho they are used multiple times, I had to prevent using variables because we cannot use dynamic classe names in `sx` selector. Those are not interpreted during run time but build and "transformed" as static css on the flight. Even with unchanged const it don't work.
- Kept usage of `setResponsiveProperty` and added a test for it. Also even tho this util is only used in Table today I preferred to keep it in it's separated file. It now return an object and not a component css object anymore.
- rewrote the Box wrapper with a div & classNames. Note the transform was useless and removed.
- You can notice a custom `captionHl` defined in a style of `TableInnerCell`. I choose not to turn that into a TW className that could help, but rather keep it like this and maybe migrate later if sure reproduce too often.

There should be no change in UI and UX

Fixes LAGO-422

### DIFF
--- a/src/components/designSystem/Table/Table.tsx
+++ b/src/components/designSystem/Table/Table.tsx
@@ -1,13 +1,13 @@
 import {
-  Box,
   Table as MUITable,
   TableBody as MUITableBody,
   TableCell as MUITableCell,
   TableHead as MUITableHead,
   TableRow as MUITableRow,
+  TableCellProps,
+  TableRowProps,
 } from '@mui/material'
-import { MouseEvent, ReactNode, useRef } from 'react'
-import styled from 'styled-components'
+import { MouseEvent, PropsWithChildren, ReactNode, useRef } from 'react'
 
 import { Button, IconName, Popper, Skeleton, Tooltip, Typography } from '~/components/designSystem'
 import { GenericPlaceholder, GenericPlaceholderProps } from '~/components/GenericPlaceholder'
@@ -17,6 +17,7 @@ import { useListKeysNavigation } from '~/hooks/ui/useListKeyNavigation'
 import EmptyImage from '~/public/images/maneki/empty.svg'
 import ErrorImage from '~/public/images/maneki/error.svg'
 import { MenuPopper, PopperOpener, theme } from '~/styles'
+import { tw } from '~/styles/utils'
 
 const PADDING_SPACING_RIGHT_PX = 32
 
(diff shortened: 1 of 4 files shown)


src/components/designSystem/Table/Table.tsx
```diff
@@ -216,38 +436,51 @@ export const Table = <T extends DataItem>({
...
+          },
+        }}
       >
-        <TableHead>
+        <MUITableHead>
           <tr>
             <>
               {filteredColumns.map((column, i) => (
                 <TableCell
+                  className="sticky top-0 z-[1] border-b-0 bg-white shadow-b"
```
keellyp, line 462: I would have gone with `z-10` directly to not enforce arbitrary value usage

src/components/designSystem/Table/Table.tsx
```diff
@@ -92,6 +93,225 @@ const countMaxSpaceColumns = <T,>(columns: TableColumn<T>[]) =>
...
+}: PropsWithChildren &
+  TableCellProps & {
+    className?: string
+    isBlurred?: boolean
+    hasPlaceholderDisplayed?: boolean
+    maxSpace?: number
+  }) => {
+  return (
+    <MUITableCell
+      className={tw('lago-table-cell w-auto whitespace-nowrap p-0', className)}
```
keellyp, line 152: WDYT about separating classnames like this to avoid mixing tailwind classes and custom ones?

```suggestion
      className={tw('lago-table-cell', 'w-auto whitespace-nowrap p-0', className)}
```

keellyp (approved):
Nitpick suggestions but overall LGTM 👏

src/components/designSystem/Table/Table.tsx
ansmonjol: Yeah indeed, and I don't think it matters that much to have 1 instead of 10, changing!

src/components/designSystem/Table/Table.tsx
ansmonjol: that's a good point, love it
updating

PR #1907, opened 2024-12-05
misc: fix some ts warnings

ansmonjol:
## Context

We recently updated TS in our app and some new linter are configured to show some warnings.

## Description

Those are to be solved along the way, and this is just a small attempt for few of them.

Gonna be a long road.


stephenlago99 (approved):
Looks good! Thanks for fixing those 🙌

PR #1911, opened 2024-12-05
bug: prevent deletion of default filter price in subscription form

ansmonjol:
## Context

We allow to remove the default pricing in a subscription even tho it should not happen

## Description

This PR disables the button in that specific case and shows a specific message on hover's tooltip

Fixes ISSUE-571


stephenlago99 (approved):
Looks good.

Side note: I think that at some point we should separate this into multiple smaller components, the file already has 1100+ lines, and having small pieces of logic hidden in the JSX will hurt maintainability.

PR #1913, opened 2024-12-05
misc(AmountInput): migrate to Tailwind

ansmonjol:
Fixes LAGO-436

### DIFF
--- a/src/components/form/AmountInput/AmountInput.tsx
+++ b/src/components/form/AmountInput/AmountInput.tsx
@@ -1,5 +1,4 @@
 import { forwardRef } from 'react'
-import styled, { css } from 'styled-components'
 
 import { getCurrencyPrecision } from '~/core/serializers/serializeAmount'
 import { CurrencyEnum } from '~/generated/graphql'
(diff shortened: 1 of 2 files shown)


src/components/form/AmountInput/AmountInput.tsx
```diff
@@ -77,9 +76,11 @@ export const AmountInput = forwardRef<HTMLDivElement, AmountInputProps>(
     const newPlaceholder = placeholder ?? definedDefaultPlaceholder(currency, translate)
 
     return (
-      <StyledTextInput
+      <TextInput
         ref={ref}
-        $displayErrorText={displayErrorText}
+        InputProps={{
+          style: !displayErrorText ? { marginBottom: '0 !important' } : {},
```
keellyp, line 82: is !important mandatory?

keellyp (approved):

src/components/form/AmountInput/AmountInput.tsx
ansmonjol: well, need to find an implem but have tried to be iso in the meantime

Will look

ansmonjol: that's been unessesary recently, since the `TextInput` does not use margin but gap internally now

so the spacing is automatically not present if there is no error displayed

Will remove!

ansmonjol: Done

PR #1915, opened 2024-12-09
fix: use string instead of number to avoid scientific notation

keellyp:
## Context

We had a decimal limitation on the volume and graduated table
image

## Description

Turn the value into a number breaks the input as it turns the value into a scientific notation and after 1e-7 we were not able to add more decimals. Turning into a string should fix the issue.


stephenlago99 (approved):
Looks good!

PR #1921, opened 2024-12-11
refactor: migrate switch to tailwind

keellyp:
## Context

Tailwind migration

## Description

Migrate Switch component

Fixes LAGO-428


stephenlago99 (approved):
Looks good!

Just as a side-note, I don't think comments like `// This is for preventing setState on unmounted component` add any real value (I know you've just updated it, not added it). In general, code should be self descriptive, and the ref.current pattern is well-known in the React world. It's ok to include for now.

PR #1928, opened 2024-12-20
refactor: migrate styled-components usage of form components

keellyp:
## Context

Part of the tailwind migration

## Description

Remove usage of `styled(Component)` of form components.

Better to review commit by commit

Partially fixes LAGO-332


src/components/form/TextInput/TextInput.tsx
```diff
@@ -248,6 +251,31 @@ export const TextInput = forwardRef<HTMLDivElement, TextInputProps>(
                 : {}),
             ...InputProps,
           }}
+          sx={
+            resetStyle
```
ansmonjol, line 255: What about having the `props.sx` applied if present, otherwise applying the default style written here.
That would allow removing this new props and only places that needs a custom style can define them on invocation.

Also, can we have this "reset" style in the MUI config or does it conflict with another component using the input?

src/components/SearchInput.tsx
```diff
@@ -17,8 +16,8 @@ export const SearchInput = ({ className, onChange, placeholder }: SearchInputPro
   const [localValue, setLocalValue] = useState<string>('')
 
   return (
-    <TextInputForSearch
-      className={className}
+    <TextInput
+      className={tw('max-w-60 [&_input]:h-10 [&_input]:!pl-3', className)}
```
ansmonjol, line 20: Those custom `inputAdornedStart` style sounds like they could be the same for all usage WDYT? Maybe defining them in the MUI default them should do the job?

src/components/form/TextInput/TextInputField.tsx
```diff
@@ -30,8 +29,14 @@ export const TextInputField = memo(
       const { values, errors, touched, handleBlur, setFieldValue } = formikProps
 
       return (
-        <StyledTextInput
-          $displayErrorText={displayErrorText}
+        <TextInput
+          sx={
+            !displayErrorText
+              ? {
+                  '.MuiTextField-root': { marginBottom: '0 !important' },
```
ansmonjol, line 36: Just a guess here, if the elements use a gap between elements rather than a margin bottom, we could get rid of this `sx` definition correct?

src/components/plans/ChargeAccordion.tsx
```diff
@@ -854,7 +854,8 @@ export const ChargeAccordion = memo(
                   </Button>
                 ) : (
                   <SpendingMinimumWrapper>
-                    <SpendingMinimumInput
+                    <AmountInput
+                      className="flex-1"
```
ansmonjol, line 858: Maybe not something we want to tackle not but felt worth mentioning: all inputs (and generally all elements in our design system) should take the maximum width by default. And only when needed we can define a custom max-width.
Feel free to ignore as it may lead to many check and fixes to migrate now :)

src/components/form/MultipleComboBox/MultipleComboBox.tsx
```diff
@@ -229,7 +235,7 @@ const Container = styled.div`
 
   /* Fix the placement of the adornment elements */
   .MuiAutocomplete-endAdornment {
-    top: calc(50% - 19px);
+    top: calc(50% - 12px);
```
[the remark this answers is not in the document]
keellyp: We had an issue with the button placement
Applying a size and center it like this fixed it

src/components/SearchInput.tsx
keellyp: Do you mean a custom class in the muiTheme?

src/components/form/TextInput/TextInput.tsx
keellyp: Not sure to understand your point, do you mean that all invocations should pass a props `sx` instead of `resetStyle` ?

src/components/form/TextInput/TextInputField.tsx
keellyp: I guess too! But changing it everywhere in this PR would have been too much imo

src/components/form/TextInput/TextInput.tsx
ansmonjol: You should apply the sx props only if it's passed. Otherwise (in all places you use `resetStyle` in invocation), the default style will apply by default.

src/components/SearchInput.tsx
ansmonjol: More like a style defined in muiTheme for `MuiInputBase-inputAdornedStart`

We already have one defined under the `MuiOutlinedInput` key and it's probably the exact same place you could update 🤔

src/components/form/TextInput/TextInputField.tsx
ansmonjol: Ok! But may be worth to check, I have the feeling this rule is obsolete as the TextInput wrapper already apply those rules `flex flex-col gap-1`

src/components/form/MultipleComboBox/MultipleComboBox.tsx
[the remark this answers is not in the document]
ansmonjol: ok!

src/components/form/TextInput/TextInput.tsx
ansmonjol: Discussed together, this attribute cannot be factorised so better to keep such approach. We'll think about maybe a better naming but keep the props way

src/components/SearchInput.tsx
ansmonjol: used only for search

ansmonjol (approved):

PR #1960, opened 2024-12-31
misc: use Table component for AddOns list

ansmonjol:
Uses Table component for the Addon list page


keellyp (approved):
I think you need to udpate the e2e tests

PR #1966, opened 2025-01-03
feat: customer creation revamp

keellyp:
## Context

Customer creation and edition revamp 

## Description

This PR adds 2 new routes `/customer/create` and `/customer/:id/edit`
It also removes the usage of the drawer to create and edit a customer.

**Need design review ⚠️**


src/hooks/useCreateEditCustomer.ts
```diff
@@ -172,49 +198,64 @@ export const useCreateEditCustomer: UseCreateEditCustomer = ({ customer }) => {
...
-                    : providerCustomer?.providerPaymentMethods,
+    loading,
+    isEdition: !!customerId,
+    customer: customer || undefined,
+    onClose: () =>
+      customerId
+        ? navigate(generatePath(CUSTOMER_DETAILS_ROUTE, { customerId }))
+        : navigate(CUSTOMERS_LIST_ROUTE),
+    onSave:
+      !!customer && !!customerId
```
stephenlago99, line 222: It's a bit to follow the logic defined here. We could extract it in three separate functions:

```
const onSave = ({ ... }) => {
  const input = {
                  id: customer?.id as string,
                  paymentProvider,
                  providerCustomer: {
                    providerCustomerId: !paymentProvider
                      ? null
                      : providerCustomer?.providerCustomerId,
                    syncWithProvider: !paymentProvider ? null : providerCustomer?.syncWithProvider,
                    providerPaymentMethods: !providerCustomer?.providerPaymentMethods?.length
                      ? [ProviderPaymentMethodsEnum.Card]
                      : providerCustomer?.providerPaymentMethods,
                  },
                  ...values,
   };
   
  if (customer && customerId) {
    return onCustomerUpdate(input);
  }
  
  return onCustomerCreate(input);
};

return {
  ...,
  onSave
}
```

This way we only do the boolean logic once and re-use it, and also the `onSave` function becomes more clear.

keellyp: You're right, updated here: [fix: code review](https://github.com/getlago/lago-front/pull/1966/commits/53f8bec3b169eb0de24ca014da81edc4d0bf9fdd)

stephenlago99 (approved):

PR #1970, opened 2025-01-06
misc(invoice-list-item): add issue a credit note action

ansmonjol:
A customer mentioned it would be great to have this action "issue a credit note" from the invoice item actions directly

Fixes ISSUE-616

### DIFF
--- a/src/components/customers/CustomerInvoicesList.tsx
+++ b/src/components/customers/CustomerInvoicesList.tsx
@@ -7,7 +7,10 @@ import { addToast, hasDefinedGQLError } from '~/core/apolloClient'
 import { CustomerInvoiceDetailsTabsOptionsEnum } from '~/core/constants/NavigationEnum'
 import { invoiceStatusMapping, paymentStatusMapping } from '~/core/constants/statusInvoiceMapping'
 import { intlFormatNumber } from '~/core/formats/intlFormatNumber'
-import { CUSTOMER_INVOICE_DETAILS_ROUTE } from '~/core/router'
+import {
+  CUSTOMER_INVOICE_CREATE_CREDIT_NOTE_ROUTE,
+  CUSTOMER_INVOICE_DETAILS_ROUTE,
+} from '~/core/router'
 import { deserializeAmount } from '~/core/serializers/serializeAmount'
 import { formatDateToTZ, getTimezoneConfig } from '~/core/timezone'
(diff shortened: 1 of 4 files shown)


src/components/customers/CustomerInvoicesList.tsx
```diff
@@ -335,6 +356,9 @@ export const CustomerInvoicesList: FC<CustomerInvoicesListProps> = ({
                 paymentStatus,
               ) &&
               hasPermissions(['invoicesVoid'])
+            const canIssueCreditNote =
+              ![InvoiceStatusTypeEnum.Draft, InvoiceStatusTypeEnum.Voided].includes(status) &&
+              hasPermissions(['creditNotesCreate'])
```
stephenlago99, line 361: I've noticed that we also check for this in `CustomerInvoiceDetails`.

Maybe we can extract this into an utility so we can guarantee consistent behaviour.

src/components/invoices/InvoicesList.tsx
```diff
@@ -218,17 +251,31 @@ const InvoicesList = ({
...
-                  title: translate('text_63ac86d897f728a87b2fa031'),
-                  onAction: ({ id }) => {
-                    copyToClipboard(id)
-                    addToast({
-                      severity: 'info',
-                      translateKey: 'text_63ac86d897f728a87b2fa0b0',
-                    })
-                  },
-                },
+                canIssueCreditNote
```
stephenlago99, line 254: Having nested boolean logic like this can make it harder to follow the flow.

Maybe we could replace it with something that's a bit more clear:

```
(canIssueCreditNote && isPremium) ? {
  ...
} : null,
(canIssueCreditNote && !isPremium) ? {
  ...
} : null,
```

And if the underlying logic works with any falsy value, it can be even cleaner:

```
canIssueCreditNote && isPremium && {
  ...
},
canIssueCreditNote && !isPremium && {
  ...
},
```

src/components/customers/CustomerInvoicesList.tsx
ansmonjol: I would prefer to duplicate for now, while it's not too big and only occurs twice in the app

src/components/invoices/InvoicesList.tsx
ansmonjol: Sure will update

PR #1973, opened 2025-01-07
misc(Table): allow to open row link in new tab

ansmonjol:
## Context

A customer asked to be able to open row link redirection in a new tab.

## Description

We're not the best in term of accessibility and it's kinda hard to have a native `a` link into a table structure without increasing the complexity of the Table component.

I noticed we only use the `onRowAction` to execute a navigate all over the app, so decided to rename it and for it to return a string.
So we can check if meta or ctrl key are pressed to open in new tab, otherwise use the navigate from `rect-router-dom` as usual.
Put that into the same commit as pretty tied

Fixes ISSUE-615

### DIFF
--- a/src/components/creditNote/CreditNotesTable.tsx
+++ b/src/components/creditNote/CreditNotesTable.tsx
@@ -298,15 +298,13 @@ const CreditNotesTable = ({
 
                   return actions
                 }}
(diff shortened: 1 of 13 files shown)


stephenlago99 (approved):
Looks good, nice idea of unifying the logic like that!

PR #1976, opened 2025-01-08
feat: yarn -> pnpm

stephenlago99:
Replace `yarn` with `pnpm` to improve the dependency installation speeds.


stephenlago99:
Converted this to a draft since some tasks with more priority came in.

To wrap this up, only the following need to be done:
- Fix `.github/workflows/cypress.ts` by caching the Cypress binary between runs (currently it's failing because it can't find the binary)

Suggestion:

```
      - name: Cache Cypress binary
        uses: actions/cache@v3
        with:
          path: ~/.cache/Cypress
          key: cypress-binary-${{ hashFiles('pnpm-lock.yaml') }}
          restore-keys: |
            cypress-binary-
```

- Fix `.github/workflows/codegen.ts` by waiting until the graphql endpoint is up (currently it's failing because the install takes just a few seconds and when `codegen` is ran the API is not up yet)

Suggestion:

```
      - name: Launch API
        run: docker compose -f ./ci/docker-compose.ci.yml --env-file ./.env up -d api

      - name: Wait for API to be ready
        run: |
          timeout 60s bash -c '
          until curl -s -f -o /dev/null "http://localhost:3000/graphql"; do
            echo "Waiting for API to be ready..."
            sleep 2
          done
          '
```

ansmonjol (approved):

ansmonjol:
I'm fixing lago build before merging

keellyp (approved):

PR #1977, opened 2025-01-08
bug(Combobox): fix element height with description

ansmonjol:
Fixes the Combobox element height if it contains a description.

Note: I'm tired of those shenanigans and random calculation values there and in the multiple combobox. I think we should (i) upgrade MUI to v6, (ii) negociate to have a combobox looking like the MUI doc with design, (iii) remove all our custom implementation to get something simpler that works without all this magic and the world would be much safer place.

Fixes ISSUE-619


stephenlago99 (approved):
Looks good, (i) sounds good!

ansmonjol:
> Looks good, (i) sounds good, especially if it would enable us to get rid of some custom stuff.

Yeah it appear they changed their API on v6 also removed some very custom ways to implement and are more into overriding their style. Which is pretty handy as we want to benefit some of their internal logic such as item height calculation 😅 
Will empty out backlog and try to upgrade.

PR #1985, opened 2025-01-10
misc: unify coupons experience

ansmonjol:
## Context

We figured out we had discrepancy in the coupon actions over the app.
1. You were able to terminate a coupon anytime from the details, but not from the coupon list (conditionally)
2. You were not able to terminate a coupon from the details page

## Description

This PR fixes those and also
- displays the coupon status in the details page. After adding the terminate action it was blurry about the coupon status there.
- make sure active queries (coupon details, delete or terminate coupon) are not fired again after a coupon is deleted. That used to trigger an error modal in the app before redirection to coupon list
- display the correct wording in the delete coupon modal depending if the coupon have been applied to a customer or not

Fixes ISSUE-620


src/components/coupons/DeleteCouponDialog.tsx
```diff
@@ -69,10 +73,34 @@ export const DeleteCouponDialog = forwardRef<DeleteCouponDialogRef>((_, ref) =>
...
+        couponLoading ? (
+          <Skeleton className="mb-5 h-4 w-full" variant="text" />
+        ) : (
+          translate('text_628b432fd8f2bc0105b973ee', {
+            couponName: coupon?.name,
+          })
+        )
+      }
+      description={
+        couponLoading ? (
```
stephenlago99, line 86: Not something that needs to be fixed necessarily, but we should try to stay away from nested ternaries, as it makes logic hard to follow:

```ts
<>
  {couponLoading && (
    <>
      <Skeleton className="mb-4 w-full" variant="text" />
      <Skeleton className="w-full" variant="text" />
     </>
  )}
  
  {!couponLoading && !!coupon?.appliedCouponsCount && (
    <Typography
      html={translate('text_17364422965884zgujkr1l7j',
        { appliedCouponsCount: coupon.appliedCouponsCount },
        coupon.appliedCouponsCount,
      )}
    />  
  )}
  
  {!couponLoading && !coupon?.appliedCoponsCount && (
    <Typography html={translate('text_628b432fd8f2bc0105b973f6')} />
  )}
</>
```

While a counter-argument can be made about the fact that we increase the number of lines of code by 2-3, I think that it is an ok trade-off for gaining the ability to follow the logic more easily.

stephenlago99 (approved):

PR #1997, opened 2025-01-21
feat(CustomerRevamp): Refactor customer subscription list, customer coupons and customer details

stephenlago99:
## Description

This PR includes the full customer revamp:
- Removal of the customer details drawer
- Change from lists to <Table>s (Subscriptions / Coupons)
- Add more formatting to Customer details
- Add/change titles/subtitles to all the sections


src/components/coupons/CouponCaption.tsx
```diff
@@ -44,104 +44,123 @@ export interface CouponMixedType extends CouponItemFragment {
...
-})
+    return (
+      <>
+        {!className && (
+          <Typography variant={variant} color="grey600" noWrap data-test="coupon-caption">
+            {getCaption()}
+          </Typography>
+        )}
+
+        {className && (
```
keellyp, line 156: Why do we need to check `className` to render the coupon caption?

stephenlago99: Good question! I didn't want to change the underlying implementation of CouponCaption too much, as it's used in a bunch of places. Hence, old places use the `variant` prop and new places use the `className` prop.

PR #1999, opened 2025-01-22
bug: define allowedHosts in vite config

ansmonjol:
A recent [update of vite](https://github.com/getlago/lago-front/pull/1998) package requires us to configure the allowed hosts for local dev, in order to whitelist the proxy.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -81,6 +81,7 @@ export default defineConfig(({ mode }) => {
       port,
       host: true,
       strictPort: true,
+      allowedHosts: ['app.lago.dev'],
     },
     preview: {
       port,


vite.config.ts
```diff
@@ -81,6 +81,7 @@ export default defineConfig(({ mode }) => {
       port,
       host: true,
       strictPort: true,
+      allowedHosts: ['app.lago.dev'],
```
keellyp, line 84: shouldn't we use a env variable instead ?

ansmonjol: Yes we can
Just pushed an update

stephenlago99 (approved):

vite.config.ts
ansmonjol: well, i'm struggling to have it work during build, not sure why tho
will spend some time later on this, let's fix the main branch first

keellyp (approved):

PR #2001, opened 2025-01-23
feat: create payment page

keellyp:
## Context

Related to partial-payment feature

## Description

Add creation page.
Can be safely merged when back-end is ready as the page is not yet used in the app.


src/core/timezone/utils.ts
```diff
@@ -53,3 +53,33 @@ export const intlFormatDateToDateMed = (
     locale: locale,
   }).toLocaleString(DateTime.DATE_MED)
 }
+
+export const intlFormatDateTime = (
+  date: string,
+  {
+    timezone,
```
ansmonjol, line 60: What about having UTC as default here? `TimezoneEnum.TzUtc`

PR #2006, opened 2025-01-30
feat(lago-668) - Manual payment

keellyp:
## Context

We want to introduce the concept of partial payment and record partial manual payment.

## Description

This PR should: 
- create a new page `/create/payment`
- add new `payment` tab across the app
- display new `Record payment` button in the app
- handle partial payment new status
- handle refund of partial payment on credit notes

It also extracts Metadata from getInvoiceDetails query to avoid complexity issue.

Status: QA approved

Note: It's better to review commit by commit as each one is linked to an issue if you need more context.

Fixes LAGO-668


src/components/invoices/Metadatas.tsx
```diff
@@ -32,18 +33,40 @@ gql`
...
     }
   }
-`
 
-interface MetadatasProps {
-  customer: CustomerMetadatasForInvoiceOverviewFragment
-  invoice: InvoiceMetadatasForInvoiceOverviewFragment
-}
+  query getInvoiceMetadatas($id: ID!) {
+    invoice(id: $id) {
```
ansmonjol, line 38: Something I tend to do all the time on the first model (at least) requested on a query is to hardcode the `id` property, to help the cache storing those results.
We don't have any rules enforcing such not convention in the team, but if you also feel that's relevant, even if the id is maybe spread in one of the fragments bellow, maybe you can consider adding the `id` here

ansmonjol: I also need to dig into store behaviour cause not having the id probably works perfectly fine. Just IIRC values are stored by `MODEL_NAME:ID`

src/components/customers/CustomerPaymentsList.tsx
```diff
@@ -0,0 +1,142 @@
...
+              />
+            ),
+          },
+          {
+            key: 'payable.__typename',
+            title: translate('text_63ac86d797f728a87b2f9fad'),
+            minWidth: 160,
+            maxSpace: true,
+            content: ({ payable }) => {
+              if (payable.__typename === 'Invoice') {
```
ansmonjol, line 95: I understand this value is not accessible from another attribute right?
Is there a way to have such returned from BE in a new attribute? Could be a graphQl api only attribute maybe
I always find it dangerous to rely on the `typeName`

PR #2009, opened 2025-02-03
feat: Revenue share

stephenlago99:
## Context

This enables creating partners / creating invoices for partners

## Description

- A new quick filter was added for Customer type
- A new filter was added for self billed invoices
- A new field was added when creating/editing a customer
- Invoices page:
  - Flipped columns that contain details (since the billing is inversed)
  - Display a new Alert
  - Instead of "Customer name", use "Partner name"


src/pages/CreateCustomer.tsx
```diff
@@ -255,6 +266,27 @@ const CreateCustomer = () => {
           </div>
 
           <div className="mb-8 flex flex-col gap-12 not-last-child:pb-12 not-last-child:shadow-b">
+            {/* eslint-disable-next-line */}
```
ansmonjol, line 269: why do we have to disable it here?

ansmonjol (approved):

src/pages/CreateCustomer.tsx
ansmonjol: What about using the Switch component directly and handle this premium case in the switch onClick function?

Just a suggestion up to you to implement!

stephenlago99: Good suggestion, but unfortunately looks like onClick is implemented in a very custom way inside Switch. I wouldn't change it now, as it could break the existing logic.

ansmonjol: You have an usage example in `src/pages/settings/EmailSettings.tsx:128`
Feel free to implement the same way and/or resolve this conversation 🙌

keellyp (approved):
LGTM!

PR #2010, opened 2025-02-03
misc: add new HorizontalDataTable

ansmonjol:
We need a new table component for the analytics v2. This uses tanstack virtual internally to display lot of DOM elements with performances.
Also it adopts a column approach, to fit with the data endpoints data shape

This components can evolve and will not be used on user-facing features yet.


keellyp (approved):
Nitpick but should this component be located under `Table` folder?

PR #2014, opened 2025-02-05
feat: Filters prefix

stephenlago99:
## Context

The current Filters architecture makes it impossible to have 2 sets of filters on the same page, because of how the filters are stored in the URL params.

## Description

Added a new prop `filtersNamePrefix` which enables us to have multiple Filters on the same page by prefixing their name:

```
https://app.lago.dev/invoices/invoices?
  in_amount=isBetween%2C5%2C100
  &in_currency=USD
  &cn_amount=isBetween%2C5%2C6
  &cn_creditNoteCreditStatus=voided
```

To use this functionality, we need to pass the new prop in two places:
- The Filters.Provider
```
  <Filters.Provider
    quickFiltersType={AvailableQuickFilters.InvoiceStatus}
    filtersNamePrefix="in"
    ...
```

- When calling the `formatFiltersForQuery` function
```
  return formatFiltersForQuery({
    searchParams,
    availableFilters: InvoiceAvailableFilters,
    filtersNamePrefix: 'in',
  })
```

## Implementation

I've tried to keep the existing structure, without changing too many things. Most of the work was done inside `useFilters`, since it assumed a single set of filters in most cases (for example when setting a quick filter, or when resetting filters, it was deleting everything from the search params).

## Screenshot

image

### DIFF
--- a/src/components/designSystem/Filters/useFilters.ts
+++ b/src/components/designSystem/Filters/useFilters.ts
@@ -10,37 +10,50 @@ export const useFilters = () => {
   const [searchParams] = useSearchParams()
   const searchParamsObject = Object.fromEntries(searchParams.entries())
 
-  const resetFilters = () => {
+  const prefix = context.filtersNamePrefix
+
+  const keyWithoutPrefix = (key: string) => (prefix ? key.replace(`${prefix}_`, '') : key)
+  const keyWithPrefix = (key: string) => (prefix ? `${prefix}_${key}` : key)
+
+  const removeExistingFilters = () => {
     // Only remove the filters from the URL that are currently applied and are removable (availableFilters)
     for (const search in searchParamsObject) {
-      // @ts-expect-error - searchParams is typed as a string and not as AvailableFiltersEnum
-      if (context.availableFilters.includes(search)) {
+      const key = keyWithoutPrefix(search) as AvailableFiltersEnum
+
+      if (context.availableFilters.includes(key)) {
         searchParams.delete(search)
       }
     }
+  }
+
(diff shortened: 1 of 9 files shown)


keellyp (approved):

src/components/designSystem/Filters/useFilters.ts
```diff
@@ -10,37 +10,50 @@ export const useFilters = () => {
...
+  const keyWithPrefix = (key: string) => (prefix ? `${prefix}_${key}` : key)
+
+  const removeExistingFilters = () => {
     // Only remove the filters from the URL that are currently applied and are removable (availableFilters)
     for (const search in searchParamsObject) {
-      // @ts-expect-error - searchParams is typed as a string and not as AvailableFiltersEnum
-      if (context.availableFilters.includes(search)) {
+      const key = keyWithoutPrefix(search) as AvailableFiltersEnum
+
+      if (context.availableFilters.includes(key)) {
```
ansmonjol, line 23: I wonder if the key could not be an issue if one filter have a prefix, and a second filter on the same page doesn't.

I imagine something like removing `currency` when `in_currency` exists too. I guess both would be removed right?

ansmonjol: Probably the same for all includes used in this file. Could be a find index by exact value and then mutate the object of search params?

stephenlago99: We have decided to make the filters mandatory, as having it optional would open the door to too many edge cases.

ansmonjol (approved):

PR #2033, opened 2025-02-19
feat: add stripe crypto payment method

keellyp:
## Context

- Add crypto support
- Refactor a bit the file to avoid repeating `onPaymentSet` function 
- Removing eslint warnings and Stack usage

Status: QA review in progress

### DIFF
--- a/src/components/customers/CustomerMainInfos.tsx
+++ b/src/components/customers/CustomerMainInfos.tsx
@@ -1,11 +1,11 @@
 import { gql } from '@apollo/client'
-import { Stack } from '@mui/material'
 import { FC, PropsWithChildren } from 'react'
 import { Link, LinkProps } from 'react-router-dom'
 
 import { TRANSLATIONS_MAP_CUSTOMER_TYPE } from '~/components/customers/utils'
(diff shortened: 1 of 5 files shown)


src/components/customers/CustomerMainInfos.tsx
```diff
@@ -552,20 +558,10 @@ export const CustomerMainInfos = ({ loading, customer, onEdit }: CustomerMainInf
                   {translate('text_62b1edddbf5f461ab9712795')}
                 </Typography>
                 <div>
-                  <Stack direction="row" spacing={2} alignItems="center">
-                    <Avatar variant="connector-full" size="small">
-                      {paymentProvider === ProviderTypeEnum?.Stripe ? (
-                        <Stripe />
```
ansmonjol, line 558: We do not display the PP image anymore but the default receipt icon instead?

keellyp: The payment provider chip is taking care for the logic of rendering the correct logo + name, it should not change anything in the UI

ansmonjol (approved):

PR #2053, opened 2025-03-04
misc: add new filters for revenue streams

ansmonjol:
## Context

We're building the new analytics section of the app. This section is not accessible yet and we're pushing code piece by piece.

## Description

This PR adds the new filters needed for the revenue streams query. It makes sure it's correctly implemented with the existing filters system and has proper tests for formating

### DIFF
--- a/src/components/designSystem/Filters/filtersElements/FiltersItemDate.tsx
+++ b/src/components/designSystem/Filters/filtersElements/FiltersItemDate.tsx
@@ -1,0 +1,44 @@
+import { DateTime } from 'luxon'
+
+import { DatePicker } from '~/components/form'
+import { useInternationalization } from '~/hooks/core/useInternationalization'
+
+import { Typography } from '../../Typography'
+import { FiltersFormValues } from '../types'
+
+type FiltersItemDateProps = {
+  value: FiltersFormValues['filters'][0]['value']
+  setFilterValue: (value: string) => void
+}
+
+export const FiltersItemDate = ({ value = ',', setFilterValue }: FiltersItemDateProps) => {
+  const { translate } = useInternationalization()
+  const [givenValueFrom, givenValueTo] = value.split(',')
+
+  return (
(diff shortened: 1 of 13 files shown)


src/components/designSystem/Filters/filtersElements/FiltersItemDate.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  return (
+    <div className="flex items-center gap-2 lg:gap-3">
+      <DatePicker
+        showErrorInTooltip
+        className="flex-1"
+        onChange={(dateFrom) => {
+          // replace the value.split(',')[0] with the new value
+          setFilterValue(
+            `${DateTime.fromISO(dateFrom as string).startOf('day')},${value.split(',')[1]}`,
```
stephenlago99, line 25: Since we use `value.split(',')` in a couple of places in this file, I think it would improve the readability if we assigned it to variables at the beginning of the component & we use those throughout the file:

```
export const FiltersItemDate = ({ value = ',', setFilterValue }: FiltersItemDateProps) => {
  ...
  const [x, y] = value.split(',');
  ...
```

src/components/designSystem/Filters/filtersElements/FiltersItemPlanCode.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { translate } = useInternationalization()
+  const [getPlans, { data, loading }] = useGetPlansForFiltersItemPlanCodeLazyQuery({
+    variables: { page: 1, limit: 10 },
+  })
+
+  const comboboxPlansData = useMemo(() => {
+    if (!data?.plans?.collection) return []
+
+    return data.plans.collection.map((plan) => {
+      const { code } = plan
```
stephenlago99, line 40: We could save a few lines of code by directly using the value:

```
return data.plans.collection.map((plan) => ({
  value: plan.code
}))
```

stephenlago99 (approved):

PR #2058, opened 2025-03-07
feat: revamp create wallet and create topup layout

keellyp:
## Context

Part of the wallet improvement feature

## Description

Revamp the layout to use new centered one

Fixes LAGO-759


src/components/layouts/CenteredPage.tsx
```diff
@@ -38,10 +40,22 @@ const CenteredStickyFooter = ({ children }: PropsWithChildren) => {
   )
 }
 
+const PageTitle = ({ title, description }: { title: string; description?: string }) => {
+  return (
+    <div className="not-last-child:mb-1">
```
ansmonjol, line 45: Code preference here so don't feel you have to implement it.

I find it hard to read those not last child classes at first look, also wonder if we could have a flex and gap approach, in case the description becomes optional (saying cause it usually is) and that having a margin could bring some alignment issues, if that make sense.

Up to you!

src/pages/wallet/CreateWalletTopUp.tsx
```diff
@@ -0,0 +1,332 @@
...
+
+  const onAbort = useCallback(() => {
+    formikProps.dirty ? warningDialogRef.current?.openDialog() : navigateToCustomerWalletTab()
+  }, [formikProps.dirty, navigateToCustomerWalletTab])
+
+  const setCurrencyPrecision = useCallback(
+    (value: number) => {
+      let precision
+
+      if (currencyPrecision === 3) {
```
ansmonjol, line 141: Could we leverage `getCurrencyPrecision` method here by any chance? To prevent having logic spread

ansmonjol (approved):
Just minor comments, approving in the meantime

src/pages/wallet/CreateWalletTopUp.tsx
keellyp: Addressed in this commit: [fix: code review](https://github.com/getlago/lago-front/pull/2058/commits/164854faa820af26c159fb41254d939b260c84dc)

src/components/layouts/CenteredPage.tsx
keellyp: See [fix: code review](https://github.com/getlago/lago-front/pull/2058/commits/164854faa820af26c159fb41254d939b260c84dc)

PR #2061, opened 2025-03-10
misc: format date correctly in table's header

ansmonjol:
## Context

We're building the new version of analytics on a page non-accessible by customers. Pushing update as they are built as they are not user-facing yet.

## Description

This PR makes sure the date format is correct according to the selected filter.
This introduces a new filter util with it's corresponding tests.

| Format | Display |
|--------|---------|
| Daily | image |
| Weekly | image |
| Monthly | image |

### DIFF
--- a/src/components/analytics/RevenueStreamsOverviewSection.tsx
+++ b/src/components/analytics/RevenueStreamsOverviewSection.tsx
@@ -5,9 +5,11 @@ import { useSearchParams } from 'react-router-dom'
 
 import { HorizontalDataTable, Icon, Typography } from '~/components/designSystem'
 import {
+  AvailableFiltersEnum,
   AvailableQuickFilters,
   Filters,
   formatFiltersForRevenueStreamsQuery,
+  getFilterValue,
   RevenueStreamsAvailablePopperFilters,
 } from '~/components/designSystem/Filters'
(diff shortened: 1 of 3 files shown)


src/components/analytics/RevenueStreamsOverviewSection.tsx
```diff
@@ -256,11 +258,26 @@ const RevenueStreamsOverviewSection = () => {
...
+                  const currentTimeGranularity = getFilterValue({
+                    key: AvailableFiltersEnum.timeGranularity,
+                    searchParams,
+                    prefix: REVENUE_STREAMS_OVERVIEW_FILTER_PREFIX,
+                  })
+
+                  const getFormatedTimeGramularity = (): string => {
+                    switch (currentTimeGranularity) {
+                      case TimeGranularityEnum.Daily:
+                        return formatDateToTZ(item.startOfPeriodDt, TimezoneEnum.TzUtc, 'LLL d')
```
keellyp, line 271: can we use the intlFormat function instead of using this schema?

ansmonjol: `intlFormatDateToDateMed` you mean? Didn't found a correct date formater for all formats here but I can probably tweak the options 👀

ansmonjol: Found it!

PR #2065, opened 2025-03-11
misc: move revenue streams overview logic into hook

ansmonjol:
## Context

We're building the new version of analytics on a page non-accessible by customers. Pushing update as they are built as they are not user-facing yet.

## Description

This PR simply move revenue streams overview logic into a hook, that will later have it's own tests. Also, a util have been created for the data padding logic and I'll implement it in a following PR

### DIFF
--- a/src/components/analytics/useRevenueAnalyticsOverview.ts
+++ b/src/components/analytics/useRevenueAnalyticsOverview.ts
@@ -1,0 +1,145 @@
+import { gql } from '@apollo/client'
+import Decimal from 'decimal.js'
+import { useMemo } from 'react'
+import { useSearchParams } from 'react-router-dom'
+
+import { formatRevenueStreamsData } from '~/components/analytics/utils'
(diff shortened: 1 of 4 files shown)


src/components/analytics/useRevenueAnalyticsOverview.ts
```diff
@@ -0,0 +1,145 @@
...
+  }, [searchParams, organization])
+
+  const formattedRevenueStreamsData = useMemo(() => {
+    return formatRevenueStreamsData({
+      searchParams,
+      data: revenueStreamsData?.dataApiRevenueStreams.collection,
+    })
+  }, [revenueStreamsData, searchParams])
+
+  const { lastNetRevenueAmountCents, netRevenueAmountCentsProgressionOnPeriod } = useMemo(() => {
```
stephenlago99, line 107: Can we extract the inner function into an utils file?

keellyp (approved):

PR #2074, opened 2025-03-17
misc(NavigationTab): allow to rely on tab index rather than url link

ansmonjol:
## Context

This is an improvement needed for the new analytics page. 
There is the need to have 2 navigation tab system in the same page.

## Description

This PR allows the navigation tab to be run without relying on the page URL but rather on the tabs index number.
pretty easy as internally it relies on a `value` that correspond to the selected tab index.

Doing so, we prevent any URL search params update and make it seamless to add any numbers of tabs in the same page.
The only inconvenience is that the url cannot be shared on those index-based tabs.
This was found as a quick win rather than having to update the whole filters and tab url management system


stephenlago99 (approved):
Really clean! Happy to see that the map idea worked.

PR #2080, opened 2025-03-19
feat: release revenue streams new analytics dashboard

ansmonjol:
## Context

We're releasing a first chunk of the new analytics dashboard to customers.

## Description

This PR performs few actions
- move some files to different folders, for better architecture and ease the integrations of future graphs
- improve the table component, so we can define the number of rows to show during loading state (and keeping the default behaviour)
- have a "switch" to allow rendering the new or old analytics depending on a flag on orga. Note I had to show a spinner while the orga loads, otherwise we can have a glitch in the different loading states of each analytics pages.
- prevent some re-redirection, if the page is currently active. Having some redirection can mess up with the search params initialised on the new analytics page. Also it's good to not force a redirection event if it's not necessary.
- rename the used premium integration addon

So only users with this premium orga flag will see the revenue streams page. All other users will have the current analytics page of the app. Nothing changes for them
That also allow to test the new data deployment pipeline, before we can enable the flag for specific users.

Better reviewed commit by commit

Fixes LAGO-751


src/pages/analytics/NewAnalytics.tsx
```diff
@@ -27,45 +23,44 @@ const NewAnalytics = () => {
...
+          //   }),
+          //   component: <PrepaidCredits />,
+          // },
+          // {
+          //   title: translate('text_63ac86d797f728a87b2f9f85'),
+          //   link: generatePath(ANALYTIC_TABS_ROUTE, {
+          //     tab: NewAnalyticsTabsOptionsEnum.invoices,
+          //   }),
+          //   component: <Invoices />,
+          // },
```
keellyp, line 63: is it still relevant?

keellyp (approved):

src/pages/analytics/NewAnalytics.tsx
ansmonjol: I think I commented them by laziness haha.
So the say I work back on those page that needs to be added I will only uncomment.
They were commented in the first place cause For now only the first tab exists, and so we prevent displaying "dead" links let's say
But indeed it's not very beautiful shipping this, I'll remove!

PR #2115, opened 2025-04-04
refactor: update dates util function 

keellyp:
## Context 

Using Luxon [formatting presets](https://github.com/moment/luxon/blob/master/docs/formatting.md#presets) is not enough for our use case : 
1. It doesn't handle the timezone display like `UTC+4`
2. It doesn't handle the dates display like `Apr 2024` or `Apr 18, 25` 
3. It's a mix of DATE_... TIME_... and DATETIME_... which is not ideal for us as many times, we need to add text like `From ... to ...` or `October 14, 1983 at 1:30 UTC+1` or we need DATETIME_MED_WITH_WEEKDAY but with TIME_24_SIMPLE instead of using AM/PM...

They still provide the `toLocaleString` method but there is an issue with their typings as using `.toLocaleString({ dateStyle: 'short', })` throws an error while typescript is "technically" allowing it. 
> Seems like: type DateTimeFormatOptions = Intl.DateTimeFormatOptions
> But the [Intl API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat) handles way more properties than [Luxon](https://github.com/moment/luxon/blob/master/docs/formatting.md#intldatetimeformat)
> So when using "unknown" attributes, we get "Invalid DateTime" errors.

**Hence I've created two new preset : `DATE_MONTH_YEAR` and `DATE_MED_SHORT_YEAR` (2.) and I've created the `getTimezoneString` function (1.)**

## Description

- Improve `intlFormatDateTime` function
- Add tests 
- Update occurrences 

Usage: 
```typescript
const {date, time, timezone} = intlFormatDateTime('2025-04-18T13:41:39Z', {
  formatDate: DateFormat.DATE_HUGE,
  formatTime: TimeFormat.TIME_SIMPLE,
  formatTimezone: TimezoneFormat.TIMEZONE_OFFSET,
})

console.log({
  date: 'Saturday, April 19, 2025'
  time: '1:41 PM'
  timezone:  'GMT-7'
})
```

This way it will be easier to handle date.

### DIFF
--- a/src/core/timezone/utils.ts
+++ b/src/core/timezone/utils.ts
@@ -44,20 +44,6 @@ export const isSameDay = (a: DateTime, b: DateTime): boolean => {
   return a.hasSame(b, 'day') && a.hasSame(b, 'month') && a.hasSame(b, 'year')
 }
 
-/**
- * @deprecated Use `intlFormatDateTime` instead.
- */
-export const intlFormatDateToDateMed = (
-  date: string,
-  timezone: TimezoneEnum | null | undefined,
-  locale: LocaleEnum,
-) => {
-  return DateTime.fromISO(date, {
-    zone: getTimezoneConfig(timezone).name,
-    locale: locale,
-  }).toLocaleString(DateTime.DATE_MED)
-}
-
 export enum DateFormat {
   /** Apr 18, 2025 */
   DATE_MED = 'DATE_MED',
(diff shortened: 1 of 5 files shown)


src/core/timezone/utils.ts
```diff
@@ -52,36 +58,125 @@ export const intlFormatDateToDateMed = (
...
+          timeZoneName: timeZoneName,
+        })
+        .find((part) => part.type === 'timeZoneName')?.value || ''
+  } else {
+    const timezoneOffset = dateTime.offset / 60
+
+    timezoneString =
+      timezoneOffset === 0
+        ? 'UTC'
+        : `UTC${timezoneOffset > 0 ? `+${timezoneOffset}` : timezoneOffset}`
```
ansmonjol, line 145: Should the else condition add a `-` sign ?

ansmonjol (approved):

PR #2135, opened 2025-04-24
fix: export provider from new design-system package

keellyp:
## Context

As part of creating the new SDK package, we need to export the design system’s provider to ensure that all components rendered within it share the same theme context. 

If the provider is used outside the package, styling breaks—because components can’t correctly inherit styles from a different theme context. 

**In other words, the theme must be shared consistently from the same provider instance.**

## Description

- Extract the theme provider from the design system package.

- Temporarily wrap both the Material and design system providers in app.tsx during the migration phase (with the goal of removing the Material provider once migration is complete).

- Update peer dependencies to ensure compatibility and avoid conflicts.

- Migrate `<TabButton />`, which appears to be the last component not compatible with the new provider setup.


packages/design-system/src/style.css
```diff
@@ -36,9 +36,4 @@
   span.line-break-anywhere {
     @apply break-all;
   }
-
-  /* TODO: Might be able to delete after migration */
-  button {
-    @apply h-10 cursor-pointer rounded-xl border-0 px-3 py-[6px] font-sans outline-none active:outline-none;
```
ansmonjol, line 42: not sure I found it, can you point me where those rules are applied now? Or was it just unnecessary somehow?

ansmonjol: I can see they are replicated in the tab button but what about other buttons of the app?

ansmonjol (approved):

PR #2144, opened 2025-04-28
fix: improve copy to clipboard function

keellyp:
## Context

Attempt to fix copy to clipboard function that sometimes fails

## Description

Fixes issue-844

### DIFF
--- a/src/core/utils/copyToClipboard.ts
+++ b/src/core/utils/copyToClipboard.ts
@@ -7,6 +7,26 @@ const filterComment = (value: string) => {
     .join('\n')
 }
 
+const unsecuredCopyToClipboard = (text: string) => {
+  const textArea = document.createElement('textarea')
+
+  textArea.value = text.trim()
+  document.body.appendChild(textArea)
+  textArea.focus()
+  textArea.select()
+  try {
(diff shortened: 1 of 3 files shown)


src/core/utils/copyToClipboard.ts
```diff
@@ -16,10 +36,6 @@ export const copyToClipboard: (value: string, options?: { ignoreComment?: boolea
   try {
     navigator.clipboard.writeText(serializedValue)
   } catch {
-    addToast({
-      severity: 'danger',
-      translateKey: 'text_63a5ba11eb4e7e17ef88e9f0',
-    })
-    throw new Error('Browser running in non secure context')
+    unsecuredCopyToClipboard(serializedValue)
```
ansmonjol, line 39: I think we should still warn users about their unsecure context when calling `unsecuredCopyToClipboard`.
Lago is supposed to run in https context but probably throwing an error here is not necessary now.
maybe we should have 2 copies then?

Probably warning (not danger severity maybe?) about unsecure context and later another error toast not having access to clipboard would be ideal?

ansmonjol (approved):

PR #2152, opened 2025-04-29
fix: add babel to jest

keellyp:
## Context

Related to this issue https://github.com/getlago/lago-front/actions/runs/14728160893/job/41335728042?pr=2149 since the migration to the monorepo

The issue was caused by `lago-design-system` being an ESM Modules in the Jest Environment (that does not support it).

## Description

We switched from ts-jest to babel-jest for transpilation because it is more straightforward to setup the compilation from esm to commonjs.


ansmonjol (approved):
Interesting! 
Can ts-jest package be removed then?

PR #2166, opened 2025-05-05
bug: prevent sub display if parent is terminated

ansmonjol:
Make sure we don't displayed unnecessary children sub if the parent one is terminated. It happen when we perform a downgrade.
Occurs since we removed the sub filter by status 

Fixes ISSUE-866


ancorcruz (approved):
LGTM!

keellyp (approved):

PR #2177, opened 2025-05-09
feat: Add alert tab and part of the form

ansmonjol:
## Context

We're adding a new alerting system on top of subscriptions.

## Description

This PR adds some part of this feature, hidden behind a FE feature flag.
It contains the different accesses to the page, the list of alerts and the begining of the alert form.
It requires some BE work to be able to complete, but probably better to merge this as it's not reachable by customer.


src/pages/SubscriptionDetails.tsx
```diff
@@ -1,6 +1,6 @@
 import { gql } from '@apollo/client'
-import { Avatar } from 'lago-design-system'
-import { useRef } from 'react'
+import { Avatar, ButtonLink } from 'lago-design-system'
```
keellyp, line 2: 😍

PR #2192, opened 2025-05-14
feat: prepare design system for audit logs feature

keellyp:
## Context

For the audit logs feature, we need a bunch of new stuff in the design-system, this PR should introduce them


src/components/designSystem/VerticalMenu.tsx
```diff
@@ -48,34 +50,58 @@ export const VerticalMenu = ({
...
+                    {!!external && <Icon name="outside" />}
                   </div>
-                  {!!external && <Icon name="outside" />}
-                </div>
-              </ButtonLink>
-            )
+                </ButtonLink>
+              )
+            }
+            if (onAction) {
```
ansmonjol, line 83: Should it be an `else if` here? The type for props does not explicitly prevent to pass both and I don't think it would be expected

src/components/designSystem/NavigationTab.tsx
```diff
@@ -103,7 +110,7 @@ export const NavigationTab = ({
 
     // NOTE: window.location.pathname has to be watched for programatic navigation (without clicking on tabs)
```
ansmonjol, line 111: What about this comment then? Should it be updated and does this programatic url change still triggers the hook? I guess yes but just double checking as copy seems wrong now

packages/design-system/src/lib/muiTheme.ts
```diff
@@ -680,6 +685,33 @@ export const theme = createTheme({
               backgroundColor: alpha(palette.grey[700], 0.2),
             },
           },
+          '&.button-inline': {
+            position: 'relative',
```
ansmonjol, line 689: Does it really have to be relative?

src/public/icons/terminal.svg
```diff
@@ -0,0 +1,8 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ansmonjol, line 1: I think icon usually don't have width and height defined in our codebase, worth double check
i know there is an option in svgomg to remove them

packages/design-system/src/lib/muiTheme.ts
keellyp: Capture d’écran 2025-05-14 à 11 13 42

I think so, the focus-visible state should look like this

ansmonjol: Ah yes missed the absolute bellow when reviewing

ansmonjol (approved):

PR #2200, opened 2025-05-15
feat: new developer console

keellyp:
## Context

Related to the new developer experience, we're migrating existing views into a separate console

## Description

- Usage of `react-resizable-panels`
- Usage of <MemoryRouter/> so the console can have its own router without impacting the browser router one
- Migration of API keys, Webhooks and Debugger pages
- Introduce new single queries to get single webhook and single event 

Fixes LAGO-827

QA : https://lago-827-app.staging.getlago.com/


src/core/apolloClient/cacheUtils.ts
```diff
@@ -51,6 +52,10 @@ export const logOut = async (client: ApolloClient<object>, resetLocationHistory?
   // Clear store
   await client.cache.reset()
   updateAuthTokenVar()
+
+  // Close the devtools
+  removeItemFromLS(DEVTOOL_STORAGE_KEY)
```
ansmonjol, line 57: Just wondering
We also close the devtool on login here: https://github.com/getlago/lago-front/pull/2200/files#diff-319e2b0e2419e2e0741f6481be2c924f8158491f2add91e412122af742ff0366R44

Won't that be better to keep this LS devtool for having the same context when user logs in back? Or does that conflict with some logic and it could messup somehow?
Seems that key is only used for the open state but maybe it should hold more?
Just thinking as I type, just wanted to make sure there is no redundant logic here

ansmonjol (approved):

PR #2206, opened 2025-05-19
feat: release avalara integration feature

ansmonjol:
## Context

We're adding Avalara as a new tax provider.
This PR is the last piece of work needed to complete the integration.
This also removed the FE feature flag to show to feature to all customers

## Description

This PR does
- take care of displaying the Avalara link on the associated object (invoice or credit note) and displays the re-sync link if it failed.
- take care of the error handling for those tax provider. Some unused error codes have been removed and new ones added. I made sure it's easy to determine which tax provider emits a specific error
- Removes the FE feature flag so anyone can see the Avalara integration

Also fixes ISSUE-896


src/pages/CreateInvoice.tsx
```diff
@@ -1200,18 +1200,36 @@ const CreateInvoice = () => {
...
+                        hasDefinedGQLError('CurrencyCodeNotSupported', errors) ||
+                        // Avalara
+                        hasDefinedGQLError('InvalidEnumValue', errors)
+                      ) {
                         setTaxProviderTaxesErrorMessage(
                           LocalTaxProviderErrorsEnum.CurrencyCodeNotSupported,
                         )
                       } else if (
+                        // Anrok
                         hasDefinedGQLError('CustomerAddressCountryNotSupported', errors) ||
```
stephenlago99, line 1214: Calling `hasDefinedGQLError` this many times in nested if / else etc is really hard to debug. We should use functions and objects to make this clearer:

```
const ERRORS = {
  address: {
    errors: [ 'CustomerAddressCountryNotSupported', 'CustomerAddressCouldNotResolve'],
    callback: () => {
      setTax..
     }
  }
}
```

src/pages/CustomerInvoiceDetails.tsx
```diff
@@ -245,18 +258,36 @@ const getErrorMessageFromErrorDetails = (
...
+      errorDetails === LagoApiError.InvalidEnumValue
+    ) {
       return LocalTaxProviderErrorsEnum.CurrencyCodeNotSupported
     }
 
     if (
+      // Anrok
       errorDetails === LagoApiError.CustomerAddressCouldNotResolve ||
-      errorDetails === LagoApiError.CustomerAddressCountryNotSupported
+      errorDetails === LagoApiError.CustomerAddressCountryNotSupported ||
```
stephenlago99, line 273: We can replace these with

```
if ([LagoApiError.CustomerAddressCountryNotSupported, LagoApiError.MissingAddress].includes(errorDetails)) {
  ...
}
```

Maybe we could even re-use the same object from the first comment, that would be great

PR #2213, opened 2025-05-23
feat: add alerts sections to subscriptions

ansmonjol:
## Context

Users are willing to have a way to get when usage reach certain thresholds on a subscription, without having to constantly fetch the current usage.

## Description

This PR allows to create alerts on subscriptions.
We are adding 2 types of alerts today and 2 more are planned to be developed soon.

Fixes LAGO-833


src/components/alerts/Thresholds.tsx
```diff
@@ -0,0 +1,264 @@
...
+      recurring,
+    }
+
+    // Adds the new threshold at the correct place in the array
+    if (!recurring && hasRecurringThreshold) {
+      const recurringIndex = thresholds.findIndex((t) => t.recurring)
+
+      // If recurringIndex is -1 (not found) or 0, we should still push at the end
+      if (recurringIndex <= 0) {
+        thresholds.push(newThreshold)
```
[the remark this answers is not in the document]
ansmonjol: Good point. i'll also look into maybe refactoring cause I feel the 2 push may be turned into one.
Will explore

src/components/alerts/Thresholds.tsx
```diff
@@ -0,0 +1,264 @@
...
+            ]}
+          />
+        </div>
+      </div>
+      <Switch
+        name="recurring-threshold"
+        className="w-fit cursor-pointer"
+        data-test="add-new-recurring-threshold-switch"
+        checked={hasRecurringThreshold}
+        onChange={() => {
```
stephenlago99, line 182: We could extract this into a function, as it would make it easier to see the whole logic.

src/components/alerts/Thresholds.tsx
```diff
@@ -0,0 +1,264 @@
...
+        }}
+        label={translate('text_1724234174945ztq15pvmty3')}
+        subLabel={translate('text_172423417494563qf45qet2d')}
+      />
+      {hasRecurringThreshold && (
+        <div className="-mx-4 -my-1 overflow-auto px-4 py-1">
+          <ChargeTable
+            name={'progressive-billing-recurring'}
+            data={thresholds.filter((threshold) => threshold.recurring) || []}
+            columns={[
```
stephenlago99, line 201: The columns array seems 90% identical with the one above. Could we extract these to a helper function, so we won't have duplicated code? Looks like the main difference is how we set the index, we could do that based on a param.

src/components/alerts/Thresholds.tsx
ansmonjol: it's used just once, I'm not sure if I want to extract it honestly 🤔 What could be the advantages? The logic matters here and having it elsewhere could also make it hard to track when investigating later and looking at the component invocation directly no?

src/components/alerts/Thresholds.tsx
ansmonjol: hum probably yeah will look into that

src/pages/AlertForm.tsx
```diff
@@ -180,6 +271,62 @@ const AlertForm = () => {
...
+  }, [existingAlertsData?.alerts.collection])
+
+  const hasAnyNonRecurringThresholdError = useMemo(() => {
+    const localNonRecurringThresholds = formikProps.values.thresholds.filter(
+      (threshold) => !threshold.recurring,
+    )
+
+    return localNonRecurringThresholds.some(
+      (threshold, i) =>
+        i > 0 &&
```
stephenlago99, line 324: There's a comment above suggesting that this logic can be extracted to a helper, we could use that here.

src/pages/AlertForm.tsx
```diff
@@ -259,30 +408,47 @@ const AlertForm = () => {
...
                       </>
                     )}
+
+                    {showThresholdTable && (
+                      <AlertThresholds
+                        thresholds={formikProps.values.thresholds}
+                        setThresholds={(thresholds) => {
+                          formikProps.setFieldValue('thresholds', thresholds)
+                        }}
+                        setThresholdValue={({
```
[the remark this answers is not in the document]
ansmonjol: yup and maybe in a useCallback 
Will change

src/pages/AlertForm.tsx
ansmonjol: good catch!

ansmonjol: I won't extract `localNonRecurringThresholds` logic as the first one is ina. method outside of the component and I don't want to pass it as arg as it's for initialisation only, and I won't bring back this method in the component to prevent it to be re-calculated on re-render.

However the error logic (`i > 0 && ...`) can be mutualised between here and the Threshold component.
on it

stephenlago99 (approved):

PR #2216, opened 2025-05-26
bug(coupons): expirationAt should be set at end of day

ansmonjol:
## Context

When creating a coupon, you can set a date so it expires.

You can set this date to the current day (expected). However the BE checks the date is in the future.

When clicking on a date with the DatePicker, it set's it with the current datetime. Hence the date sent to the BE for today never valid.

## Description

This PR transforms the date set to be at the endOf day time. 

Fixes ISSUE-894

### DIFF
--- a/src/pages/CreateCoupon.tsx
+++ b/src/pages/CreateCoupon.tsx
@@ -28,6 +28,7 @@ import { WarningDialog, WarningDialogRef } from '~/components/WarningDialog'
 import { FORM_ERRORS_ENUM } from '~/core/constants/form'
 import { COUPON_DETAILS_ROUTE, COUPONS_ROUTE } from '~/core/router'
 import { deserializeAmount } from '~/core/serializers/serializeAmount'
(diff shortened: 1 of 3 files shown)


src/pages/CreateCoupon.tsx
```diff
@@ -454,13 +455,23 @@ const CreateCoupon = () => {
...
                           disablePast
+                          className="flex-1"
                           name="expirationAt"
                           placement="top-end"
                           placeholder={translate('text_632d68358f1fedc68eed3ea5')}
-                          formikProps={formikProps}
+                          onChange={(expirationAt) => {
+                            formikProps.setFieldValue(
+                              'expirationAt',
+                              expirationAt
```
stephenlago99, line 467: Is there a case in which this could be empty?

ansmonjol: If user `CMD+DELETE` yes

PR #2218, opened 2025-05-26
fix(console): set active row

keellyp:
## Description

- When using transactionId instead of eventId, i've forgotten a logic to set the active row that was still using eventId.
- I've also fixed the webhook active row that was having an issue on closing/opening the console

### DIFF
--- a/src/components/developers/webhooks/WebhookLogs.tsx
+++ b/src/components/developers/webhooks/WebhookLogs.tsx
@@ -129,14 +129,15 @@ export const WebhookLogs = () => {
       navigateToFirstLog(data?.webhooks?.collection)
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [data, logId])
+  }, [data?.webhooks.collection, logId])
 
   // The table should highlight the selected row when the logId is provided in params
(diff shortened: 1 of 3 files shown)


src/components/developers/webhooks/WebhookLogs.tsx
```diff
@@ -129,14 +129,15 @@ export const WebhookLogs = () => {
...
 
   // The table should highlight the selected row when the logId is provided in params
   useLayoutEffect(() => {
-    if (logId) {
-      logListRef.current?.setActiveRow(logId)
+    if (logId && logListRef.current) {
+      logListRef.current.setActiveRow(logId)
     }
-  }, [logId])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ansmonjol, line 139: Wondering why we disable them here and in the useEffect above. 

What about removing them and letting the deps list be what it expect to be?

keellyp: Weird behavior but the ref was null on mount and it didn't work if i didn't watch it

ansmonjol (approved):

PR #2224, opened 2025-05-28
feat: add fee type limitation on wallets form

ansmonjol:
## Context

Wallets credits are applied to all charges fee of an invoice. Which is not always convenient and customer neeeded to be able to have more granularity on their application.

## Description

This PR allow to define a `appliesTo` scope to a wallet. It also changes some UI elements and copies.

Fixes LAGO-851

### DIFF
--- a/src/pages/wallet/components/TopUpSection.tsx
+++ b/src/pages/wallet/components/TopUpSection.tsx
@@ -510,7 +510,7 @@ export const TopUpSection: FC<TopUpSectionProps> = ({
                 <Button
                   className="self-start"
                   startIcon="plus"
-                  variant="quaternary"
+                  variant="inline"
                   onClick={() =>
                     formikProps.setFieldValue('recurringTransactionRules.0.expirationAt', '')
                   }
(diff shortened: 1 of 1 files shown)


src/pages/wallet/components/SettingsSection.tsx
```diff
@@ -26,6 +33,18 @@ interface SettingsSectionProps {
   formType: keyof typeof FORM_TYPE_ENUM
 }
 
+type AvailableFeeTypes = {
+  [FeeTypesEnum.Charge]: FeeTypesEnum.Charge
+  [FeeTypesEnum.Commitment]: FeeTypesEnum.Commitment
+  [FeeTypesEnum.Subscription]: FeeTypesEnum.Subscription
+}
```
keellyp: Why not simple use this notation? 
```suggestion
type AvailableFeeTypes = FeeTypesEnum.Charge | FeeTypesEnum.Commitment | FeeTypesEnum.Subscription
```

keellyp (approved):

src/pages/wallet/components/SettingsSection.tsx
ansmonjol: Fair point 😄 
Will update, indeed that makes no sense

ansmonjol: Fixed

PR #2240, opened 2025-06-04
feat(billing-entities): Billing entity invoice custom sections

stephenlago99:
## Context

We need to have a new page on billing entities where the user can apply custom sections defined on the organization level.

## Description

- Added a new route for billing entity custom invoice sections
- Created the page and the apply / remove modals


src/hooks/useCreateEditBillingEntity.ts
```diff
@@ -43,6 +43,10 @@ gql`
     isDefault
     defaultCurrency
     euTaxManagement
+    selectedInvoiceCustomSections {
```
ansmonjol, line 46: Is there a reason we define the object shape here instead of importing a fragment from the component that uses the data?

ansmonjol (approved):
One comment otherwise lgtm

PR #2248, opened 2025-06-10
bug: fix issueCreditNote permission's condition

ansmonjol:
## Context

We were previously able to access the credit note (CN) form for a failed invoice, leading to an error on CN creation

## Description

This fixes the condition and unify the usage, so all conditions on the app uses this hook's method now.

Fixes ISSUE-918


src/hooks/usePermissionsInvoiceActions.ts
```diff
@@ -70,7 +70,7 @@ export const usePermissionsInvoiceActions = () => {
 
   const canIssueCreditNote = (invoice: Pick<Invoice, 'status'>): boolean => {
     return (
-      ![InvoiceStatusTypeEnum.Draft, InvoiceStatusTypeEnum.Voided].includes(invoice.status) &&
+      [InvoiceStatusTypeEnum.Draft, InvoiceStatusTypeEnum.Finalized].includes(invoice.status) &&
```
keellyp: Can we actually issue a credit note for a draft invoice?

keellyp (commented):
Just a tiny question

src/hooks/usePermissionsInvoiceActions.ts
ansmonjol: Fail point, BE allows it but we don't show the credit note table in such case.
I'll remove the draft and flag that issue to product and see how we can unify later

PR #2272, opened 2025-06-18
fix: no hyperlinks for deleted activity type

keellyp:
## Context

Related to: https://lago.sentry.io/issues/6581756497/?referrer=slack&notification_uuid=537ffab5-b092-431b-80ca-0d56fce68978&environment=production&alert_rule_id=14434893&alert_type=issue


ansmonjol (commented):
I assume user will face an error toast when receiving a 404 on those 2 queries?

Maybe we should then add a `context: { silentErrorCodes: [LagoApiError.NotFound] },` to those WDYT?

PR #2281, opened 2025-06-25
feat: API logs in developer console

keellyp:
## Context

This PR should introduce API log in the developer console

## Description

- It adds new filters
- New route in the console

Status: QA approved

### DIFF
--- a/src/components/designSystem/Status.tsx
+++ b/src/components/designSystem/Status.tsx
@@ -24,6 +24,7 @@ type StatusLabelSuccess =
   | 'available'
   | 'refunded'
   | 'delivered'
+  | number
 type StatusLabelWarning = 'failed'
 type StatusLabelOutline = 'draft'
 type StatusLabelDefault = 'downgrade' | 'scheduled' | 'pending' | 'toPay' | 'processing' | 'n/a'
@@ -37,6 +38,7 @@ type StatusLabelDanger =
   | 'overdue'
   | 'canceled'
   | 'failed'
+  | number
 
 type StatusLabelDisabled = 'voided'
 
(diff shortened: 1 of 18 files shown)


src/components/designSystem/Status.tsx
```diff
@@ -178,7 +180,7 @@ export const Status: FC<StatusProps> = ({
       data-test="status"
     >
       <Typography variant="captionHl" color={config.color} noWrap>
-        {translate(statusLabel, labelVariables ?? {})}
+        {typeof label === 'number' ? label : translate(statusLabel, labelVariables ?? {})}
```
ansmonjol: _Nitpicking_
I wonder if doing this could be a better alternative: 
`{typeof label === 'string' ? translate(statusLabel, labelVariables ?? {}) : label}`
So in that case, if later we want to update the type to accept a `ReactNode`, we won't have to update the template too.

src/components/developers/apiLogs/ApiLogs.tsx
```diff
@@ -0,0 +1,167 @@
...
+    (logCollection?: ApiLogItemFragment[], currentSearchParams?: URLSearchParams) => {
+      if (logCollection?.length) {
+        const firstLog = logCollection[0]
+
+        if (firstLog && getCurrentBreakpoint() !== 'sm') {
+          const path = generatePath(API_LOG_ROUTE, { logId: firstLog.requestId })
+          const query = currentSearchParams?.toString()
+          const search = query ? `?${query}` : ''
+
+          navigate(`${path}${search}`, {
```
ansmonjol: You can use the `search` attribute in the navigate options.
Then remove the logic handling the `?` 
Something like this should work
```
navigate(
    generatePath(API_LOG_ROUTE, { logId: firstLog.requestId }), 
    { 
        search: currentSearchParams?.toString()
        replace: true
     }
 )
 ```

ansmonjol (approved):

PR #2282, opened 2025-06-25
misc: rework navigation UI

ansmonjol:
## Context

Include relevant motivation and context.

## Description

Describe your changes in detail.

Fixes LAGO-XXX


packages/design-system/src/components/ButtonLink.tsx
```diff
@@ -28,7 +28,7 @@ export interface ButtonLinkTabProps extends ButtonLinkBaseProps {
   title?: string
```
keellyp, line 28: If you perform changes in the design-system package i think you should impact the component in the main app as well until we do the migration to the package to avoid unsynchronized components

keellyp (approved):

PR #2285, opened 2025-06-27
misc(Customer): add random external id generate button

ansmonjol:
Adding a button to generate a random UUID for the externalCustomerId field.

This is only visible on creation form.

Aim to improve the experience on this form, as it's the only field required.


ansmonjol:
Decided to take another approach.

PR #2292, opened 2025-06-30
fix: include search params when navigate to api logs

keellyp:
## Context

A bug was reported where navigating to audit log details causes the current filters to be lost

### DIFF
--- a/src/components/developers/apiLogs/ApiLogsTable.tsx
+++ b/src/components/developers/apiLogs/ApiLogsTable.tsx
@@ -1,6 +1,6 @@
 import { Typography } from 'lago-design-system'
 import { FC, RefObject, useMemo } from 'react'
-import { generatePath } from 'react-router-dom'
+import { generatePath, useSearchParams } from 'react-router-dom'
 
 import { InfiniteScroll, Status, StatusType, Table } from '~/components/designSystem'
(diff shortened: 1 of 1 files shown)


src/components/developers/apiLogs/ApiLogsTable.tsx
```diff
@@ -54,6 +55,15 @@ export const ApiLogsTable: FC<ApiLogsTableProps> = ({ getApiLogsResult, logListR
...
           }
 
+          const path = generatePath(API_LOG_ROUTE, {
+            logId: id,
+          })
+
+          const query = searchParams.toString()
+          const search = query ? `?${query}` : ''
+
+          return `${path}${search}`
```
ansmonjol: I guess we can do something like this?

```
const path = generatePath(API_LOG_ROUTE, { logId: id as string })
const search = searchParams.toString()
const fullPath = `${path}${search ? `?${search}` : ''}`

return fullPath
```

ansmonjol: Sad we have to return a string here 🤔 Did that on purpose in the past but using the `navigate` method would have been handy here

keellyp: Ahah yes at the beginning it was expecting a `function` but i don't remember why it changed

PR #2283, opened 2025-06-25
Adding pt_BR translation

fgmacedo:
## Context

Adding Brazilian Portuguese (pt_BR) translation.

PR #2302, opened 2025-07-02
feat: Allow user to change UI locale

fgmacedo:
## Context

As pointed out on this [thread](https://github.com/getlago/lago-front/pull/2283#issuecomment-3027863940), there's no current way to allow users change the UI locale to match their preference. I've added complete localization for `pt_BR`, but Brazilian users are unable to use it.

## Description

Allow user to change UI locale.

I've refactored the `LanguageSettingsButton` to allow being configured with an option array of locales, to reuse the component from the Invoice selection (all locales) to the UI language selection, current only with `en` and `pt_BR`. The code improved IMO, as it's more DRY.

https://github.com/user-attachments/assets/2cd9ed32-fe86-41a0-987e-3e21b00abf33

Best!


fgmacedo:
I was not able to run the e2e tests locally, but I watched the cypress videos the errors seems as unrelated...  can I do something on my side?  

Best regards!

ansmonjol:
Thank you @fgmacedo,
I'll check internally with the team if we're comfortable with supporting new locale for the product UI.
That could have impact on our workflow we may want to mitigate or ovoid for now.
I’ll keep you posted!

PS: My bad I should have spotted the whole translation diff on your previous PR that was hidden by default on Github 😓

ansmonjol:
Hey @fgmacedo,

Thanks for opening the PR and for the initiative. We really appreciate the effort and the intention to help improving Lago.

We discussed it internally, and we’ve decided not to move forward with the Brazilian Portuguese translation for now. 
The main concern is the long-term maintenance it would require. Supporting one new language also opens the door to others, and with our current release pace (we ship one to two big features per week), keeping all translations up-to-date would add a layer of complexity that’s difficult to sustain without slowing down delivery.

We totally get how a translated UI could help adoption internally, and it’s something we might revisit in the future. If we do, we’ll likely explore tooling that can automate part of the process to reduce the maintenance burden.

Thanks again for your understanding and support  🙏

src/pages/settings/BillingEntity/sections/general/BillingEntityGeneral.tsx
```diff
@@ -65,6 +66,8 @@ const BillingEntityGeneral = () => {
           <>
             <TimezoneBlock billingEntity={billingEntity as BillingEntity} />
 
+            <LocaleBlock />
```
fgmacedo, line 69: If you still want to merge to get the improvements in the translation and in the LanguageSettingsButton, only this line (and the import) needs to be removed and the UI will keep the same.

fgmacedo:
Hey @ansmonjol and team 👋

Thanks for the thoughtful reply — I completely understand your concerns about maintenance and delivery speed.

I was wondering if there might be a middle ground to explore here, where adding `pt_BR` (and potentially others translations) wouldn’t add any significant engineering overhead for Lago for new developments.

For example:
- Would the main blocker be the way the i18n library behaves today? From what I’ve seen, when there’s a missing token, the UI doesn’t fall back to `en`, and the raw key is shown instead. This behaviour enforces that all translations should be keept up to date or the UI is broken.
- If fallback to `en` were enabled, even partially translated localizations could already provide a lot of value to users, while ensuring untranslated strings don’t break the experience.
- This would also allow localization to be treated as a community-supported feature. Whenever a language falls too far behind, contributors could update it without requiring core team involvement.

This way, new features wouldn’t be blocked by missing translations, and users could still benefit from translated portions of the UI.

Do you think this approach could make sense for Lago? If so, I can work on this.

Thanks again for considering and for all the great work you’re doing 🚀

ansmonjol:
Hey @fgmacedo,

Indeed this approach makes total sense and reduces the "pressure" around the copy management along the way.
Those would be great additions :) 


BTW I've open [this bug](https://github.com/getlago/lago-front/issues/2420), related to your initial contribution adding the `pt_BR` locale. Happy to help, just hard to make it fit on my roadmap today.

ansmonjol:
Hey @fgmacedo,

This PR have been opened for a while and does not seem to move forward.

Also the concerns related to locale maintenance are still there and we don't plan to put effort on this soon.

I prefer to close this PR now and re-open later, with a wider scope (handling more locales)

PR #2316, opened 2025-07-08
feat: new subscription list page

keellyp:
## Context

Related to navigation revamp

## Description

- New subscription page
- New entry in the navigation menu

QA: https://navigation-entries-app.staging.getlago.com/


src/pages/SubscriptionsPage.tsx
```diff
@@ -0,0 +1,292 @@
...
+      variables: {
+        limit: 20,
+        ...filtersForSubscriptionQuery,
+      },
+    })
+
+  const { debouncedSearch, isLoading } = useDebouncedSearch(getSubscriptions, loading)
+
+  const subscriptions = data?.subscriptions.collection as Subscription[]
+  const hasSearchParams =
```
ansmonjol, line 113: I wonder how not check individually each params. If we add one search param later we may miss updating this condition maybe?
Could that be done differently?

keellyp: Yes i'll remove limit and page params instead 👍

PR #2317, opened 2025-07-08
bug: adding .dockerignore

Harbinder04:
## Context
Docker layering was missing, causing slower build times.
.dockerignore was also not present, leading to unnecessary files being copied.

## Description
Added .dockerignore and optimized Dockerfile layering to speed up image builds and CI workflows.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20-alpine as build
+FROM node:20-alpine AS build
 
 WORKDIR /app
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,30 @@
+# Node modules
+node_modules
+**/node_modules
+
+# Local dev config
+.vscode
+.cursor
+.env.template
+.github
+
+# Test & mock
+__mocks__
+cypress
+
+# Git and CI-related
+.git
+.gitignore
+.github
+
+# Build
+dist
+
+# Documentation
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -1,7 +1,10 @@
-FROM node:20-alpine as build
+FROM node:20-alpine AS build
 
 WORKDIR /app
 
+COPY package.json pnpm-lock.yaml tsconfig.json ./
+COPY scripts ./scripts
+COPY packages ./packages
```
ansmonjol: Why adding those when the `COPY . .` is present bellow?

Harbinder04: Even though COPY . . is present later, adding the specific COPY package.json pnpm-lock.yaml earlier helps Docker take advantage of layer caching.
If dependencies haven't changed, the pnpm install step can be reused from cache, making the build faster.

ansmonjol:
Just checked and the `Copy . .` takes 0.2 sec, hence having no impact on the build speed.

I would feel more comfortable discarding this change but the ignore is a great addition.

Would you mind updating this PR?
I'll have to re-open it on our side BTW, I'll keep you as author of the commit ofc

Harbinder04:
Sure I will update it.

ansmonjol:
Re-opened here: https://github.com/getlago/lago-front/pull/2366

PR #2344, opened 2025-07-17
fix: do not clean taxes in plan values

keellyp:
## Context

When updating a subscription, plan taxes were removed from the subscription

## Description

We removed this line so it’s passed along the `planValues` in the mutation

Fixes ISSUE-1002


ansmonjol (commented):
Is there an advantage on keeping 4 different tests called `when formType is...` then? As those all have the same behaviour we can probably merge the tests together WDYT?

keellyp:
Agree done !

PR #2347, opened 2025-07-17
feat: improve components ds

keellyp:
## Context

Related to sso-enforcement feature, we need to improve the design-system

## Description

- Ability to update the icon color in a chip
- Render div in the component is not clickable

### DIFF
--- a/packages/design-system/src/components/Selector.tsx
+++ b/packages/design-system/src/components/Selector.tsx
@@ -73,8 +73,10 @@ export const Selector = ({
   const [loading, setLoading] = useState(false)
   const clickable = !!onClick && !loading && !disabled
 
+  const Component = clickable ? 'button' : 'div'
+
   return (
-    <button
+    <Component
       disabled={disabled}
       tabIndex={clickable ? 0 : -1}
       className={tw(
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/Chip.tsx
```diff
@@ -80,7 +58,16 @@ export const Chip = ({
         },
         className,
       )}
-      icon={icon ? <Icon className="!m-0" name={icon} /> : undefined}
+      icon={
+        icon ? (
+          <Icon
+            className="!m-0"
+            name={icon}
+            size={iconSize ?? 'small'}
```
ansmonjol, line 66: `<Icon />` component's size default value is `medium`.
This change makes all Chip now have a small `Icon` by default.
Is that expected? Or maybe the `Chip` was always provided an `Icon` tsx element on invocation?

keellyp: Yes the icon chip was using small variant by default: https://github.com/getlago/lago-front/pull/2347/commits/80afc746eb161419b6bb72d1f01a4491e94f409e#diff-7f29e7e3b5288becd2d4e381cf10545c981f53e797a09a40c670c051b6cfc39cL235-L242

ansmonjol: glad you removed it from the theme then!

PR #2354, opened 2025-07-21
feat: sso-enforcement

keellyp:
## Context

Related to sso-enforcement project

## Description

- Update authentication pages 
- Update login flow

QA: approved


src/pages/settings/Authentication/Authentication.tsx
```diff
@@ -42,33 +57,231 @@ gql`
         }
       }
     }
-  }
 
-  ${AddOktaIntegrationDialogFragmentDoc}
-  ${DeleteOktaIntegrationDialogFragmentDoc}
+    ${AddOktaIntegrationDialogFragmentDoc}
+    ${DeleteOktaIntegrationDialogFragmentDoc}
+  }
```
ansmonjol, line 63: Is this a valid change? Thought fragment import should always be placed at the bottom of the `gql` definition.
Don't think it fails, and you would have notice, but let's try to keep the same convention in all files!

src/core/apolloClient/cacheUtils.ts
```diff
@@ -56,27 +64,96 @@ export const logOut = async (client: ApolloClient<object>, resetLocationHistory?
...
+    // If an error occurs, display a toast to inform the user that the login failed
+    addToast({
+      severity: 'danger',
+      translateKey: 'text_622f7a3dc32ce100c46a5154',
+    })
+
+    // Remove all local storage items related to the auth
+    removeItemFromLS(AUTH_TOKEN_LS_KEY)
+    removeItemFromLS(ORGANIZATION_LS_KEY_ID)
+    removeItemFromLS(LAST_PRIVATE_VISITED_ROUTE_WHILE_NOT_CONNECTED_LS_KEY)
```
ansmonjol: Do we want to remove `LAST_PRIVATE_VISITED_ROUTE_WHILE_NOT_CONNECTED_LS_KEY` then?
As a user I would love to still be redirected to the correct page even if I missed my first auth process. 

Can the data stored there have a wrong behaviour that brings the need to remove it during the login process?

src/layouts/SettingsNavLayout.tsx
```diff
@@ -67,7 +68,8 @@ const generateTabs = ({
   {
     title: translate('text_664c732c264d7eed1c74fd96'),
     link: AUTHENTICATION_ROUTE,
-    hidden: !hasPermissions(['organizationIntegrationsView']),
+    match: [OKTA_AUTHENTICATION_ROUTE],
```
ansmonjol: Guess this match array should also include `AUTHENTICATION_ROUTE` correct?

src/pages/Invitation.tsx
```diff
@@ -412,17 +422,18 @@ const Invitation = () => {
                       )
                     })}
                   </div>
-                ) : (
+                ) : !errorTranslation ? (
```
ansmonjol: Adding this condition adds a linter warning because of nested ternary I guess?

src/pages/auth/LoginOkta.tsx
```diff
@@ -138,7 +138,7 @@ const LoginOkta = () => {
               formikProps.touched.email && formikProps.errors.email
                 ? formikProps.errors.email
                 : errorField
-                  ? translate(getErrorKey(errorField))
+                  ? translate(getErrorKey(errorField) ?? '')
```
ansmonjol: `getErrorKey` seems to always return a string. Do we really need to fallback to empty string here?

ansmonjol (approved):
Few non critical comments

src/pages/settings/Authentication/Authentication.tsx
keellyp: Not sure to understand this comment, it's already placed at the bottom of the gql definition: https://github.com/getlago/lago-front/blob/e219f89d9896355c6d1ebe886fd03708940d02cd/src/pages/settings/Authentication/Authentication.tsx#L49-L64

ansmonjol: On the diff I can see it above the last closing `}` so I was wondering if that was intended. 
Diff could be messy too. Feel free to discard this comment if it fon't make sense!

PR #2379, opened 2025-08-05
misc: unify address city state and zipcode display

ansmonjol:
## Context

We've been noticed that the current format of addresses `{zipcode} {city} {state}` is wrong in english and should be `{city}, {zipcode} {state}`

Also, depending on locales we can have different display rules for it.

Same change should be implemented on BE side. Will wait for them before merging 

## Description

I took a composable approach where you define your format first, and the helper builds it correctly.

Was not really possible to get a method to format the other fields as they are often displayed with different orientation or typography.

Fixes ISSUE-975


keellyp (approved):

ansmonjol:
Closed in favour of https://github.com/getlago/lago-front/pull/2427

PR #2390, opened 2025-08-08
feat: define boundaries on one-off fees

ansmonjol:
## Context

We are allowing customers to set specific boundaries on fees when they create a one-off invoice. 

## Description

This pull request does two things:
1. It allows this definition through the invoice form
2. It removes some unused attributes from the query because it was too complex
I also ensured the type used in the transform function was using the fragment to make sure we are not missing any attributes when performing those transformations.

Fixes LAGO-945


src/components/invoices/EditFeeBillingPeriod.tsx
```diff
@@ -0,0 +1,150 @@
...
+          // Make sure value has correct format
+          if (!DateTime.fromISO(value).isValid) {
+            return this.createError({
+              path,
+              message: dateErrorCodes.wrongFormat,
+            })
+          }
+
+          // If subscription at is present
+          if (from && from[0] && from[0].value && from[0].value.fromDatetime) {
```
stephenlago99: We can reduce this to `from?.[0]?.value?.fromDatetime`

stephenlago99 (approved):

src/components/invoices/EditFeeBillingPeriod.tsx
ansmonjol:
```suggestion
          if (from?.[0]?.value?.fromDatetime) {
```

ansmonjol: Done

Issue #2407, opened 2025-08-15
[BUG] Combox re render makes edition of a plan clunky


bruno-guirra:
**Describe the bug**

When editing a charge on a plan, the application get slowed a lot if there are a large amount of charges in the plan. This is caused because each key stroke on the amount input component triggers a re-render of all the ComboBox components on the page. Other actions such as opening a ComboBox also trigger a re-render of all ComboBox components in the page.

**To Reproduce**

1. Add a large volume of metrics to a plan
2. Access the plan edit's page
3. Edit the amount of charge in the model (you should notice that it gets really slow)
4. If you want to debug how many renders are happening, a console.log can be added to the ComboBox component useEffect in this part:
```typescript
  useEffect(() => {
    prevRawDataRef.current = rawData
  })
``` 

**Expected behavior**
The UI shouldn't be fell slow while edition only one charge.

**Environment infos**

**App environment :** development
**API URL :** http://localhost:3000 
**Version :** 1.0.0
**User Id :** e59b50c1-9bd4-4cb4-8f9d-ad3835a7cc13

**Additional context**
Tested locally in lago-front project

![Image](https://github.com/user-attachments/assets/2e2969c4-8ce1-443f-ad4b-eec5dc8297fe)

keellyp:
Thank you for reporting, merged it in https://github.com/getlago/lago-front/pull/2411 🙏

PR #2408, opened 2025-08-15
fix(ComboBox): add dependency to useEffect for rawData updates

bruno-guirra:
## Context

During the edition of a plan, each key stroke in the amount input is trigger a re-render of the ComboBox components that encapsulate the charge model. In a plan with a large amount of charges it gets really clunky and this makes difficult to edit the amount value. I gave more details of the in the issue #2407. 

## Description

Added the rawData as dependency in the useEffect from ComboBox component, this prevents the re-renders.

### DIFF
--- a/src/components/form/ComboBox/ComboBox.tsx
+++ b/src/components/form/ComboBox/ComboBox.tsx
@@ -57,7 +57,7 @@ export const ComboBox = ({
 
   useEffect(() => {
     prevRawDataRef.current = rawData
-  })
+  }, [rawData])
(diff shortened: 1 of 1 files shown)


ansmonjol:
Thanks @bruno-guirra, we'll have a look once we have time!

keellyp:
We had to create a new PR because our CI is failing for external contrib.

Merged it in this PR https://github.com/getlago/lago-front/pull/2411.

Thank you for your contribution @bruno-guirra  🙏

PR #2413, opened 2025-08-20
feat: create multiple wallets with priority

ansmonjol:
## Context

Our users have the need to create multiple wallets per customer

## Description

This pull request allows customers to have multiple wallets. It also allow setting priority between wallets so they can be withdrawn in order depending on this value. 

Fixes LAGO-947


src/pages/wallet/CreateWalletTopUp.tsx
```diff
@@ -504,4 +499,30 @@ const CreateWalletTopUp = () => {
   )
 }
 
+const WalletSettingsInfosDisplay = ({
```
stephenlago99, line 585: Nice idea of creating a reusable component! Since we use it to display a list of labels, I wonder if we can name it something more generic. Or we can make this one accept { name, priority and rateAmount } as parameters and use it directly (if not, we would have to duplicate the `infos` array in every place where we use this.

ansmonjol: > I wonder if we can name it something more generic

You mean generic across the app?

> Or we can make this one accept { name, priority and rateAmount } as parameters and use it directly

Do you mean the props naming don't make much sense here?

I'm trying to understand better if you recommend a different approach so I can change my PR, or just commenting on the future possibilities here.
Let me know!

stephenlago99: So, the function currently renders a list of lines, it doesn't do anything specific to wallets (those lines can be anything). Hence, the naming can be generic.

But also, we could make it more re-usable in other parts of the application (if we wanted to show this info in another page/section) by making it accept a wallet as a parameter.

ansmonjol: > re-usable in other parts of the application

Are you fine if I create a ticket for this? Sounds like another initiative rather than a fix for this PR.
I would prefer to prio the feature addition and focus on this refactor later. 
I would love to create the component and migrate usages at the same time to make sure I didn't miss any cases for this component style in other parts of the app. Like those who renders cropped link or other situations we could have

src/components/customerPortal/wallet/WalletSection.tsx
```diff
@@ -113,108 +113,131 @@ const WalletSection = ({ viewWallet }: WalletSectionProps) => {
...
-        loading={isLoading}
+        loading={customerPortalUserLoading}
       />
 
-      {isLoading && <LoaderWalletSection />}
+      {isLoading && !wallets?.length && <LoaderWalletSection />}
+
+      {!!wallets?.length &&
+        wallets.map((wallet, index) => {
+          const [creditAmountUnit = '0', creditAmountCents = '00'] = String(
```
stephenlago99: This (to line 132) should be extracted to a function that returns [units, credits, balance] directly (based on isPremium). This way, we will have a testable & re-usable function.

src/components/customerPortal/wallet/WalletSection.tsx
```diff
@@ -113,108 +113,131 @@ const WalletSection = ({ viewWallet }: WalletSectionProps) => {
...
+                    </Typography>
+                  </div>
+                </div>
+
+                <div className="flex flex-col gap-1">
+                  <Typography variant="caption" color="grey600">
+                    {translate('text_1728377307160dqj0b2q59f6')}
+                  </Typography>
+
+                  {wallet?.expirationAt ? (
```
stephenlago99: We can simplify this to:

```
<Typography variant="body" color="grey700">
  {wallet?.expirationAt && (
    <>{intl...}</>
  )}
   
  {!walet?.expirationAt && (
    <>{translate(...)}</>
  )}
</Typography>
```

src/pages/wallet/CreateWalletTopUp.tsx
stephenlago99: Sounds good!

ansmonjol: Done

src/components/customerPortal/wallet/WalletSection.tsx
ansmonjol: Done

src/components/customerPortal/wallet/WalletSection.tsx
ansmonjol: Done

stephenlago99 (approved):

PR #2421, opened 2025-08-26
fix: generate payment link tab

keellyp:
## Context

On Safari, the copy to clipboard should follow a direct user interaction.
But in this case we're waiting for the BE to answer with the call. It's called transient user activation

## Description

When the url is fetched, open a new tab with the URL.
That will allow to manually copy the URL and fix the issue with Safari users

I've also add an error catching in case we're flagging an error with the PSP to display relevant toast message to the customer

Fixes issue-1044

### DIFF
--- a/translations/base.json
+++ b/translations/base.json
@@ -3420,7 +3420,6 @@
   "text_1755599398258j905gj9xihx": "Unlock projected usage",
   "text_1755599398258ce1ilgc5swg": "Unlock projected usage so customers can see their expected usage at the end of each period and plan for upcoming costs.",
   "text_1755599398258mj61iwjhhfk": "Request access to projected usage",
-  "text_1755599398258w59pin31rfe": "Hello, I would like to access your projected usageadd-on. \n\nPlease let me know if you need more info!",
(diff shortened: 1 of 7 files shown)


translations/base.json
```diff
@@ -3420,5 +3419,6 @@
   "text_1755599398258j905gj9xihx": "Unlock projected usage",
   "text_1755599398258ce1ilgc5swg": "Unlock projected usage so customers can see their expected usage at the end of each period and plan for upcoming costs.",
   "text_1755599398258mj61iwjhhfk": "Request access to projected usage",
-  "text_1755599398258w59pin31rfe": "Hello, I would like to access your projected usageadd-on. \n\nPlease let me know if you need more info!"
+  "text_1755599398258w59pin31rfe": "Hello, I would like to access your projected usageadd-on. \n\nPlease let me know if you need more info!",
```
ansmonjol: Not your change but let's fix it while we're around

```suggestion
  "text_1755599398258w59pin31rfe": "Hello, I would like to access your projected usage add-on. \n\nPlease let me know if you need more info!",
```

PR #2434, opened 2025-09-02
fix: undo dynamic width / move delete button in td

stephenlago99:
## Context

Reverts the change done in https://github.com/getlago/lago-front/pull/2415 and moves the delete button into the first cell.

## Description

The fluid width was breaking the table in some screens (progressive billing recurring thresholds, alert thresholds), because the first cell was being width-greedy. Reverted the change, and fixed the initial issue (the delete button making the table render and extra cell, thus making it jump around).

Fixes LAGO-1098

### DIFF
--- a/src/components/designSystem/Table/ChargeTable.tsx
+++ b/src/components/designSystem/Table/ChargeTable.tsx
@@ -92,15 +92,43 @@ export const ChargeTable = <T extends Record<string, unknown>>({
               data-test={`row-${i}`}
             >
               <>
-                {columns.map(({ content, mapKey }, j) => {
+                {columns.map(({ content, mapKey, size = 124 }, j) => {
+                  const sizeStyle = {
+                    width: `${size}px`,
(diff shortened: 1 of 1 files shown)


src/components/designSystem/Table/ChargeTable.tsx
```diff
@@ -92,15 +92,43 @@ export const ChargeTable = <T extends Record<string, unknown>>({
...
+                    width: `${size}px`,
+                    minWidth: `${size}px`,
+                    maxWidth: `${size}px`,
+                  }
                   return (
                     <td
                       className={tw(
                         CELL_HEIGHT,
-                        'w-full border-b border-r border-solid border-grey-300 p-0',
+                        'border-b border-r border-solid border-grey-300 p-0',
```
ansmonjol: This `td` need to get a `relative` class, otherwise some browsers won't make the assumption and check for upper parent with the relative position (e.g. Safari)

PR #2437, opened 2025-09-03
fix: Revenue streams - Show customers gross instead of net

stephenlago99:
## Context

We need to show the gross amounts instead of the net ones.

## Description

- Updated graphql query
- Updated rendering


keellyp (approved):

ansmonjol (approved):
Is it worth changing `text_17422232171950c2u9u3vuq7` copy to mention that this is a "Gross revenue" then?

PR #2444, opened 2025-09-08
feat(ai-agent): introduce new chat components

keellyp:
## Context

To prepare the ai-agent feature, this PR should introduce new UI components (not used yet in the app) for LLM outputs and displaying the new ai panel.

To display the LLM output, I used llm-ui lib: https://llm-ui.com/docs/concepts


src/components/aiAgent/ChatMessages.tsx
```diff
@@ -0,0 +1,128 @@
+import { Button, Icon, Typography } from 'lago-design-system'
+import { ReactNode } from 'react'
+
+/**
+ * Converts a number (in seconds) into a human-readable duration string
+ * @param seconds - The duration in seconds
+ * @returns A human-readable duration string (e.g., "2m 30s", "1h 15m", "45s")
+ */
+const formatDuration = (seconds: number): string => {
```
ansmonjol: Why not using `Intl.DurationFormat` API ?

If there is a real need of having that logic handmade, we will need a test for the result of this method

stephenlago99: Since we use `luxon`, could we re-use functionality from it instead of recreating it?

src/components/aiAgent/llmOutputs/hook.ts
```diff
@@ -0,0 +1,32 @@
...
+import { markdownLookBack } from '@llm-ui/markdown'
+import { throttleBasic, useLLMOutput } from '@llm-ui/react'
+
+import { Codeblock } from './Codeblock'
+import { Markdown } from './Markdown'
+
+export const useCustomLLMOutput = (output: string, isStreamFinished: boolean) => {
+  const { blockMatches } = useLLMOutput({
+    llmOutput: output,
+    // Default fallback to markdown component
```
stephenlago99: The code is pretty clear, we can remove this comment

src/components/aiAgent/llmOutputs/hook.ts
```diff
@@ -0,0 +1,32 @@
...
+    fallbackBlock: {
+      component: Markdown,
+      lookBack: markdownLookBack(),
+    },
+    blocks: [
+      {
+        // If the code block is found, use the codeblock component
```
stephenlago99: The code is pretty clear, we can remove this comment

src/components/aiAgent/ChatMessages.tsx
ansmonjol: Guess the DurationFormat api does not allow to transform `400s` into `6min 40s`, but at least once we separated the seconds into hours, minutes and seconds we could use this API to format depending on the locale.
Making later's app translations easier.

Pretty sure there should be something about this time split (`400s` into `6min 40s`) api somewhere, haven't dig

ansmonjol: Indeed Luxon seems to handle the 2nd point raised into my comment above. You can ignore mine then

stephenlago99 (approved):

src/components/aiAgent/ChatMessages.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { ReactNode } from 'react'
+
+/**
+ * Converts a number (in seconds) into a human-readable duration string
+ * @param seconds - The duration in seconds
+ * @returns A human-readable duration string (e.g., "2m 30s", "1h 15m", "45s")
+ */
+export const formatDuration = (seconds: number): string => {
+  const duration = Duration.fromObject({ seconds }).reconfigure({ locale: 'en-US' })
+  const shifted = duration.shiftTo('minutes', 'seconds').toObject()
```
ansmonjol, line 12: Just curious, what's the benefit of using `.shiftTo('minutes', 'seconds').toObject()` vs `.normalize().toObject()` ?

Also as it's an object returned you could instantiate `minutes` and `secondsStr` directly from this line

src/components/aiAgent/ChatMessages.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const minutes = shifted.minutes ? `${shifted.minutes}m` : ''
+  const secondsStr = shifted.seconds !== undefined ? `${shifted.seconds}s` : ''
+
+  return [minutes, secondsStr].filter(Boolean).join(' ')
```
ansmonjol, line 17: Again, I would change this to rely on the `Intl.DurationFormat`.

I'm not a big fan of manually manipulating strings for those kind of display.

Using this API also helps passing data from the Luxon's `Duration` result.
It also handle returning an empty string if no data is present.
Probably something like this in the end:
```
const locale = 'en' // Can later be changed to user's locale, when app will be translated

const durationObject = Duration
    .fromObject({ seconds })
    .reconfigure({ locale }) // I assume locale can be 'en' instead of 'en-US'
    .normalize()
    .toObject()

return new Intl.DurationFormat(locale, { style: "narrow" }).format(durationObject)
```
Worth the try!

ansmonjol: The only diff is that minutes will be displayed with `min` instead of `m` but probably something not deal breaker to check with product

src/components/aiAgent/ChatMessages.tsx
```diff
@@ -0,0 +1,114 @@
+import { Button, Icon, Typography } from 'lago-design-system'
+import { Duration } from 'luxon'
+import { ReactNode } from 'react'
+
+/**
+ * Converts a number (in seconds) into a human-readable duration string
+ * @param seconds - The duration in seconds
+ * @returns A human-readable duration string (e.g., "2m 30s", "1h 15m", "45s")
+ */
+export const formatDuration = (seconds: number): string => {
```
ansmonjol, line 10: Maybe this helper can be placed in a more "general" util file rather than in one related to aiAgent

ansmonjol: And have tests

ansmonjol:
```suggestion
export const formatSecondsToReadableDuration = (seconds: number): string => {
```

PR #2448, opened 2025-09-09
feat(ai-agent): support graphql subscriptions

keellyp:
## Context

Related to ai-agent feature.

This PR should update the instantiation of graphql to handle subscriptions.

## Description

I added actioncable because that's the protocol we're using on the backend side.
I've also split the `init` file to extract links for readability. 

As subscriptions are not used in the app yet, this can be safely merged

### DIFF
--- a/src/core/apolloClient/init.ts
+++ b/src/core/apolloClient/init.ts
@@ -9,7 +9,6 @@ import {
   timeoutLink,
   uploadLink,
 } from '~/core/apolloClient/links'
-// IMPORTANT: Keep reactiveVars import before cacheUtils
 import { envGlobalVar } from '~/core/apolloClient/reactiveVars'
 
 import { cache } from './cache'
@@ -72,8 +71,6 @@ export const initializeApolloClient = async () => {
   })
 
   globalApolloClient = client
-
-  client.setLink(createLinkChain(client))
 
   return client
 }


src/core/apolloClient/links.ts
```diff
@@ -0,0 +1,165 @@
...
+export const uploadLink = createUploadLink({
+  uri: `${apiUrl}/graphql`,
+}) as unknown as ApolloLink
+
+const TIMEOUT = 300000 // 5 minutes timeout
+
+export const timeoutLink = new ApolloLinkTimeout(TIMEOUT)
+
+export const subscriptionLink = new ActionCableLink({
+  cable: ActionCable.createConsumer('wss://api.lago.dev/cable'),
```
ansmonjol: Does that work on non-local environments?

That needs to be an env variable, otherwise OSS users with custom setup won't have the feature working afaik

src/core/apolloClient/init.ts
```diff
@@ -147,5 +72,17 @@ export const initializeApolloClient = async () => {
...
+  // Error link should be placed after network operations to catch their errors
+  const finalLinks = ApolloLink.from([
+    initialLink,
+    timeoutLink,
+    errorLink(globalApolloClient),
+    splitLink,
+  ])
+
+  // Update the client's link with the error handling
+  client.setLink(finalLinks)
```
ansmonjol: Can we either setup `client` without links or set it up with the `errorLink` directly and remove the `finalLinks` logic.

Maybe related to the comment above `finalLinks` creation. But as  copilot pointed out, it's confusing like this don't you think?

src/core/apolloClient/links.ts
```diff
@@ -0,0 +1,173 @@
...
+
+export const timeoutLink = new ApolloLinkTimeout(TIMEOUT)
+
+// Robustly construct websocket URL from apiUrl
+const apiUrlObj = new URL(apiUrl)
+const wsProtocol = apiUrlObj.protocol === 'https:' ? 'wss:' : 'ws:'
+apiUrlObj.protocol = wsProtocol
+apiUrlObj.pathname = apiUrlObj.pathname.replace(/\/$/, '') // Remove trailing slash if present
+const websocketUrl = apiUrlObj.toString().replace(/\/$/, '') // Remove trailing slash if present
+const cableUrl = `${websocketUrl}/cable`
```
ansmonjol: Ca we extract this logic and test it? Not sure what it does exactly when reading it.
Also to make sure we don't break it later if we change/add something

src/core/apolloClient/init.ts
```diff
@@ -147,5 +74,7 @@ export const initializeApolloClient = async () => {
 
   globalApolloClient = client
 
+  client.setLink(createLinkChain(client))
```
ansmonjol: Is it an issue then if the `globalApolloClient` does not have the correct link?

Reading the change, I don't understand why the link is not initialised with the correct `ApolloLink.from` on line 57.
Does it need a variable declared later?

src/core/apolloClient/links.ts
```diff
@@ -0,0 +1,173 @@
...
+    // Handle network errors
+    if (networkError) {
+      // eslint-disable-next-line no-console
+      console.error('[Network error]:', networkError)
+
+      // Check if it's an authentication error (401, 403)
+      if ('statusCode' in networkError) {
+        const statusCode = (networkError as { statusCode: number }).statusCode
+
+        if (statusCode === 401 || statusCode === 403) {
```
ansmonjol: Does it make sense to rely on the `AUTH_ERRORS` as for the "classic" http calls?

src/core/apolloClient/links.ts
```diff
@@ -0,0 +1,173 @@
...
+          }
+          return
+        }
+      }
+
+      // Handle connection errors
+      if ('message' in networkError) {
+        const message = (networkError as { message: string }).message
+
+        if (message.includes('Network request failed') || message.includes('fetch')) {
```
ansmonjol: Do we have another way to catch this, instead of relying on a string value? Like a code?

src/core/apolloClient/links.ts
```diff
@@ -0,0 +1,173 @@
...
+                variables: operation.variables,
+                context: operation.getContext(),
+              },
+            })
+
+            addToast({
+              severity: 'danger',
+              translateKey: 'text_622f7a3dc32ce100c46a5154',
+            })
+          }
```
ansmonjol: Seems similar to above HTTP error handling. Is there a way to mutualise their usage?

src/core/apolloClient/init.ts
keellyp: I've updated it with initial links

src/core/apolloClient/init.ts
```diff
@@ -1,124 +1,49 @@
...
-} from '~/core/apolloClient/reactiveVars'
-import { ORGANIZATION_LS_KEY_ID } from '~/core/constants/localStorageKeys'
-import { LagoApiError } from '~/generated/graphql'
+  errorLink,
+  initialLink,
+  subscriptionLink,
+  timeoutLink,
+  uploadLink,
+} from '~/core/apolloClient/links'
+// IMPORTANT: Keep reactiveVars import before cacheUtils
```
ansmonjol: Is this comment still relevant for this file?

src/core/apolloClient/init.ts
```diff
@@ -147,5 +73,7 @@ export const initializeApolloClient = async () => {
 
   globalApolloClient = client
 
+  client.setLink(createLinkChain(client))
```
ansmonjol: I still don't get why we must re-set the links with the same value.
What's the benefit?

You mentioned something related to error handling before but that's related to the orders of the links in `createLinkChain` no?

ansmonjol: We probably don't need the `globalApolloClient` in the error link WDYT?

I feel that if the error is received, we should just logout and don't really case if the client have been set before?

keellyp: cf this comment : https://github.com/getlago/lago-front/pull/2448#discussion_r2359163909

src/core/apolloClient/init.ts
keellyp: Nop i'll remove it

src/core/apolloClient/init.ts
ansmonjol: Yep, so we probably don't need the globalApolloClient in the error link WDYT?

I feel that if the error is received, we should just logout and don't really case if the client have been set before?

That would solve the issue and prevent us to pass any props to the error link

keellyp: We need this `globalApolloClient` in the errorLink function because the logOut function needs it.
It might need a bigger refactor to clean this

PR #2453, opened 2025-09-15
misc(creditNotes): validate form has at least one fee checked

ansmonjol:
## Context

When creating a credit note, you can subit the form after unchecking all fees.

Recently, Backend made a bugfix to return an error un such case, but it's still allowed to submit the form like this and receive an error.

This form can accept 3 "shapes" of fee.
Classic ones for addOn and credit fees. And more "complex" ones that has a different level of "checked" attribute.

## Description

This PR creates a check to prevent submitting the form if no fees are selected.

Had to make a different function is used at the submit button level. As it's checking on different attributes of the form, it was creating a complex code there and was hard to test that way.
Preferred to extract the logic in the existing util and test the method there.

Fixes ISSUE-1108

### DIFF
--- a/src/components/creditNote/utils.ts
+++ b/src/components/creditNote/utils.ts
@@ -8,7 +8,7 @@ import {
   InvoiceTypeEnum,
 } from '~/generated/graphql'
 
-import { FeesPerInvoice, FromFee, GroupedFee } from './types'
+import { CreditNoteForm, FeesPerInvoice, FromFee, GroupedFee } from './types'
 
 export type CreditNoteFormCalculationCalculationProps = {
   currency: CurrencyEnum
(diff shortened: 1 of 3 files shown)


src/components/creditNote/utils.ts
```diff
@@ -160,3 +160,30 @@ export const createCreditNoteForInvoiceButtonProps = ({
...
+    })
+  } else if (!!creditFee?.length) {
+    return creditFee?.some((cf) => {
+      return cf.checked
+    })
+  } else if (!!fees && !!Object.values(fees).length) {
+    return Object.values(fees).some((fee) => {
+      return Object.values<FromFee | GroupedFee>(fee.fees).some((f) => {
+        if ('grouped' in f) {
+          return Object.values((f as GroupedFee)?.grouped || {}).some((g) => g.checked)
```
keellyp: This isn't easy to read, can we extract it into another function or at least leave a comment explaining what's happening?

keellyp (approved):

src/components/creditNote/utils.ts
ansmonjol: I can split them into explicit-named variable. Creating extra functions won't make it more clear IMO

I find the code explicit enough but I understand your point, will try to make it look better.

src/components/creditNote/utils.ts
```diff
@@ -160,3 +160,38 @@
...
+        }
+
+        // Fees of type FromFee
+        return (f as FromFee).checked
+      })
+    })
+  }
+
+  return false
+}
```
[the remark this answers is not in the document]
ansmonjol: Changed the code and added a test for this case

PR #2483, opened 2025-10-06
misc: restore WebSocket connection with Apollo

ansmonjol:
We reverted WS setup recently cause it was creating an auth bug.

This PR restores the WS setup after we've been improving the app auth and booting process

Fixes ISSUE-1175


package.json
```diff
@@ -115,13 +115,16 @@
     "@nangohq/frontend": "^0.67.8",
     "@sentry/react": "9.11.0",
     "@tanstack/react-virtual": "3.11.2",
+    "@types/actioncable": "^5.2.11",
```
AllanMichay: Should be in devDependencies

AllanMichay (commented):
I lack the full overview of the issue for now but I trust the reason for this change

package.json
ansmonjol: Indeed! Just moved it

src/core/apolloClient/init.ts
```diff
@@ -74,9 +90,9 @@ export const initializeApolloClient = async () => {
 
     if (graphQLErrors) {
       graphQLErrors.forEach((value) => {
-        const { message, path, locations, extensions } = value as LagoGQLError
+        const { message, path, locations, extensions } = value
 
-        const isUnauthorized = extensions && AUTH_ERRORS.includes(extensions?.code)
+        const isUnauthorized = extensions && AUTH_ERRORS.includes(extensions?.code as LagoApiError)
```
AllanMichay, line 117: Just a simple question for this (and a global). We use `as` here, is there a reason we do not use type guards? Like [here](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

ansmonjol: no reason, just laziness I would say. But that's definitely something we should avoir indeed.
This one might be a bit tricky tho, will check

AllanMichay: Seen together, no use right now, might now to deviate from using enum in the future

AllanMichay (approved):

PR #2502, opened 2025-10-09
bug: make sure the dev panel is closed on on orga switch

ansmonjol:
## Context

You can open the developer panel and then switch organization.

In such cases, the developer panel remains open, and the problem is its attempt to perform the call and sometimes with some ideas that are not available on the new organization. 
That results in the user receiving 404 errors and seeing an error toast. 
The user could also land into an organization where his permission doesn't allow him to see the developer panel. 

## Description

This pull request simply closes the developer panel whenever the user tries to switch organizations. 

Fixes ISSUE-1084

### DIFF
--- a/src/hooks/useDeveloperTool.tsx
+++ b/src/hooks/useDeveloperTool.tsx
@@ -80,6 +80,8 @@ export function useDeveloperTool(): DeveloperToolContextType {
   useEffect(() => {
     // On mounted, check the params from the URL
     checkParamsFromUrl()
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
   // Throw an error if the hook is used outside of the provider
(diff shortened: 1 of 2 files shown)


src/hooks/useDeveloperTool.tsx
```diff
@@ -80,6 +80,8 @@ export function useDeveloperTool(): DeveloperToolContextType {
   useEffect(() => {
     // On mounted, check the params from the URL
     checkParamsFromUrl()
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
AllanMichay, line 84: Feels unnecessary in this commit

ansmonjol: Indeed, could have put that in a different commit but felt right fixing to remove the linter warning right away!
I can split if you insist 😄

AllanMichay: No need 😄

PR #2503, opened 2025-10-09
misc: use cached contextual translations

ansmonjol:
## Context

We are having issues in the application where we are importing contextual translations. 
Those translations are different than the English one, and usually they are imported in some components or they are used within the whole Customer Portal side of the application. 

We face two issues:
1. When a component is rendered in a list, every time a new component is drawn on the page, the hook about loading the contextual translation will fire again, and you will face some kind of loading state even though those translations are already available somehow.
2. The second issue is about the customer portal that explicitly fetches the new local every time we navigate within the customer portal. That led to having some flag-key translations displayed, and it would be better to rely on the cache as the translation in the local is not supposed to change when you are navigating within the same customer portal instance. 

## Description

This pull request solves those two issues:
1. It fixes it by just caching them at the module level within the simple constant. So if that constant is fulfilled with the correct local, we'll just use the same values there, preventing the glitch in the UI when we display the same component multiple times.
2. Solved by simply changing the cache policy for the portal local query that fetches those translations to be displayed on the screen.

Fixes ISSUE-1145

### DIFF
--- a/src/hooks/core/useContextualLocale.ts
+++ b/src/hooks/core/useContextualLocale.ts
@@ -11,21 +11,28 @@ import {
 
 const { appEnv } = envGlobalVar()
 
+// Simple module-level cache for contextual translations
+const contextualTranslationCache = new Map<Locale, Translation>()
+
 type UseContextualLocale = (locale: Locale) => {
   translateWithContextualLocal: (key: string, data?: TranslateData, plural?: number) => string
 }
 
 export const useContextualLocale: UseContextualLocale = (locale) => {
-  const [translations, setTranslations] = useState<Translation>()
+  const [translations, setTranslations] = useState<Translation | undefined>(() =>
+    contextualTranslationCache.get(locale),
+  )
 
   useEffect(() => {
-    const updateTranslations = async () => {
-      const contextualTranslations = await getTranslations(locale)
-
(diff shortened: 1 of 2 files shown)


src/hooks/core/useContextualLocale.ts
```diff
@@ -11,21 +11,28 @@ import {
...
-      const contextualTranslations = await getTranslations(locale)
-
-      setTranslations(contextualTranslations)
+    if (contextualTranslationCache.has(locale)) {
+      setTranslations(contextualTranslationCache.get(locale))
+      return
     }
 
-    updateTranslations()
+    getTranslations(locale).then((loadedTranslations) => {
```
AllanMichay, line 32: Wouldn't getting translations after trying to affect it lead to not having translation if we have no caching policy?

ansmonjol: Not sure to understand 🤔

AllanMichay: It's because there is nothing to understand, I don't no why I thought there would be an issue here. Might have done the review before my coffee

LGTM 💅

ansmonjol: haha ok 🙌

PR #2505, opened 2025-10-09
misc(auth): use new token if provided by Backend

ansmonjol:
## Context

Our customers are "savagely" being kicked off the application after three hours of usage of their authentication token.
We do not have any renew token logic in the application. However, the backend does send us a new token in our responses header if the token only has one hour remaining of activities allowed. 

## Description

I'm just creating here a new Apollo link that is about to catch this token, and if it's ever present in the header, it will replace the current authentication token.

Fixes ISSUE-978

### DIFF
--- a/src/core/apolloClient/init.ts
+++ b/src/core/apolloClient/init.ts
@@ -13,6 +13,7 @@ import {
   CUSTOMER_PORTAL_TOKEN_LS_KEY,
   envGlobalVar,
   TMP_AUTH_TOKEN_LS_KEY,
+  updateAuthTokenVar,
 } from '~/core/apolloClient/reactiveVars'
 import { ORGANIZATION_LS_KEY_ID } from '~/core/constants/localStorageKeys'
 import { LagoApiError } from '~/generated/graphql'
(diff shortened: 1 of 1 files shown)


src/core/apolloClient/init.ts
```diff
@@ -57,6 +58,27 @@ export const initializeApolloClient = async () => {
...
+        if (newToken) {
+          const currentToken = getItemFromLS(AUTH_TOKEN_LS_KEY)
+
+          if (newToken && newToken !== currentToken) {
+            updateAuthTokenVar(newToken)
+          }
+        }
+      }
+
+      return response
```
AllanMichay, line 78: Is there a reason we are not using return first coding style? I've noticed it through out the application but wouldn't it be easier for us to maintain if we write it like this

```javascript
if (!httpResponse?.headers) {
  return response
}

const newToken = httpResponse.headers.get('X-Lago-Token')

if (!newToken) {
  return response
}

const currentToken = getItemFromLS(AUTH_TOKEN_LS_KEY)

if (newToken && newToken !== currentToken) {
  updateAuthTokenVar(newToken)
}

return response
```

Other than this question, everything seems in order

ansmonjol: Good question. Felt the same when writing it but I was not liking much having multiple `return response` and felt it was more verbose (which does not always mean it's bad tho)

So I would say it's more a "personal" interpretation in this case, however I agree that really returns should be preferred when it makes more sense and the code clearer

We do not enforce a strong coding style today and prefer to let us the "freedom" to write the code as long as it does not create bugs or side effect.
Felt it was at first something that could ease the review process but as the team grows maybe we should change our habbits!

PR #2516, opened 2025-10-10
misc: adjust renovabot config

ansmonjol:
Changing some options for renovabot after it ran few times.
- increase the `minimumReleaseAge` value from 3 to 7. 
We know that recently there were some malicious upgrade from some packages. Even tho they have been catch fast let's be safe here. It also let time to the community to report some bugs and see of the version is "really" working
- add `"pinVersions": true`
Makes sure the version in package.json is pined and does not allow patch to be installed without them to be referenced in the package.json. Feel giving more control and we'll have renovate update the version in the package.json anyway
- remove `bumpVersion` option
It has not the effect I expected. Thought it would bump the package.json of the upgraded package when it has a new version. Otherwise it does update the root version of our package.json. Not sure about our strategy there and we may need to consider other aspects before starting to often bump our version


renovate.json
```diff
@@ -4,9 +4,9 @@
   "prHourlyLimit": 10,
   "branchConcurrentLimit": 10,
   "prCreation": "immediate",
-  "minimumReleaseAge": "3 days",
+  "minimumReleaseAge": "7 days",
+  "pinVersions": true,
```
AllanMichay, line 8: Won't pinning version add too much noise in the PRs? Otherwise LGTM

ansmonjol: that will add an extra change, not an extra PR from what I understand

PR #2541, opened 2025-10-14
feat(invoicing): add e-invoicing and allow the download of correct documents

AllanMichay:
## Context

By 2026, e-invoicing will become mandatory in France. This PR aims to add this featureso we can be more futureproof.

Back End can be found [here](https://github.com/getlago/lago-api/pull/4123)

## Description

This PR adds
- A way to define if the billing entity can do e-invoicing
- A way to download Invoices as XML File
- A way to download Credit notes as XML File

I took the liberty to remove some ternary nesting just to make the code more readable and more easy to modify

Fixes [LAGO-961](https://linear.app/getlago/issue/LAGO-961/e-invoicing)

### DIFF
--- a/src/pages/PaymentDetails.tsx
+++ b/src/pages/PaymentDetails.tsx
@@ -182,7 +182,9 @@ const PaymentDetails = () => {
 
   const { canDownloadPaymentReceipts, downloadPaymentReceipts, downloadPaymentXmlReceipts } =
     useDownloadPaymentReceipts()
-  const canDownloadXmlFile = canDownloadPaymentReceipts && !!payment?.paymentReceipt?.xmlUrl
+  const canDownloadXmlFile =
+    canDownloadPaymentReceipts &&
+    (!!payment?.paymentReceipt?.xmlUrl || !!payment?.customer?.billingEntity?.einvoicing)
 
   const goToPreviousRoute = useCallback(
     () =>


src/components/creditNote/CreditNoteDetailsOverview.tsx
```diff
@@ -92,19 +96,68 @@ export const CreditNoteDetailsOverview: FC<CreditNoteDetailsOverviewProps> = ({
...
-              })
-            }}
-          >
-            {translate('text_637655cb50f04bf1c8379cf8')}
-          </Button>
-        )}
+        {!hasError &&
+          !loading &&
+          hasPermissions(['creditNotesView']) &&
+          !disablePdfGeneration &&
```
ansmonjol: Nitpicking: Could we make that condition into a constant that can be re-used in the block at the bottom too?

Having multiple lines like that makes it hard to spot the difference

src/hooks/useCreateEditBillingEntity.ts
```diff
@@ -52,6 +52,7 @@ gql`
       name
       code
     }
+    einvoicing
```
ansmonjol, line 55: Nitpicking and commenting knowing we tend not to do it all the time.

Could we instead use a fragment created at the place that needs this attribute and spread in this hook?

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,15 @@
+export const downloadFileFromURL = async (fileName?: string, url?: string | null) => {
```
ansmonjol: Could that be part of the `src/core/utils/downloadFiles.ts` file?

src/utils/fileUtils.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (!url) return
+
+  const response = await fetch(url)
+  const blob = await response.blob()
```
ansmonjol: maybe this download method should be part of a hook, so we can have a try-catch that displays an error toast?

AllanMichay: You're absolutely right, gong for it

src/hooks/useCreateEditBillingEntity.ts
AllanMichay: Hmmm I'm not entirely sold on this for this example. In general, I agree we should do this.

But this hook is used in only one file and this is already a fragment. I feel like moving this in the file that uses the hook would make unnecessary back and forth between the two files to understand the whole flow.

I'm open for sync discussion if you want, I think it would be easier to explain this way

src/components/creditNote/CreditNoteDetailsOverview.tsx
AllanMichay: Absolutely, on it 👍

src/hooks/useDownloadFile.ts
```diff
@@ -0,0 +1,35 @@
+import { addToast } from '~/core/apolloClient'
+
+export const useDownloadFile = () => {
```
ansmonjol, line 3: As a next step
Can we plan a refactor to centralise methods from this file and `src/core/utils/downloadFiles.ts`, and check why `openNewTab` there is exported? IMO it shouldn't 
We can create a Linear improvement or tackle right away

AllanMichay: I think this task is outside the scope of this feature but definitely we need to tackle it

AllanMichay: Actually, had to do it since we had CORS issue with our other method of downloading files

PR #2542, opened 2025-10-14
misc: prepare plan form for new charge type

ansmonjol:
## Context

I'm working to perform important change to the plan form, and the current code implementation is limiting us.

This PR is a preliminary work in order to make this part of the app more "scalable", and specifically the charges related parts.

I'm planning a followup "preliminary" PR but it's kinda nasty and will touch many files. Preferred having a first chunk for 

## Description

This Pr is better reviewed commit by commit, even if the last one does a lot.
Here is a description of each
- 38cdee80595d303c9f71576e6893001cac6018f7
This one make the fragment of some `Properties` attribute be relying on their inner type rather that on the `Properties` type directly.
That will ease the addition of a new charge type, that will use those inner type but from a different `Properties` type. 
Not a big deal and a simple change, but helps decoupling some logic in the future.
- e589423f0d03d6fb40813ef8be22d6d72b435465
Simply centralise a translation key mapping in an `util` file
- 6482292bd90253ae0f44c3bb1e7dbb3cf4250704
This commit has a lot of change.
It's main goal it to rename non-generic `Charge` files/constants/types to be prefixed with `Usage`.
As you may understand now it's because of the future introduction of the new Charge type.
I've also split the `ChargeAccordion` in there but realised too late it was hard to extract it into a different commit.
That's also with the intention to ease the addition of a different accordion logic for this new charge type.


src/hooks/plans/useChargeForm.tsx
```diff
@@ -5,20 +5,20 @@ import { BasicComboBoxData } from '~/components/form'
 import { AggregationTypeEnum, ChargeModelEnum } from '~/generated/graphql'
 import { useInternationalization } from '~/hooks/core/useInternationalization'
 
-export type TGetChargeModelComboboxDataProps = {
+export type TGetUsageChargeModelComboboxDataProps = {
```
AllanMichay, line 8: Just asking and not blocker but is it normal that those types start with the `T` but there are some other types in the app that doesn't? 

I think it's a good timing to discuss some coding style and start cleaning our codebase to have something more in line with the coding style

src/hooks/plans/usePlanForm.tsx
```diff
@@ -474,8 +478,9 @@ export const usePlanForm: ({
   useEffect(() => {
     if (errorCode === FORM_ERRORS_ENUM.existingCode) {
       formikProps.setFieldError('code', 'text_632a2d437e341dcc76817556')
-      scrollToTop()
+      scrollToTop('#root')
```
AllanMichay: Any reason we have to precise this?

ansmonjol: Very good catch, rebase mess. Will fix

src/hooks/plans/useChargeForm.tsx
ansmonjol: Totally agree

PR #2586, opened 2025-10-20
fix: Fix icon alignment in Plan -> Unlock Entitlements section

domenicofalco:
## Context

The icon in Plan → Unlock Entitlements is misaligned with the title.

## Description

This PR fixes the icon alignment issue using CSS adjustments.
I've also added a typescript configuration to better parse typescript-jsx syntax

Fixes ISSUE-1220

### DIFF
--- a/src/components/plans/FeatureEntitlementSection.tsx
+++ b/src/components/plans/FeatureEntitlementSection.tsx
@@ -123,13 +123,12 @@ export const FeatureEntitlementSection: FC<FeatureEntitlementSectionProps> = ({
       {!isPremium && (
         <div className="flex items-center justify-between gap-4 rounded-lg bg-grey-100 px-6 py-4">
           <Box>
-            <Typography
-              className="flex items-center justify-start"
-              variant="bodyHl"
-              color="textSecondary"
-            >
-              {translate('text_1753864223060kmmxk4a59aj')} <Icon className="ml-2" name="sparkles" />
-            </Typography>
+            <div className="flex items-center justify-start gap-2">
+              <Typography variant="bodyHl" color="textSecondary">
+                {translate('text_1753864223060kmmxk4a59aj')}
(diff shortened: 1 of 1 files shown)


ansmonjol (commented):
You could replace the `Box` with a regular div using Tailwind’s `flex` and `gap` classes.

Saying as we might get rid of MUI later, and `Box` comes from their exported components.

Also, small note, there’s a potential risk of altering the spark SVG color here. 
It works now because the color is hardcoded, but that’s not usually the case for our app’s SVGs 🙂

good news for the tsconfig!

src/components/plans/FeatureEntitlementSection.tsx
```diff
@@ -123,8 +123,12 @@ export const FeatureEntitlementSection: FC<FeatureEntitlementSectionProps> = ({
...
         <div className="flex items-center justify-between gap-4 rounded-lg bg-grey-100 px-6 py-4">
           <Box>
-            <Typography variant="bodyHl" color="textSecondary">
-              {translate('text_1753864223060kmmxk4a59aj')} <Icon name="sparkles" />
+            <Typography
+              className="flex items-center justify-start"
+              variant="bodyHl"
+              color="textSecondary"
+            >
+              {translate('text_1753864223060kmmxk4a59aj')} <Icon className="ml-2" name="sparkles" />
```
AllanMichay: What do we think of putting the Icon outside of the Typography? 
Doing this would mean that we lose our color fallback but in the mean time, we could do something like

```suggestion
<div className="flex items-center justify-start gap-2">
  <Typography variant="bodyHl" color="textSecondary">
  {translate('text_1753864223060kmmxk4a59aj')}
  </Typography>
  <Icon name="sparkles" />
</div>
```

AllanMichay (approved):
LGTM 💅

PR #2599, opened 2025-10-22
bug: fix array properties validations

ansmonjol:
## Context

While working on the fixed charge addition, I noticed that our existing validations are buggy and that they are not really testing the validation schema for the charges. 

## Description

What happened is that the property checked was not the good one and was checking the singular attribute name instead of the plural one, which is used in the application. 
It was also asserting successful test as falsy, which makes the test completely useless. 

These pull requests ensure that all the cases are tested and that the existing tests are passing with the new rules. There were also some gaps in the validations, and I tried to make them more explicit within the charge schema validation. 

I also refactored this validation file to centralize the helper functions, making the code clearer and easier to read.

### DIFF
--- a/src/formValidation/chargeSchema.ts
+++ b/src/formValidation/chargeSchema.ts
@@ -2,7 +2,71 @@ import { array, number, object, string } from 'yup'
 
 import { LocalChargeFilterInput, LocalPricingUnitType } from '~/components/plans/types'
 import { MIN_AMOUNT_SHOULD_BE_LOWER_THAN_MAX_ERROR } from '~/core/constants/form'
-import { BillableMetric, ChargeModelEnum, Properties } from '~/generated/graphql'
+import {
+  BillableMetric,
+  ChargeModelEnum,
+  GraduatedPercentageRangeInput,
+  GraduatedRangeInput,
+  Properties,
+  VolumeRangeInput,
+} from '~/generated/graphql'
+
+// Helper function to validate fromValue is less than toValue
+const validateFromToValues = (
+  fromValue: string | number | undefined,
+  toValue: string | number | undefined,
+  isLastRange: boolean,
+): boolean => {
+  // For the last range, toValue can be empty/undefined (meaning infinity)
+  if (isLastRange && (toValue === undefined || toValue === null || toValue === '')) {
+    return true
+  }
+
+  return Number(fromValue || 0) < Number(toValue || 0)
+}
+
+// Helper function to validate perUnitAmount and flatAmount
+const validateRangeAmounts = (perUnitAmount: string, flatAmount: string): boolean => {
+  // Check if both perUnitAmount and flatAmount are invalid
+  if (isNaN(Number(perUnitAmount)) && isNaN(Number(flatAmount))) {
+    return false
+  }
+
+  // Check if either perUnitAmount or flatAmount is invalid (but not empty/undefined)
+  if (
+    (perUnitAmount !== undefined && perUnitAmount !== '' && isNaN(Number(perUnitAmount))) ||
+    (flatAmount !== undefined && flatAmount !== '' && isNaN(Number(flatAmount)))
+  ) {
+    return false
+  }
+
+  return true
+}
+
+// Helper function to validate rate field
+const validateRate = (rate: string | number | null | undefined): boolean => {
+  return !(isNaN(Number(rate)) || rate === '' || rate === null)
+}
+
+// Helper function to create chargeModel conditional check
+const isChargeModel = (chargeModel: ChargeModelEnum, targetModel: ChargeModelEnum): boolean => {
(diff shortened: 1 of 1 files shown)


src/formValidation/chargeSchema.ts
```diff
@@ -4,6 +4,66 @@ import { LocalChargeFilterInput, LocalPricingUnitType } from '~/components/plans
 import { MIN_AMOUNT_SHOULD_BE_LOWER_THAN_MAX_ERROR } from '~/core/constants/form'
 import { BillableMetric, ChargeModelEnum, Properties } from '~/generated/graphql'
 
+// Helper function to validate fromValue is less than toValue
+const validateFromToValues = (
+  fromValue: string | number | undefined,
```
domenicofalco, line 16: should it be more strict in accepting only number or undefined (no string)? eventually it evaluates only numbers and we may want it to be very explicit thanks to typescript. this way we have less ambiguity and we are force to use it that way.

src/formValidation/chargeSchema.ts
```diff
@@ -4,6 +4,66 @@ import { LocalChargeFilterInput, LocalPricingUnitType } from '~/components/plans
...
+  toValue: string | number | undefined,
+  isLastRange: boolean,
+): boolean => {
+  // For the last range, toValue can be empty/undefined (meaning infinity)
+  if (isLastRange && (toValue === undefined || toValue === null || toValue === '')) {
```
domenicofalco, line 21: i feel that we do not need the extra param **isLastRange** but rather `toValue !== 0 && !toValue` may by sufficient? this is how I see the new params types

`const validateFromToValues = (
  fromValue: number,
  toValue: number | undefined // indeed toValue can be empty/undefined (infinity)
)`

src/formValidation/chargeSchema.ts
```diff
@@ -4,6 +4,66 @@ import { LocalChargeFilterInput, LocalPricingUnitType } from '~/components/plans
...
+    return true
+  }
+
+  return Number(fromValue || 0) < Number(toValue || 0)
```
domenicofalco, line 25: if we strict the params to only accept numbers we do not need the casting here

src/formValidation/chargeSchema.ts
```diff
@@ -4,6 +4,66 @@ import { LocalChargeFilterInput, LocalPricingUnitType } from '~/components/plans
...
+}
+
+// Helper function to validate perUnitAmount and flatAmount
+const validateRangeAmounts = (
```
domenicofalco: can we enforce to have `number` type here as well so to best leverage typescript? else it feels like falling back to javascript dynamic types

src/formValidation/chargeSchema.ts
ansmonjol: I would love to but I see 2 limitations to this
- We do not have anyh type for this data in the validation. I'll check if I can cast a type to those yup validaitons
- The inputs in the app can return string or number if i'm correct 🤔 While this is an issue, I would prefer to now touch this sensible part as it would have a wide impact on the app 😓 

That being said, I agree the code is not ideal and I'll look at making it better!
Thanks for your comment

src/formValidation/chargeSchema.ts
ansmonjol: Very good point! I'll check how we define the last row's `toValue` but it's probably `undefined` all the time.
That would make the method easier but I wonder if that would make it more explicit.
Will dig into this

src/formValidation/chargeSchema.ts
ansmonjol: I can indeed improve it slightly.
I need to cast the type of the data manually so it's better to handle data within it.
However I noticed that some use the bigInt type and it does use `any` for it. So not all attributes will be well checked.

I'll create an improvement for me to configure the codegen to properly type those when running it, solving those kind of issues later 🔥

ansmonjol: So the method you pointer could not be immediatly fixed, however `validateRangeAmounts` can!

src/formValidation/chargeSchema.ts
ansmonjol: The validation needs to run only for all range _except_ last range, so we need to keep the `isLastRange` logic 

Digged a bit and feels that trying to improve it here would not make that much sense honestly 🤔

src/formValidation/chargeSchema.ts
ansmonjol: Did but not for this method. c.f. discussion above about bigInt typing

src/formValidation/chargeSchema.ts
ansmonjol: The type defined in the model is string here so I'll keep them as string params and transform internally rather than during invocation

domenicofalco (approved):

PR #2606, opened 2025-10-23
misc: minor changes to prepare fixed charges

ansmonjol:
## Context

I’m still preparing the codebase to integrate the new fixed charge into the plans. It’s a big change, so I’m gradually introducing small, non-breaking improvements to set things up. The idea is to reorganize parts of the app now to reduce the scope and risk of the upcoming large pull request.

## Description

This pull request implements few small changes, and are better to review commit-by-commit. 

There will be more to come, I'm trying to split the code as much as possible and incrementally add them to the codebase.

### DIFF
--- a/src/components/plans/chargeAccordion/ChargeWrapperSwitch.tsx
+++ b/src/components/plans/chargeAccordion/ChargeWrapperSwitch.tsx
@@ -15,8 +15,18 @@ import { StandardCharge } from '~/components/plans/StandardCharge'
 import { LocalChargeFilterInput, PlanFormInput } from '~/components/plans/types'
 import { VolumeChargeTable } from '~/components/plans/VolumeChargeTable'
 import { PremiumWarningDialogRef } from '~/components/PremiumWarningDialog'
-import { CurrencyEnum, PropertiesInput } from '~/generated/graphql'
+import {
+  ChargeModelEnum,
+  CurrencyEnum,
+  FixedChargeChargeModelEnum,
+  PropertiesInput,
+} from '~/generated/graphql'
 import { useInternationalization } from '~/hooks/core/useInternationalization'
+
+const CHARGE_MODEL = {
+  ...ChargeModelEnum,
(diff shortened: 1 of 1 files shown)


src/components/plans/chargeAccordion/ChargeWrapperSwitch.tsx
```diff
@@ -146,7 +148,7 @@ export const ChargeWrapperSwitch = memo(
             valuePointer={valuePointer}
           />
         )}
-        {localCharge?.chargeModel === ChargeModelEnum.Dynamic && <DynamicCharge />}
+        {localCharge?.chargeModel === 'dynamic' && <DynamicCharge />}
```
AllanMichay: I don't understand, why did we stop using the enum here?
I'd at least create a const for this. Let's avoid "magic strings / numbers"

AllanMichay (changes requested):

src/components/plans/chargeAccordion/ChargeWrapperSwitch.tsx
ansmonjol: > I don't understand, why did we stop using the enum here?

The main reason is because I'm soon gonna add a check on a new `FixedChargeModelEnum`
And as the value is the same, I felt it was ok to use it directly here

> I'd at least create a const for this. Let's avoid "magic strings / numbers"

I felt the ideal situation was about using the values directly, to prevent importing enums (and relying on them).
I has not been defined in out yet-to-be-written coding styleguide but I indeed felt it could be a local const based on both enum's keys and values to check it here 🤔 
I'll give it a shot

ansmonjol: I gave it a try in the last commit @AllanMichay and I love the approach! 
Let me know what you think about it

AllanMichay: That is way better and more readable, thanks!

PR #2635, opened 2025-10-26
misc: fix some linter warnings

ansmonjol:
Removed some linter warning we had in the app

I tried as much as possible to preserve the logic to be equally the same as before the refactoring. Otherwise, I took some time and the liberty to refactor a bit. 

When to each places to make sure it's not breaking anything

### DIFF
--- a/src/hooks/plans/useVolumeChargeForm.ts
+++ b/src/hooks/plans/useVolumeChargeForm.ts
@@ -6,6 +6,7 @@ import { ChargeCursor } from '~/components/plans/chargeAccordion/ChargeWrapperSw
 import { LocalChargeFilterInput, PlanFormInput } from '~/components/plans/types'
 import { ONE_TIER_EXAMPLE_UNITS } from '~/core/constants/form'
 import { PropertiesInput, VolumeRangeInput } from '~/generated/graphql'
+import { formataAnyToValueForChargeFormArrays } from '~/hooks/plans/utils'
 
 export const DEFAULT_VOLUME_CHARGES = [
   {
@@ -137,20 +138,12 @@ export const useVolumeChargeForm: UseVolumeChargeForm = ({
             // fromValue should always be toValueOfPreviousRange + 1
             const { toValue } = acc[i - 1]
             const fromValue = String(Number(toValue || 0) + 1)
-            const formattedToValue = () => {
-              if (range.toValue === null) return null
-
-              if (Number(range.toValue || 0) <= Number(fromValue)) {
-                return String(Number(fromValue) + 1)
-              }
-
-              return String(range.toValue || 0)
-            }
+            const formattedToValue = formataAnyToValueForChargeFormArrays(range.toValue, fromValue)
 
             acc.push({
               ...range,
               fromValue,
-              toValue: formattedToValue(),
+              toValue: formattedToValue,
             })
           } else {
             acc.push(range)
--- a/src/hooks/plans/utils.ts
+++ b/src/hooks/plans/utils.ts
@@ -1,0 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const formataAnyToValueForChargeFormArrays = (toValue: any, fromValue: string) => {
+  if (toValue === null) return null
+
+  if (Number(toValue || 0) <= Number(fromValue)) {
+    return String(Number(fromValue) + 1)
+  }
+
+  return String(toValue || 0)
+}
--- a/src/hooks/plans/__tests__/utils.test.ts
+++ b/src/hooks/plans/__tests__/utils.test.ts
@@ -1,0 +1,73 @@
+import { formataAnyToValueForChargeFormArrays } from '../utils'
+
+describe('formattedToValue', () => {
+  describe('when toValue is null', () => {
+    it('returns null', () => {
+      expect(formataAnyToValueForChargeFormArrays(null, '10')).toBeNull()
+    })
+  })
+
+  describe('when toValue is less than fromValue', () => {
+    it('returns fromValue + 1 as a string', () => {
+      expect(formataAnyToValueForChargeFormArrays(5, '10')).toBe('11')
+    })
+
+    it('handles string numbers', () => {
+      expect(formataAnyToValueForChargeFormArrays('5', '10')).toBe('11')
+    })
+  })
+
+  describe('when toValue equals fromValue', () => {
+    it('returns fromValue + 1 as a string', () => {
+      expect(formataAnyToValueForChargeFormArrays(10, '10')).toBe('11')
+    })
+
+    it('handles string numbers', () => {
+      expect(formataAnyToValueForChargeFormArrays('10', '10')).toBe('11')
+    })
+  })
+
+  describe('when toValue is greater than fromValue', () => {
+    it('returns toValue as a string', () => {
+      expect(formataAnyToValueForChargeFormArrays(15, '10')).toBe('15')
+    })
+
+    it('handles string numbers', () => {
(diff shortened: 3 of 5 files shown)


src/components/settings/integrations/AnrokIntegrationMapItemDialog.tsx
```diff
@@ -116,12 +116,15 @@ export const AnrokIntegrationMapItemDialog = forwardRef<AnrokIntegrationMapItemD
     const isCollectionContext = !Object.values(MappableTypeEnum).includes(
       localData?.type as MappableTypeEnum,
     )
-    const refetchQueries =
-      localData?.type === MappableTypeEnum.AddOn
-        ? ['getAddOnsForAnrokItemsList']
-        : localData?.type === MappableTypeEnum.BillableMetric
-          ? ['getBillableMetricsForAnrokItemsList']
-          : ['getAnrokIntegrationCollectionMappings']
+    const refetchQueries = useMemo(() => {
```
domenicofalco, line 119: is useMemo really needed in here?

src/hooks/plans/useVolumeChargeForm.ts
```diff
@@ -137,16 +137,20 @@ export const useVolumeChargeForm: UseVolumeChargeForm = ({
             // fromValue should always be toValueOfPreviousRange + 1
             const { toValue } = acc[i - 1]
             const fromValue = String(Number(toValue || 0) + 1)
+            const formattedToValue = () => {
```
domenicofalco: it seems that this utils is repeated 2/3 times along the PR. should we think of a separated utils to be shared? This way we can also add unit tests around it. 

Also, we may add some typescript around the function so that we could potentially limit some of the explicit type conversion

src/pages/PlansList.tsx
```diff
@@ -58,6 +58,28 @@ const PlansList = () => {
   const list = data?.plans?.collection || []
   const shouldShowItemActions = hasPermissions(['plansCreate', 'plansUpdate', 'plansDelete'])
 
+  const getEmptyState = (): Partial<GenericPlaceholderProps> => {
```
domenicofalco, line 61: it would be good to have some simple unit tests around this

src/components/taxes/TaxCodeSnippet.tsx
```diff
@@ -9,10 +9,24 @@ const { apiUrl } = envGlobalVar()
...
+
+    if (isEdition) {
+      if (initialTaxCode) {
+        sufix = `/${initialTaxCode}`
+      } else {
+        sufix = '/__CODE_OF_TAX__'
+      }
+    }
+    return `${prefix}${sufix}`
+  }
```
AllanMichay, line 22: How about something like this? (Careful, tabulations are not correct ones in my example)

```suggestion
 const urlString = () => {
  const prefix = `${apiUrl}/api/v1/taxes`

  if (!isEdition) {
    return prefix
  }

  if (initialTaxCode) {
    return `${prefix}/${initialTaxCode}`
  }

  return `${prefix}/__CODE_OF_TAX__`
}
```

src/core/apolloClient/reactiveVars/envGlobalVar.ts
```diff
@@ -13,14 +13,14 @@ interface EnvGlobal {
   disablePdfGeneration: boolean
 }
 
-const apiUrl = !!window.API_URL
-  ? window.API_URL
-  : !!window.LAGO_DOMAIN
-    ? `https://${window.LAGO_DOMAIN}/api`
-    : API_URL
+const getApiUrl = () => {
+  if (window.API_URL) return window.API_URL
```
AllanMichay: I think we should still cast it as a boolean

src/core/apolloClient/reactiveVars/envGlobalVar.ts
```diff
@@ -13,14 +13,14 @@ interface EnvGlobal {
...
+  if (window.LAGO_DOMAIN) return `https://${window.LAGO_DOMAIN}/api`
```
AllanMichay: Cast it as a boolean too

src/pages/settings/Dunnings/Dunnings.tsx
```diff
@@ -159,77 +159,85 @@ const Dunnings = () => {
...
-                            },
-                          },
-                          {
-                            startIcon: 'trash',
-                            title: translate('text_1732187313660we30lb9kg57'),
-                            onAction: () => {
-                              deleteCampaignDialogRef.current?.openDialog(campaign)
+                )}
+
+                {hasAccessToFeature && (
```
AllanMichay, line 164: I feel like I see this logic of "has access to feature > display infinite scroll" a few times in the app. 

Should we consider using a component to handle this? Especially since the logic is also complex in the scolling

AllanMichay (commented):
Left a few suggestions. Overall, thanks for doing this.

I feel like our codebase would be way ore readable if we did not allow the usage of functions in the middle of component declaration.

Like if instead of this

```jsx
<MyComponent onExit={() => {
/**
* Complex logic here that creates multiple levels of tabulations
*/
}} />
```

We had

```jsx
const handleExit = () => {

}

<MyComponent onExit={handleExit} />
```

src/core/apolloClient/reactiveVars/envGlobalVar.ts
ansmonjol: done

src/core/apolloClient/reactiveVars/envGlobalVar.ts
ansmonjol: Agree done

src/pages/PlansList.tsx
ansmonjol: Hum I'm not a big fan of testing those. It's mainly translations and logic is straightforward IMO.
not sure if it's personal preference or not tho

src/hooks/plans/useVolumeChargeForm.ts
ansmonjol: Good point, done!

src/components/settings/integrations/AnrokIntegrationMapItemDialog.tsx
ansmonjol: Probably not indeed, the component is not rendering that often

src/pages/settings/Dunnings/Dunnings.tsx
ansmonjol: Definitely something to consider yeah.
I would aim to keep it like this for now, cause I would prefer to have a component to integrate a non-infinite scroll pagination in the future and felt it could be grouped during this initiative

src/components/taxes/TaxCodeSnippet.tsx
ansmonjol: Love it!

src/hooks/plans/utils.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const formatToValue = (toValue: any, fromValue: string) => {
```
AllanMichay: Hmmm given the name of the function and the `any` typing, it's hard to understand what this function really does.

Maybe give it a more understandable name.

ansmonjol: Renamed to `formataAnyToValueForChargeFormArrays`, make more sense?

src/components/settings/integrations/AnrokIntegrationMapItemDialog.tsx
ansmonjol: Oh actually id does, as it relies on a data that could either be present or undefined

PR #2668, opened 2025-10-31
feat(integrations): allow integration mapping per billing entity

AllanMichay:
## Context

Integration mappings allow the use of third party services. But currently, we are limiting this mapping at one per connection.
Now, we will allow the same connection to a third party to have multiple mappings depending on Lago's billing entity

## Description

Updates the mapping drawers to allow defining multiple mappings depending on the billing entity.
the previous mappings will be defined as "Default mapping". If a billing entity doesn't have a mapping, it will fallback to the default one

Backend can be found [here](https://github.com/getlago/lago-api/pull/4458)

Fixes [ISSUE-1248](https://linear.app/getlago/issue/ISSUE-1248/allow-integration-mapping-per-billing-entity)

### DIFF
--- a/src/pages/settings/integrations/IntegrationMapItemDrawer/IntegrationMapItemDrawer.tsx
+++ b/src/pages/settings/integrations/IntegrationMapItemDrawer/IntegrationMapItemDrawer.tsx
@@ -117,7 +117,7 @@ export function IntegrationMapItemDrawer<FormValues extends FormikValues>({
       }
     >
       <div className="flex flex-col gap-12">
-        <div>
+        <div className="flex flex-col gap-1">
           <Typography variant="headline">{title}</Typography>
           <Typography>{description}</Typography>
         </div>


src/pages/settings/integrations/AvalaraIntegrationMapItemDrawer/handleIntegrationMappingCUD.ts
```diff
@@ -0,0 +1,178 @@
...
+  const isCreate = !!initialMapping && !initialMapping.itemId && hasItemValues
+  const isEdit = !isCreate && hasInitialData && hasItemValues
+  const isDelete =
+    !isCreate && !isEdit && !hasItemValues && !!initialMapping && initialMapping.itemId
+
+  /**
+   * Happens since we launch this function for each billing entity, but some billing entities
+   * might not have any data to process (no initial data and no input data) = we want to keep the default mapping
+   */
+  if (!isCreate && !isEdit && !isDelete) {
```
ansmonjol: Find this condition cryptic and not clear when reading.

I understand it makes sense to rely on those variables as they contain useful check however I don't get why it would be a success if something is not a creation, edition nor deletion.

Would checking the presence of `initialMapping`, `hasInitialData` and `hasItemValues` be more explicit on the reason of this return presence?

Let's have a call to talk about it, might not grasp some edge cases reading file-per-file this PR

src/pages/settings/integrations/IntegrationMapItemDrawer/IntegrationMapItemDrawer.tsx
```diff
@@ -0,0 +1,174 @@
...
+    onSubmit: async (values) => {
+      const promises = (billingEntities || []).map(async (billingEntity) => {
+        if (!itemMappings || !type || !integrationId) return
+
+        const billingEntityKey = billingEntity.key || DEFAULT_MAPPING_KEY
+        const inputValues = values[billingEntityKey]
+
+        const initialMapping = itemMappings[billingEntity.key || DEFAULT_MAPPING_KEY]
+
+        return await handleDataMutation(
```
ansmonjol: Note for myself: check where does this comes from

src/pages/settings/integrations/AvalaraIntegrationMapItemDrawer/handleIntegrationMappingCUD.ts
AllanMichay: Yeah, I need to find a better way to explain this.

Basically, what we do is we always launch this function for ALL the billing entity. This means that we also do this for billing entities that didn't have values in the past and don't have it now. But yeah, using initial mapping and everything might be a best way to do this

This is also a part that I'm trying to refactor since all those `handleIntegrationMappingCUD` basically do the same just with parameters that are a bit different

src/pages/settings/integrations/common/handleIntegrationMappingCreateUpdateDelete.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    if (!errors || errors.length === 0) {
+      return { success: true }
+    }
+
+    return { success: false, errors }
+  }
+
+  if (isEdit) {
+    if (!initialMapping?.itemId) return { success: false, reasons: ['No initial mapping ID found'] }
```
domenicofalco, line 144: is the reason string an internal info or something for users that needs to go into the translation file?

src/pages/settings/integrations/IntegrationItem/IntegrationItemsTable.tsx
```diff
@@ -18,25 +27,85 @@ const IntegrationItemsTable = ({
...
+        id: null,
+        key: DEFAULT_MAPPING_KEY,
+        name: translate('text_6630e3210c13c500cd398e97'),
+      },
+    ]
+
+    if (
+      !billingEntitiesData ||
+      !billingEntitiesData.billingEntities ||
+      !billingEntitiesData.billingEntities.collection
```
domenicofalco, line 49: would just this condition be enough?

`!billingEntitiesData?.billingEntities?.collection`

src/pages/settings/integrations/IntegrationItem/IntegrationItemsTable.tsx
```diff
@@ -18,25 +27,85 @@ const IntegrationItemsTable = ({
 }: IntegrationItemsTableProps) => {
   const { translate } = useInternationalization()
 
-  const findItemMapping = (item: IntegrationItemData): ItemMapping | undefined => {
-    if (!item.integrationMappings || item.integrationMappings.length === 0) {
-      return undefined
+  const { data: billingEntitiesData, loading: isLoadingBillingEntities } =
+    useGetBillingEntitiesQuery()
+
+  const billingEntitiesColumns = useMemo(() => {
```
domenicofalco: I'd challenge the need of memoization here. Is **billingEntitiesColumns**  a heavy operation?

src/pages/settings/integrations/IntegrationItem/IntegrationItemsTable.tsx
```diff
@@ -18,25 +27,85 @@ const IntegrationItemsTable = ({
...
+  const generateTableColumnDataFromBillingEntity = (
+    column: BillingEntityForIntegrationMapping,
+  ): TableColumn<IntegrationItemData> => ({
+    key: 'id',
+    title: column.name,
+    content: (item: IntegrationItemData) => {
+      const itemMapping = findItemMapping(item, column.id)
+      const mappingInfos = getMappingInfos(itemMapping, provider)
+
+      const getStatusType = (): StatusType => {
```
domenicofalco: any chance to make this a separate function? i think it is a strong candidate for a unit test

src/pages/settings/integrations/IntegrationItem/IntegrationItemsTable.tsx
```diff
@@ -18,25 +27,85 @@ const IntegrationItemsTable = ({
...
+        }
+
+        if (!!mappingInfos && !mappingInfos.name) {
+          return translate('text_17272714562192y06u5okvo4')
+        }
+
+        return `${mappingInfos.name}${!!mappingInfos.id ? ` (${mappingInfos.id})` : ''} `
+      }
+
+      return <Status type={getStatusType()} label={getStatusLabel()} />
```
domenicofalco: getStatusType() and getStatusLabel() looks like they both use the same condition (?) Would it make sense to have only one function eg **getStatusDetails** that returns the 2 props? something like 

```
function getStatusDetails(): { type: StatusType; label: StatusLabel } {
  ...  
  if conditions
  ...

  return {
    type: ...,
    label: ...,
  }
}

...

const { type, label } = getStatusDetails();

return <Status type={type} label={label} />
```

to be then passed as props to the <Status /> component

src/pages/settings/integrations/AvalaraIntegrationMapItemDrawer/handleIntegrationMappingCUD.ts
```diff
@@ -0,0 +1,178 @@
+import {
+  MappableTypeEnum,
+  MappingTypeEnum,
+  useCreateAnrokIntegrationCollectionMappingMutation,
+  useCreateAnrokIntegrationMappingMutation,
+  useDeleteAnrokIntegrationCollectionMappingMutation,
+  useDeleteAnrokIntegrationMappingMutation,
+  useUpdateAnrokIntegrationCollectionMappingMutation,
+  useUpdateAnrokIntegrationMappingMutation,
```
domenicofalco: This file belongs to **AvalaraIntegration** although I see Anrok types imported. is that intentional?

src/pages/settings/integrations/common/handleIntegrationMappingCreateUpdateDelete.ts
AllanMichay: No, I mostly used it for debugging and found it really useful at that point

src/pages/settings/integrations/IntegrationItem/IntegrationItemsTable.tsx
AllanMichay: It owuld yes, but since we said we'd rather avoid `?`, I thought it would be good to follow this principle

src/pages/settings/integrations/IntegrationItem/IntegrationItemsTable.tsx
AllanMichay: Yeah you're right, it's not that heavy and shouldn't change that much

src/pages/settings/integrations/AvalaraIntegrationMapItemDrawer/handleIntegrationMappingCUD.ts
AllanMichay: Nope, an error. but I'm currently merging all the logic of the handle integration since there are little to no change in between the elements

src/pages/settings/integrations/IntegrationItem/useGetStatusDetails.ts
```diff
@@ -0,0 +1,32 @@
+import { StatusType } from '~/components/designSystem'
+import { useInternationalization } from '~/hooks/core/useInternationalization'
+
+export const useGetStatusDetails = () => {
```
domenicofalco, line 4: I'd add a unit test for this hook as it handles some logic inside with the status type

AllanMichay: @domenicofalco I already did it, you can find it here: https://github.com/getlago/lago-front/blob/9b45dfc15ce1a1c185e2d4f998ad277363e1d1d5/src/pages/settings/integrations/IntegrationItem/__tests__/useGetStatusDetails.test.ts

yeah the PR is getting so big it's hard to see the changes, sorry

domenicofalco: ahh sorry thanks!

PR #2679, opened 2025-11-04
feature(multiple payment flow): adding new payment view on the customer details section and a dialog to generate checkout URL

domenicofalco:
## Context

This task covers a specific part of the overall flow.
The implementation will be placed behind a feature flag to enable incremental releases.

## Description

A new payment section has been added to the customer information view. Also, from there, it is possible to trigger a modal that is being used to generate checkout url to allow customers to add more payment methods

Fixes LAGO-975

---

> [!NOTE]
> Adds a feature-flagged payment methods section on customer details with a dialog to generate/copy a checkout URL, refactors integration rows to accept a linked provider, and updates GraphQL/types/tests/translations.
> 
> - **Customer details (UI)**:
>   - Adds `CustomerPaymentMethods` section (behind `FeatureFlags.MULTI_PAYMENT_FLOW`) with dialog to generate/copy a checkout URL via `generateCheckoutUrl`.
>   - Preselects provider when only one, disables action when only ineligible methods, and shows empty/eligibility states.
>   - Refactors `CustomerIntegrationRows` to accept `linkedPaymentProvider` prop and use `customerPaymentProvider`.
>   - `CustomerMainInfos` now fetches payment providers, derives `linkedPaymentProvider`, renders new section, and adds section shadow spacing.
> - **GraphQL**:
>   - Adds `generateCheckoutUrl` mutation + hook.
>   - Restores `paymentProvidersListForCustomerMainInfos` query usage in `CustomerMainInfos` and generated types.
> - **Core/Types/UI**:
>   - Adds `LinkedPaymentProvider` type and related test factory.
>   - Extends `PageSectionTitle` action with `isDisabled`.
>   - Introduces `MULTI_PAYMENT_FLOW` feature flag helpers.
> - **Tests**:
>   - New tests for `CustomerPaymentMethods` (eligibility, dialog, success/loading/error, clipboard/toast).
>   - Updates `CustomerIntegrationRows` tests to pass `linkedPaymentProvider`.
> - **Translations**:
>   - Adds strings for payment methods section, dialog, and statuses.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 35c63d6c3bac3ea1bc887e7eed6f49d0d7365322. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/components/customers/CustomerPaymentMethods.tsx
+++ b/src/components/customers/CustomerPaymentMethods.tsx
@@ -46,12 +46,12 @@ gql`
 
 export const CustomerPaymentMethods = ({ customer, linkedPaymentProvider }: Props) => {
   const { translate } = useInternationalization()
-  const [hasError, setHasError] = useState(false)
-  const [generateCheckoutUrlMutation, { data, loading, error }] = useGenerateCheckoutUrlMutation({
-    variables: {
-      input: { customerId: customer.id },
-    },
-  })
+  const [generateCheckoutUrlMutation, { data, loading, error, reset }] =
+    useGenerateCheckoutUrlMutation({
+      variables: {
+        input: { customerId: customer.id },
+      },
+    })
 
   const addPaymentDialogRef = useRef<DialogRef>(null)
 
@@ -89,14 +89,6 @@ export const CustomerPaymentMethods = ({ customer, linkedPaymentProvider }: Prop
     }
   }, [connectedPaymentProvidersData, hasOneAvailableOption])
 
-  useEffect(() => {
-    if (error) {
-      setHasError(true)
-    } else {
-      setHasError(false)
-    }
-  }, [error])
-
   const checkoutUrl = data?.generateCheckoutUrl?.checkoutUrl || ''
 
   return (
@@ -131,8 +123,8 @@ export const CustomerPaymentMethods = ({ customer, linkedPaymentProvider }: Prop
           title={translate('text_1761914802986ww4ima0w9w9')}
           description={translate('text_1761914802986ipq0aot8fas')}
           onClose={() => {
-            if (hasError) {
-              setHasError(false)
+            if (!!error) {
+              reset()
             }
           }}
           actions={({ closeDialog }) => (
@@ -174,7 +166,7 @@ export const CustomerPaymentMethods = ({ customer, linkedPaymentProvider }: Prop
               }}
             />
 
-            {hasError && (
+            {!!error && (
               <Alert type="danger" className="mb-8" data-test={ERROR_ALERT_TEST_ID}>
                 {translate('text_1762182354095wfjiizpju0e')}
               </Alert>
--- a/src/components/customers/__tests__/CustomerPaymentMethods.test.tsx
+++ b/src/components/customers/__tests__/CustomerPaymentMethods.test.tsx
@@ -36,14 +36,19 @@ const linkedPaymentProvider = createMockLinkedPaymentProvider({
 })
 
 const mockGenerateCheckoutUrlMutation = jest.fn()
+const mockReset = jest.fn(() => {
+  mockMutationState.error = null
+})
 const mockMutationState: {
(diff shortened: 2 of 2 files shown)


src/components/customers/CustomerPaymentMethods.tsx
```diff
@@ -0,0 +1,215 @@
...
+    hasAtLeastOneAvailablePaymentMethod && !isCustomerEligibleForAddingPaymentMethods
+
+  const mapOptions = linkedPaymentProvider ? [linkedPaymentProvider] : []
+
+  const connectedPaymentProvidersData = mapOptions.map(({ code, name }) => ({
+    value: code,
+    label: name,
+  }))
+
+  const hasOneAvailableOption = connectedPaymentProvidersData.length === 1
```
ansmonjol, line 84: May no need change, just trying to understand better the code.
If I understand correctly this condition will always be true, correct?
I mean, the length will either be 1 or 0, and (for now) cannot be more than that today?

src/components/customers/CustomerPaymentMethods.tsx
```diff
@@ -0,0 +1,215 @@
...
+    }
+  }, [connectedPaymentProvidersData, hasOneAvailableOption, setSelectedPaymentProvider])
+
+  useEffect(() => {
+    if (error) {
+      setHasError(true)
+    } else {
+      setHasError(false)
+    }
+  }, [error])
```
ansmonjol: Why using this useEffect?

When error is updated the component already re-renders and you could use `!!error` to assert it's presence.

Any reason to maintain this side state?

src/components/customers/CustomerPaymentMethods.tsx
domenicofalco: Yes Alex you are correct, although the condition can also be false when there no option for the combobox (which can be the case). The condition is there basically to prefill the combobox in case there's only one payment provider, so that the user does not need to select it manually. In case it is 0, no prefill

src/components/customers/CustomerPaymentMethods.tsx
domenicofalco: You are correct. Although I wanted to add an extra state for it due to user experience reason: the thing is when an error is thrown in the apollo state, it persists. This means if the user closes the dialog and open it again, the error from the apollo state is still there but since I wanted to clear it to have a better UX (imho), i've added a state to handle this scenario so that when the user opens the dialog again, they have a clean view without the error. On dialog close eventi, setHasError -> false

src/components/customers/CustomerPaymentMethods.tsx
ansmonjol: Guess it will help the day we can have more than 1, ok thanks!

domenicofalco: exactly, that is the main thought behind it as from the design I could spot that providers can be multiple in a future iteration.

src/components/customers/CustomerPaymentMethods.tsx
ansmonjol: You could check the `reset` method provided by the apollo mutation.

Something like 
```diff
+ const [generateCheckoutUrlMutation, { data, loading, error, reset }] = useGenerateCheckoutUrlMutation({
    variables: {
      input: { customerId: customer.id },
    },
  })


...



<Dialog
    onClose={() => {
+        reset()
    }}
>


```

ansmonjol: Feels it solves the issue you describe

domenicofalco: that looks way better Alex, thanks. i've made the changes based on the reset method. 

I've kept the **if (!!error) {** check though, so to not reset the generated checkout link as well if the uses closes the modal. This way we save some unnecessary calls against stripe and we only care about resetting the error onDialogClose if present

ansmonjol (approved):

PR #2697, opened 2025-11-10
chore(eslint): add custom rule to prevent formikProps array dependency issue

domenicofalco:
## Context

Add custom ESLint rule to prevent formikProps usage as dependency in React hooks

## Description

A new custom ESLint rule (`lago/no-formik-props-in-effect`) has been implemented to prevent the use of `formikProps` as a dependency in React hooks (only `useEffect` for now).

## 🎯 Problem

Using `formikProps` directly as a dependency in React hooks causes infinite re-rendering loops, since `formikProps` is a new object reference on every render. This can lead to:
- Performance degradation
- Infinite rendering loops
- Unexpected application behavior

## ✅ Temporary solution implemented to prevent its usage

A custom ESLint rule has been created that:
- **Reports an error** when `formikProps` is used directly as a dependency in React hooks
- **Allows** the use of specific `formikProps` properties (e.g., `formikProps.values`, `formikProps.setFieldValue`) as dependencies
- **Prevents** developers from accidentally adding `formikProps` to dependency arrays

Once the issues with formik are solve, we may remove this rule

Fixes LAGO-987

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,6 +8,12 @@
     ["tw\\(([^)]*)\\)", "(?:'|\"|`)([^']*)(?:'|\"|`)"]
   ],
   "eslint.useFlatConfig": true,
+  "eslint.validate": [
+    "javascript",
+    "javascriptreact",
+    "typescript",
+    "typescriptreact"
+  ],
   "javascript.preferences.importModuleSpecifier": "non-relative",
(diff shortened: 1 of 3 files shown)


AllanMichay (approved):
LGTM 💅

ansmonjol (approved):

PR #2704, opened 2025-11-12
misc: better query and state management on organization switch

ansmonjol:
## Context

The organization switch [seems](https://lago-eu.sentry.io/issues/77122518/?alert_rule_id=172021&alert_type=issue&notification_uuid=a72a1998-485a-47d5-ab69-b7212016aeea&project=4509111387619408) to trigger unhandled error
This is an attempt to handle it in a better way.

## Description

This PR extracts the orga switch logic, wrap with a try catch and manually send to sentry error cause the error object there does not bubbles correctly.

Been aggressive in the disabled state using a `useState` boolean to prevent any double click that could occur before the promise runs.

Also improved some ternary in template, removed an unused error and improved cache/query state used in this file.

### DIFF
--- a/src/layouts/MainNavLayout.tsx
+++ b/src/layouts/MainNavLayout.tsx
@@ -1,5 +1,6 @@
 import { gql, useApolloClient } from '@apollo/client'
 import { ClickAwayListener, Stack } from '@mui/material'
+import { captureException } from '@sentry/react'
 import { Avatar, ConditionalWrapper, Icon, IconName, Spinner, Tooltip } from 'lago-design-system'
 import { useEffect, useRef, useState } from 'react'
 import { Location, Outlet, useLocation, useNavigate } from 'react-router-dom'
@@ -102,6 +103,7 @@ const MainNavLayout = () => {
   const navigate = useNavigate()
   const client = useApolloClient()
   const [open, setOpen] = useState(false)
+  const [isSwitchingOrg, setIsSwitchingOrg] = useState(false)
   const { currentUser, loading: currentUserLoading, refetchCurrentUserInfos } = useCurrentUser()
   const { hasPermissions } = usePermissions()
   const {
@@ -110,7 +112,10 @@ const MainNavLayout = () => {
     refetchOrganizationInfos,
   } = useOrganizationInfos()
   const { translate } = useInternationalization()
-  const { data, loading, error } = useSideNavInfosQuery()
+  const { data, loading } = useSideNavInfosQuery({
+    errorPolicy: 'all',
+    notifyOnNetworkStatusChange: true,
+  })
   const {
     openPanel: openInspector,
     closePanel: closeInspector,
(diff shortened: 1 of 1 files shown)


src/layouts/MainNavLayout.tsx
```diff
@@ -130,6 +135,30 @@ const MainNavLayout = () => {
...
+
+      navigate(HOME_ROUTE)
+
+      const refetchPromises = [refetchOrganizationInfos(), refetchCurrentUserInfos()]
+
+      await Promise.allSettled(refetchPromises)
+    } catch (error) {
+      // Errors are automatically captured by Apollo's error link
+      // However the error details is not captured by Sentry, so we need to capture it manually
+      captureException(error)
```
domenicofalco, line 170: just an idea: as we have the errors automatically captured by Apollo error link here https://github.com/getlago/lago-front/blob/main/src/core/apolloClient/init.ts#L131 , should we find a way to bring there the error details as well rather then spamming the sentry captureException around with the risk we may forgot about it? this way we make sure we all centralized in the apolloClient init without the need of using captureException anymore and more importantly we'd have more valuable info in sentry to debug

src/layouts/MainNavLayout.tsx
```diff
@@ -130,6 +135,30 @@ const MainNavLayout = () => {
...
+    if (isSwitchingOrg) return
+
+    setIsSwitchingOrg(true)
+
+    try {
+      await switchCurrentOrganization(client, organizationId)
+
+      if (isInspectorOpen) closeInspector()
+
+      navigate(HOME_ROUTE)
```
AllanMichay, line 161: Should we maybe wait everything is settled before navigating ?

I feel like the safe ordering should be

- closeInspector
- switch navigation
- refetch
- navigate

Or maybe I'm missing something here. Like maybe do we flash informations doing so?

src/layouts/MainNavLayout.tsx
ansmonjol: Agree, I'm not sure how to make it bubble correctly. Need to spend some time here.

ansmonjol: I would say, let's be noisy at first as the initial issue relies on the fact that the Sentry catch is not containing enough error.

Hopefully this better error handling would prevent more to happen, otherwise we will (should) have a very clear error WDYT?

src/layouts/MainNavLayout.tsx
ansmonjol: Having the navigation later would also have a drawback (we tried in the past).

After navigating to home page, some queries are fired as they are necessary for the app to know where to redirect user.

However in such case we would love to navigate, first fetch orga and user new infos, then allow the home queries to fire which is not really feasible.

When you fetch the orga and user infos, the whole app shows a loading stage. So it's not "weird" to have it before redirection as the user won't see any content.

That way, we can fetch correct infos, then trigger a redirect that will handle all logic based on "correct" and final context.

Feels I should comment this section's order with a short summary of the above, gonna give it a try

ansmonjol: Just did

PR #2735, opened 2025-11-19
bug: Fix error toast generic error on payment page

domenicofalco:
## Context

Handle InvoicesNotOverdue error in payment request page

## Description

Add error handling for the `InvoicesNotOverdue` GraphQL error in the CustomerRequestOverduePayment component. When the query returns this error, the user is shown a generic error toast and redirected to the customer details page.

Changes:
- Added `silentErrorCodes` context to `useGetRequestOverduePaymentInfosQuery` to silently handle `InvoicesNotOverdue` errors
- Added `useEffect` hook to detect and handle `InvoicesNotOverdue` errors from the query, displaying an error toast and navigating to customer details (redirect)
- Added a new generic error message translation key for better user experience
- Added unit test to verify the error handling behavior

Fixes ISSUE-1106

### DIFF
--- a/src/pages/CustomerRequestOverduePayment/index.tsx
+++ b/src/pages/CustomerRequestOverduePayment/index.tsx
@@ -194,8 +194,8 @@ const CustomerRequestOverduePayment: FC = () => {
   )
 
   useEffect(() => {
-    // when user lands on this page
-    // redirect if invoices are not ready for payment processing
+    // Check and redirect if invoices are not ready for payment processing
+    // Runs when payment status changes (on mount and when dependencies update)
     if (!isPaymentProcessingStatusLoading && !isCustomerReadyForOverduePayment) {
       handlePaymentNotReady()
     }


src/pages/CustomerRequestOverduePayment/index.tsx
```diff
@@ -55,13 +57,11 @@ gql`
 
     invoices(paymentOverdue:true, customerId: $id) {
       collection {
-        ...InvoicesForDunningEmail
```
domenicofalco, line 58: This wasn't needed as its fragment was a duplication of "InvoicesForRequestOverduePaymentForm" which was already applied to the query

src/pages/CustomerRequestOverduePayment/index.tsx
```diff
@@ -175,6 +193,15 @@ const CustomerRequestOverduePayment: FC = () => {
...
   )
 
+  useEffect(() => {
+    // when user lands on this page
+    // redirect if invoices are not ready for payment processing
+    if (!isPaymentProcessingStatusLoading && !isCustomerReadyForOverduePayment) {
+      handlePaymentNotReady()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isCustomerReadyForOverduePayment, isPaymentProcessingStatusLoading])
```
ansmonjol: > when user lands on this page

Well, not exactly haha, this occur after first render and on each renders later.
Is the comment wording wrong or the way we handle that logic wrong?

domenicofalco: you are totally right, updating the comment. ty

PR #2736, opened 2025-11-19
feat: Enable custom Okta host URL

domenicofalco:
## Context

Add support to input Okta host URL.

## Description

Introduce an optional input field where users can enter their Okta host. The value must contain only the host without any protocol since the backend enforces https.

Screenshot 2025-11-19 at 16 11 10

Fixes ISSUE-1266


src/components/settings/authentication/AddOktaDialog.tsx
```diff
@@ -24,6 +24,8 @@ export interface AddOktaDialogRef {
   closeDialog: () => unknown
 }
 
+export const SUBMIT_BUTTON_TEST_ID = 'add-okta-dialog-submit-button'
```
ansmonjol: Nitpicking
Can we have a mention of Okta in the name? 
I'm afraid of later having tons of SUBMIT_BUTTON_TEST_ID and having to choose in the autocomplete without seeing the full path

domenicofalco: sure, makes totally sense

domenicofalco:
waiting for QA review and then merge it

PR #2743, opened 2025-11-20
Add test coverage and Sonarqube configuration

domenicofalco:
## Context

Add test coverage and Sonarqube configuration

Fixes ISSUE-1277


sonar-project.properties
```diff
@@ -0,0 +1,20 @@
+# Source code location
+sonar.sources=src
+
+# Exclusions
+sonar.exclusions=**/types.{ts,tsx},**/fixtures.ts,**/core/constants/**,**/generated/**,**/styles/**,**/pages/__devOnly/**
```
ansmonjol, line 9: Thought about excluding `/scripts` too WDYT?

domenicofalco: it should not be necessary as per `sonar.sources=src`

PR #2744, opened 2025-11-20
feat: allow to edit and add a fixed charge

ansmonjol:
## Context

This is the last step for the feature adding fixed charged on plans.
We're currently able to add them on a plan, got them on a  subscription and see them on an invoice.
This PR takes care of the edit fee logic.

## Description

This PR makes sure all logic around `edifFeeDrawer` is now taking into account the fixed charge.
The API of this drawer have been updated to handle all cases.

Also added some missing `refetchQueries` that were missing today, leading to some non-accurate UI states.

Fixes LAGO-967 LAGO-1030


src/components/invoices/details/EditFeeDrawer.tsx
```diff
@@ -258,45 +281,50 @@ export const EditFeeDrawer = forwardRef<EditFeeDrawerRef>((_, ref) => {
...
+        )
+
+        localData.onAdd({
           ...(localData.fee || {}),
           ...input,
+          invoiceSubscriptionId: invoiceSubscriptionId || '',
           charge: currentCharge as Charge,
-          chargeFilterId,
-          invoiceSubscriptionId: localData?.invoiceSubscriptionId,
+          fixedCharge: currentFixedCharge as FixedCharge,
```
domenicofalco, line 339: any chance to avoid the type assertion here?

src/components/invoices/details/EditFeeDrawer.tsx
```diff
@@ -468,12 +534,30 @@ export const EditFeeDrawer = forwardRef<EditFeeDrawerRef>((_, ref) => {
...
+
+                          if (isUsageCharge) {
+                            formikProps.setFieldValue('chargeId', selectedChargeId || '')
+                          } else if (isFixedCharge) {
+                            formikProps.setFieldValue('fixedChargeId', selectedChargeId || '')
+                          }
+                        }}
+                        value={
+                          formikProps.values.chargeId || formikProps.values.fixedChargeId || ''
+                        }
```
domenicofalco, line 604: what if the onChange is outside the render?

src/components/invoices/details/InvoiceDetailsTable.tsx
```diff
@@ -263,25 +264,30 @@ const AddFee = ({
...
-)
+}) => {
+  return (
+    <tr className="py-2 shadow-b">
+      <td>
+        <Button
+          variant="quaternary"
+          startIcon="plus"
+          onClick={() => {
+            if (onAdd && invoiceSubscriptionId) {
```
domenicofalco: i'd bring this outside the render and make an early return as well. wdyt?

src/components/invoices/details/EditFeeDrawer.tsx
ansmonjol: Not easilly, this onAdd logic is a mess and should be a broader refactor happening here

src/components/invoices/details/InvoiceDetailsTable.tsx
ansmonjol: I need to refactor this place, it should be unified with the current implementation and have an if/else with the different arguments passed.
Should be way better after

PR #2755, opened 2025-11-21
perf: add lazyComponent utils + suspense High Order Component

domenicofalco:
## Context

Add lazyComponent utils + suspanse high order component


pnpm-lock.yaml
```diff
@@ -16,7 +16,7 @@ importers:
         specifier: 3.14.0
         version: 3.14.0(@types/react@18.3.26)(graphql-ws@5.16.0(graphql@16.12.0))(graphql@16.12.0)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       '@fragaria/address-formatter':
-        specifier: 6.6.1
+        specifier: ^6.6.1
```
ansmonjol: Related?

domenicofalco: nope, that was after a fresh pnpm install

PR #2771, opened 2025-11-25
chore(Tooltip) style should live in the component directly

ansmonjol:
## Context

We should prevent relying on this huge style file, but rather have the style the nearest possible from the component code.

This is also a good way to easily switch to other component lib later.
Another advantage, it would simplify the style provider usage, as the component itself would own it's style, rather that expecting it to come from it's wrapper provider.

Don't mind the double tooltip implementation, it should be unified later

## Description

This PR moves the style to the component.

---

> [!NOTE]
> Moves Tooltip styling into component-level styled wrappers and removes global MuiTooltip theme overrides, updating usage to slotProps.
> 
> - **Tooltip components**:
>   - Introduce `StyledTooltip` using `styled(MuiTooltip)` in `packages/design-system/src/components/Tooltip.tsx` and `src/components/designSystem/Tooltip.tsx`.
>   - Apply inline typography, background, padding, and placement margins directly to the tooltip slot.
>   - Switch from `componentsProps` to `slotProps` and use `sx` to set `maxWidth`.
>   - Preserve open/close behavior; add focus handlers in the design-system variant.
> - **Theme**:
>   - Remove `MuiTooltip` `styleOverrides` from `src/styles/muiTheme.ts` (tooltip styling now owned by components).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 44bedc7e3df4a0bb8eaf88ad03393f1f66450255. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/design-system/src/components/Tooltip.tsx
+++ b/packages/design-system/src/components/Tooltip.tsx
@@ -1,7 +1,37 @@
 import { Tooltip as MuiTooltip, TooltipProps as MuiTooltipProps } from '@mui/material'
+import { styled } from '@mui/material/styles'
+import { colors } from 'lago-configs/tailwind'
 import { forwardRef, ReactNode, useCallback, useState } from 'react'
 
 import { tw } from '~/lib'
+
+// Styled Tooltip with design system styles applied at component level
+const StyledTooltip = styled(MuiTooltip)({
+  // Target the tooltip slot directly
+  '& .MuiTooltip-tooltip': {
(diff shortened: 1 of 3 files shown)


ansmonjol:
We agreed on performing such changes, will re-open later when relevant, working on other stuffs for now

PR #2788, opened 2025-11-28
chore(Agents.md): Split AGENTS.md into modular documentation files

domenicofalco:
## Context

The AGENTS.md file has grown significantly and contains multiple topics. To improve maintainability and organization, we've split it into separate markdown files organized by topic in the agents/ folder.


scripts/build-agents-doc.cjs
```diff
@@ -0,0 +1,67 @@
...
+-->
+
+# Lago Billing System - Development Assistant
+
+You are an expert development assistant for the Lago billing system project. Always follow these critical guidelines:
+`;
+
+  const sections = [header];
+
+  for (const filename of files) {
```
AllanMichay: Should we re-order file to be sure we use a more natural sort (like 1,2,3,4 not 1,11,12,2...) or are we sure the natural sort is existing from the file.readdirSync?

domenicofalco: not sure if we should care about the sort at all. i am open to feedback

PR #2818, opened 2025-12-03
fix(sentry): Send more context to Sentry with source maps

domenicofalco:
## Context

Improve Sentry error tracking and ensure secure handling of build-time variables. Source maps are uploaded during CI builds using environment variables that are never exposed to the frontend runtime.

### VIte
- Added new configuration to send source maps to Sentry at build time

### Apollo Client Error Link (src/core/apolloClient/init.ts):
- Added operation tag (query/mutation/subscription) for better error grouping
- Added operation query

### ErrorBoundary (src/components/ErrorBoundary.tsx):
- Added extra context (paths, error type)

### MainNavLayout (src/layouts/MainNavLayout.tsx):
- Added instanceof ApolloError check to prevent duplicate error reporting

## Security note

Sentry auth tokens and sensitive variables are only used at build time in CI/CD and never exposed to the frontend runtime. Source maps are uploaded during build and deleted from the distribution bundle.

Fixes ISSUE-845

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -118,9 +118,7 @@ export default defineConfig(({ mode }) => {
     },
     build: {
       outDir: 'dist',
-      // Use 'hidden' source maps in production for security (not publicly accessible)
-      // but still available for Sentry
-      sourcemap: isProduction ? 'hidden' : true,
+      sourcemap: true,
       target: 'esnext',
       rollupOptions: {
         output: {


vite.config.ts
```diff
@@ -85,7 +118,9 @@ export default defineConfig(({ mode }) => {
     },
     build: {
       outDir: 'dist',
-      sourcemap: true,
+      // Use 'hidden' source maps in production for security (not publicly accessible)
+      // but still available for Sentry
```
ansmonjol: What are the security concerns about them? Our code is open source and available to anyone.

Is it because it could contain some env var, and make them easier to find when inspecting the source code in browser?

I find it handy to be able to put breakpoints in the code on prod in the browser, just trying to understand if we can keep that or not

domenicofalco: Alex I've restored the previous version

PR #2829, opened 2025-12-04
misc(Wallet): increate creation limit to be 5

ansmonjol:
## Context

We were working on some BE improvement before allowing customers to have multiple wallets

## Description

It's now done and we set the rule to be 5 max


ansmonjol:
Waiting for a BE adjustment before merging

PR #2851, opened 2025-12-08
bug: prevent showing skip rule button if no rule is defined

ansmonjol:
## Context

When setting top-up min or max limits, you can choose to bypass the rules.

In the UI, it shows a ON/OFF switch to enable or not this rule.

However it's visible in some cases that does not makes sense

## Description

This PR ensure the bypass switch is only visible if a min or max rule have been defined.

It's currently flaky as it does not check for all possible values, such as empty string (when you enter a min/max value then reset the input)

Fixes ISSUE-1312

### DIFF
--- a/src/pages/wallet/CreateWalletTopUp.tsx
+++ b/src/pages/wallet/CreateWalletTopUp.tsx
@@ -16,6 +16,13 @@ import {
 } from '~/components/designSystem'
 import { AmountInputField, SwitchField, TextInputField } from '~/components/form'
 import { CenteredPage } from '~/components/layouts/CenteredPage'
+import {
+  ADD_METADATA_DATA_TEST,
+  CLOSE_CREATE_TOPUP_BUTTON_DATA_TEST,
+  IGNORE_PAID_TOPUP_LIMITS_SWITCH_DATA_TEST,
+  INVOICE_REQUIRES_SUCCESSFUL_PAYMENT_SWITCH_DATA_TEST,
+  SUBMIT_WALLET_DATA_TEST,
+} from '~/components/wallets/utils/dataTestConstants'
 import { addToast } from '~/core/apolloClient'
 import { CustomerDetailsTabsOptions } from '~/core/constants/tabsOptions'
 import { intlFormatNumber } from '~/core/formats/intlFormatNumber'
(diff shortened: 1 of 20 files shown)


src/pages/wallet/CreateWalletTopUp.tsx
```diff
@@ -247,8 +247,8 @@ const CreateWalletTopUp = () => {
   }, [formikProps.dirty, navigateBack])
 
   const hasMinMax =
-    typeof wallet?.paidTopUpMinAmountCents !== 'undefined' ||
-    typeof wallet?.paidTopUpMaxAmountCents !== 'undefined'
+    (wallet?.paidTopUpMinAmountCents !== null && wallet?.paidTopUpMinAmountCents !== undefined) ||
```
AllanMichay, line 257: Is it maybe time to create a core function `isEmpty` where we test those kind of scenarios? I feel like it's not the first time we stumble across those and this way, you can use this excuse to test empty strings and everything

src/pages/wallet/components/TopUpSection.tsx
```diff
@@ -365,6 +369,7 @@ export const TopUpSection: FC<TopUpSectionProps> = ({
                         }
                         label={translate('text_66a8aed1c3e07b277ec3990d')}
                         subLabel={translate('text_66a8aed1c3e07b277ec3990f')}
+                        data-test="recurring-invoice-requires-successful-payment-switch"
```
AllanMichay: use const

AllanMichay (commented):
Thanks for the E2E test

Use const for testing purposes instead of writing a string. this way, you can import the const into test files

src/pages/wallet/components/TopUpSection.tsx
ansmonjol: Done

src/pages/wallet/CreateWalletTopUp.tsx
ansmonjol: I don't 100% agree.
Hiding it somewhere would make it less comprehensive to me.
That method would ideally work with numbers too, and feel weird to have a check on empty string for it. Then we could have a `isNumberEmpty` but it start to be weird IMO

@AllanMichay if you think this should be discussed with the team please add this to the next FE weekly agenda!

PR #2854, opened 2025-12-09
refactor(ErrorBoundaries): Improve error monitoring and sourcemaps availability

domenicofalco:
## Context

Debugging production errors was difficult due to:
- DevTools panel errors lacked router context in Sentry (only main app pathname was reported)
- Sourcemaps were not fully available in production (design-system missing them)

## Description

- Add dedicated DevtoolsErrorBoundary component that captures MemoryRouter pathname and tags errors with context: 'Devtools'
- Add errorCategory: 'global' tag to main ErrorBoundary for better Sentry filtering
- Enable sourcemap generation for design-system package
- Remove filesToDeleteAfterUpload from Sentry config to keep sourcemaps available for self-hosted users

Fixes ISSUE-1337

### DIFF
--- a/packages/design-system/vite.config.ts
+++ b/packages/design-system/vite.config.ts
@@ -40,6 +40,7 @@ export default defineConfig({
       fileName: 'index',
       formats: ['es'],
     },
+    sourcemap: true,
     emptyOutDir: false,
     rollupOptions: {
       output: {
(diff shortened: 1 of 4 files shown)


AllanMichay (approved):
LGTM, just check the sonarcloud warnings

domenicofalco:
@AllanMichay jfyi regarding the soneq warnings about code duplication: in this case I think it is acceptable. The duplication is intentional because we want to have two separate error boundaries. Even though they share a very similar structure, this approach keeps the code clearer and more maintainable. Thanks

PR #2881, opened 2025-12-16
feat(multiple payment): Add multiple payment flow to wallet top-up and recurring rules

domenicofalco:
## Context

The PR aims to add the new multiple payment flow (including invoice custom sections) to wallet top-up and recurring rules

Fixes [LAGO-978](https://linear.app/getlago/issue/LAGO-978)


src/pages/wallet/CreateWallet.tsx
```diff
@@ -233,7 +252,32 @@ const CreateWallet = () => {
         style: 'decimal',
         minimumFractionDigits: getCurrencyPrecision(currency),
       }),
-      recurringTransactionRules: wallet?.recurringTransactionRules || undefined,
+      recurringTransactionRules:
```
ansmonjol, line 256: Is the the right moment to extract this to a deserialize method that would be independent and tested?

domenicofalco: fair point Alex. done 🙌

PR #2882, opened 2025-12-18
fix(subscriptions): Resolve Apollo cache 404 errors after customer deletion

domenicofalco:
## Context

When a customer with subscriptions was deleted, subsequent operations on other subscriptions triggered 404 errors. This occurred because Apollo's `refetchQueries` with named queries attempted to refetch data for all cached queries, including those referencing deleted customers.

## Description

- Remove problematic `refetchQueries` from subscription mutations and replace with explicit field returns (`status`, `terminatedAt`, etc.) so Apollo auto-updates cache via normalization
- Use `cache.modify` in DeleteCustomerDialog to clear subscriptions cache without triggering aggressive refetches
- Hide Edit/Upgrade/Terminate actions for already terminated or canceled subscriptions in SubscriptionDetails page
- Add `isSubscriptionEditable()` utility function with unit tests
- Redirect to subscriptions list instead of customer page after terminating a subscription (prevents 404 if customer was deleted)

Fixes ISSUE-1357

### DIFF
--- a/src/hooks/customer/useAddSubscription.tsx
+++ b/src/hooks/customer/useAddSubscription.tsx
@@ -255,14 +255,11 @@ export const useAddSubscription: UseAddSubscription = ({
           }
         : undefined
 
-      let subscriptionAtForUpdate: string | undefined
-
-      if (existingSubscription?.startedAt) {
-        subscriptionAtForUpdate =
-          DateTime.fromISO(existingSubscription.startedAt).toUTC().toISO() || undefined
-      } else if (subsDate) {
-        subscriptionAtForUpdate = DateTime.fromISO(subsDate).toUTC().toISO() || undefined
-      }
+      const subscriptionAtForUpdate = DateTime.fromISO(
+        existingSubscription?.startedAt || subsDate || '',
+      )
+        .toUTC()
+        .toISO()
 
       const { errors } =
         formType === FORM_TYPE_ENUM.creation || formType === FORM_TYPE_ENUM.upgradeDowngrade


src/hooks/customer/useAddSubscription.tsx
```diff
@@ -236,6 +255,15 @@ export const useAddSubscription: UseAddSubscription = ({
           }
         : undefined
 
+      let subscriptionAtForUpdate: string | undefined
```
stephenlago99: (small suggestion, feel free to ignore)

We could shorten this to:

```tsx
const subscriptionAtForUpdate = DateTime.fromISO(existingSubscription.startedAt || subsDate || '').toUTC().toISO() || undefined
```

domenicofalco: done, thank you

src/components/customers/subscriptions/TerminateCustomerSubscriptionDialog.tsx
```diff
@@ -83,7 +86,6 @@ export const TerminateCustomerSubscriptionDialog =
     const refundAmount = deserializeAmount(invoice?.refundableAmountCents, currency)
 
     const [terminate] = useTerminateCustomerSubscriptionMutation({
-      refetchQueries: ['getCustomerSubscriptionForList', 'getSubscriptionsList'],
```
ansmonjol, line 86: Without those, is the subscription list udpated after the modal closes?

domenicofalco: yes, I ran several tests and it looks good. 

A key insight here we may want to leverage more: by including fields like **status** and **terminatedAt** in the mutation response (https://github.com/getlago/lago-front/pull/2882/files#diff-d5af1346e33b036f796314bad2de74039e9152484bc9c95dd404086deb3aa676R27), we leverage Apollo automatic cache normalization. that is because such fields changes, apollo understands such mutation and updates the cache. this removes the need for refetchQueries, saving HTTP requests while keeping the cache naturally in sync. i'd say let's dig into this pattern more and more to see how ti goes

PR #2895, opened 2025-12-23
fix(style): Fix request payment footer layout

domenicofalco:
fix(style): Extract AI agent width constant and fix footer layout

## Context

The new AI tab caused all content to shift left. The footer in the request payment view page needed to account for the AI agent width on medium screens to prevent layout issues.

Screenshot 2025-12-23 at 10 36 17


domenicofalco:
not sure about the requirement @AllanMichay but it is a fair point. perhaps @stephenlago99 knows more about?

PR #2900, opened 2025-12-24
feat(multiple payment): Add connection name to method column on customer payments tab

domenicofalco:
## Context

When displaying payments in the customer payments tab, users need to see which specific payment provider connection was used for each payment, especially when multiple connections exist for the same provider type. Previously, only the provider type (e.g., "Stripe", "Adyen") was displayed, making it difficult to distinguish between different connections.

## Description

- Enhanced `PaymentProviderChip` component to accept an optional `label` prop that allows displaying custom text (e.g., connection name) instead of the default provider label
- Updated GraphQL queries in `PaymentsList` to fetch `paymentProvider` details including `name` and `code` for all provider types (Adyen, Cashfree, Flutterwave, Gocardless, Moneyhash, Stripe)
- Modified `CustomerPaymentsList` to pass the payment provider connection name to the `PaymentProviderChip` component
- Refactored nested ternary expression in `PaymentProviderChip` for better code readability

The payment method column now displays the connection name when available, falling back to the provider type label if no connection name is provided.

Fixes LAGO-981

### DIFF
--- a/src/components/__tests__/PaymentProviderChip.test.tsx
+++ b/src/components/__tests__/PaymentProviderChip.test.tsx
@@ -1,0 +1,56 @@
+/* eslint-disable tailwindcss/no-custom-classname */
+import { cleanup, screen } from '@testing-library/react'
+
+import { ProviderTypeEnum } from '~/generated/graphql'
+import { render } from '~/test-utils'
+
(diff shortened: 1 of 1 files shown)


domenicofalco:
‼️ waiting for the https://github.com/getlago/lago-api/pull/4776 to be approved and merged

PR #2939, opened 2026-01-13
misc: rework fee grouping logic for invoice details

ansmonjol:
ℹ️ This PR is to be merged into another one, related to the addition of fixed charges: https://github.com/getlago/lago-front/pull/2744

## Context

A subscription can generate an invoice. And an invoice can have multiple subscription in it.
This joint table is called `InvoiceSubscription`

--

The current invoice details uses the `InvoiceSubscription` object and loop over it's fees, to make then appear in the UI.

While this works separating the fees in the correct boundary or "bucket", it's sort of hard to determine if those fees eventually need to be placed somewhere else depending if they are paid in advance or in arrears.

Meaning that a fee can have different boundaries even tho it's included in the same `InvoiceSubscription`.

We had some logic implemented but with the addition of the new fixedCharged types, the logic was starting to be out of control and very much error probe

## Description

This PR have the objective to change the way we loop of objects and aim for a simpler / more maintainable logic to handle fee display.
At the same time, it allows us to have only one loop logic, no matter the fee to be displayed.

Eventually, fees are grouped this way now
- subscription ID
  - fee boundary

The fee boundaries being ordered from the oldest to the newest.

Now, all fees are surely in their correct "bucket" and no matter what other types of fees can appear later, this logic will work. The `InvoiceSubscription` object is only fetched to make the reconciliation between invoice and sub ids, and get it's `accept_new_charge_fees` value

This is a good step in simplifying the invoice details logic.
I've also reworked a lot the fragment management around those files and that should now respect more our conventions.

This is a hairy one, please use my time if you have any questions or need guidances.
I've updated and added tests to we can safely rely on them later.

My next step is to make sure the regenerate flow is perfectly working. Right now the happy path is ok but I need to ensure some edge cases are too

Fixes LAGO-1030


src/pages/CustomerInvoiceDetails.tsx
```diff
@@ -379,18 +374,9 @@ const CustomerInvoiceDetails = () => {
...
-    error: invoiceSubscriptionError,
-  } = useGetInvoiceSubscriptionsQuery({
-    variables: { id: invoiceId as string },
-    skip: !invoiceId,
-    notifyOnNetworkStatusChange: true,
-  })
   const invoice = data?.invoice
   const invoiceFees = feesData?.invoice?.fees
-  const invoiceSubscriptions = invoiceSubscriptionData?.invoice?.invoiceSubscriptions
+  // const invoiceSubscriptions = invoiceSubscriptionData?.invoice?.invoiceSubscriptions
```
domenicofalco: if unneeded the comment may be remove

ansmonjol: ah yep thanks!

PR #2943, opened 2026-01-14
feat(multiple payment): Add payment method selection dialog in resend for collection flow & request payment

domenicofalco:
## Context

- The "Resend for collection" action was directly triggering the retry mutation without allowing users to select a specific payment method
- Users need the ability to select which payment method should be used when requesting payment for overdue invoices 

All gated behind the MULTI_PAYMENT_FLOW feature flag.

## Description

- Created `ResendInvoiceForCollectionDialog` component with `PaymentMethodComboBox` for payment method selection
- Updated `InvoicesList` and `CustomerInvoicesList` to open the dialog instead of directly calling the mutation
- Added `externalId` to customer fragment for payment method fetching
- Payment method field is mandatory (resend for collection)
- Added payment method selection section to the Request Payment form (optional)

Fixes LAGO-979
Fixes LAGO-980

### DIFF
--- a/src/components/customers/__tests__/CustomerInvoicesList.test.tsx
+++ b/src/components/customers/__tests__/CustomerInvoicesList.test.tsx
@@ -32,6 +32,8 @@ jest.mock('~/hooks/usePermissionsInvoiceActions', () => ({
     canVoid: () => false,
     canIssueCreditNote: () => false,
     canRecordPayment: () => false,
+    canResendEmail: () => false,
+    canRegenerate: () => false,
   }),
 }))
 


domenicofalco:
the PR is ready to be reviewed and then merged, although the BE isn't. let's wait for BE to be merged first

src/components/invoices/ResendInvoiceForCollectionDialog.tsx
```diff
@@ -0,0 +1,143 @@
...
+              paymentMethod: {
+                paymentMethodId: selectedPaymentMethod.paymentMethodId,
+                paymentMethodType: selectedPaymentMethod.paymentMethodType,
+              },
+            }),
+          },
+        },
+      })
+
+      if (hasDefinedGQLError('PaymentProcessorIsCurrentlyHandlingPayment', errors)) {
```
stephenlago99, line 95: Could this fail because of other types of errors?

domenicofalco: this is handled differently on purpose: if you see at line [53](https://github.com/getlago/lago-front/pull/2943/files#diff-b800b759e4efb785dcc399dbf977f74c2a21655f6e7a6b15db1e033c47493bbfR52), it has been silenced so to have a specific _info message_ and it was inherited from the previous implementation. I think should be ok, wdyt?

domenicofalco: All other errors should be caught from the global error boundary

AllanMichay: Looks good to me yes

PR #2963, opened 2026-01-20
fix(mutiple payments): Improve value masking utility and add it to payment methods combobox

domenicofalco:
## Context

  The existing `obfuscateValue` function had a confusing API with positional
  parameters that made it unclear what each value represented. Above all,
  the payment method display needed to show masked card last 4 digits.

  ## Description

  - Rename `obfuscate.ts` to `maskValue.ts` with cleaner API
  - Replace positional params with named options: `dotsCount`, `visibleChars`, `withSpace`
  - Update `formatPaymentMethodDetails` to include last4 digits in display
  - Update `ApiKeys` and `Webhooks` components to use new `maskValue` function
  - Update tests

Fixes [LAGO-1095](https://linear.app/getlago/issue/LAGO-1095)

### DIFF
--- a/src/core/formats/__tests__/maskValue.test.ts
+++ b/src/core/formats/__tests__/maskValue.test.ts
@@ -1,6 +1,12 @@
 import { maskValue } from '~/core/formats/maskValue'
 
 describe('maskValue', () => {
+  describe('GIVEN empty value', () => {
+    it('THEN should return dash', () => {
+      expect(maskValue('')).toBe('-')
+    })
+  })
+
   describe('GIVEN no options', () => {
     it('THEN should prepend 4 dots without space', () => {
       expect(maskValue('4242')).toBe('••••4242')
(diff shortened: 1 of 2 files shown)


src/components/developers/webhooks/Webhooks.tsx
```diff
@@ -180,7 +180,9 @@ export const Webhooks = () => {
                                   : undefined
                               }
                             >
-                              {showOrganizationHmac ? hmacKey : obfuscateValue(hmacKey || '')}
+                              {showOrganizationHmac
+                                ? hmacKey
+                                : maskValue(hmacKey || '', { dotsCount: 8, visibleChars: 3 })}
```
AllanMichay, line 185: What happens with the maskValue when the string is empty? Do we just display the dots?

domenicofalco: fair point Allan!

domenicofalco: I made a change to cover that case

PR #2964, opened 2026-01-20
feat(dialog): create new dialog management

AllanMichay:
## Context

From the get go, we were using a simple Dialog management that consisted of ref and using the component. Dialog management has become incredibly harder and two issues popped out

- too many different dialog component
- props drilling

## Description

This PR aims to simplify dialog management using [eBay/nice-modal-react](https://github.com/eBay/nice-modal-react)

With this we will have
- no more props drilling
- create component only for the content

Fixes [LAGO-1089](https://linear.app/getlago/issue/LAGO-1089/email-resend-update-our-dialog-to-have-correct-design)

---

> [!NOTE]
> **Medium Risk**
> Adds new global modal provider/registration and introduces new async resolution/error-handling flows for dialogs, which could affect close behavior and promise consumers if misused; changes are additive and covered by new unit/snapshot tests.
> 
> **Overview**
> Introduces a new, centralized dialog system built on `@ebay/nice-modal-react`, including a shared `BaseDialog`, `CentralizedDialog` (confirm/cancel flow), `DialogOpeningDialog` (can launch a second dialog and return its promise), and a `PremiumWarningDialog`, with consistent result typing via `DialogResult` and shared action handling in `useDialogActions`.
> 
> Wires the app to support this by adding `NiceModalProvider` at the root and registering dialogs in `~/core/dialogs/registeredDialogs`, adds a dev-only design-system tab to manually exercise the new dialogs, and updates ESLint to warn on imports of legacy dialog components to drive migration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f4082fdc134644fd32db08e5f1a107159c613cb3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/configs/eslint.config.mjs
+++ b/packages/configs/eslint.config.mjs
@@ -6,16 +6,16 @@ import pluginPrettier from 'eslint-plugin-prettier/recommended'
 import pluginReact from 'eslint-plugin-react'
 import pluginReactHooks from 'eslint-plugin-react-hooks'
 import pluginTailwind from 'eslint-plugin-tailwindcss'
-import { fileURLToPath } from 'node:url'
-import { dirname, resolve } from 'node:path'
 import globals from 'globals'
+import { dirname, resolve } from 'node:path'
+import { fileURLToPath } from 'node:url'
+import pluginTypescriptEslint from 'typescript-eslint'
+
+import noFormikPropsInEffect from './eslint-rules/no-formik-props-in-effect.js'
 
 const __filename = fileURLToPath(import.meta.url)
 const __dirname = dirname(__filename)
 const tailwindConfigPath = resolve(__dirname, 'tailwind.config.ts')
-import pluginTypescriptEslint from 'typescript-eslint'
-
-import noFormikPropsInEffect from './eslint-rules/no-formik-props-in-effect.js'
 
 /** @type {import('eslint').Linter.Config[]} */
 export default [
@@ -162,17 +162,49 @@ export default [
       '@typescript-eslint/no-explicit-any': 'off',
     },
   },
-  // Formik deprecation warning - use @tanstack/react-form instead
   {
     files: ['**/*.{js,mjs,cjs,ts,mts,jsx,tsx}'],
     rules: {
       'no-restricted-imports': [
         'warn',
         {
+          // Formik deprecation warning - use @tanstack/react-form instead
           paths: [
             {
               name: 'formik',
               message: 'Formik is deprecated. Use @tanstack/react-form instead.',
+            },
+          ],
+          // Old dialog components deprecation warning - use ~/components/dialogs instead
+          patterns: [
+            {
+              group: [
+                '~/components/designSystem/Dialog',
+                '~/components/designSystem/WarningDialog',
+                '~/components/designSystem/PreventClosingDrawerDialog',
+                '~/components/PremiumWarningDialog',
+                '~/components/addOns/*Dialog*',
+                '~/components/billableMetrics/*Dialog*',
+                '~/components/coupons/*Dialog*',
+                '~/components/customers/*Dialog*',
+                '~/components/customers/**/*Dialog*',
+                '~/components/developers/**/*Dialog*',
+                '~/components/exports/*Dialog*',
+                '~/components/features/*Dialog*',
+                '~/components/invoices/*Dialog*',
+                '~/components/invoices/**/*Dialog*',
+                '~/components/invoceCustomFooter/*Dialog*',
+                '~/components/paymentMethodSelection/*Dialog*',
+                '~/components/paymentMethodsList/*Dialog*',
+                '~/components/plans/*Dialog*',
+                '~/components/settings/**/*Dialog*',
+                '~/components/subscriptions/*Dialog*',
+                '~/components/subscriptions/**/*Dialog*',
+                '~/components/taxes/*Dialog*',
+                '~/components/wallets/*Dialog*',
+              ],
+              message:
+                'This dialog component is deprecated. Please use the new dialog management system in ~/components/dialogs instead.',
             },
           ],
         },
--- a/src/components/PremiumWarningDialog.tsx
+++ b/src/components/PremiumWarningDialog.tsx
@@ -10,17 +10,11 @@ type TProps = {
   mailtoBody?: string
 }
 
-/**
- * @deprecated Please use the new PremiumWarningDialog
- */
 export interface PremiumWarningDialogRef extends DialogRef {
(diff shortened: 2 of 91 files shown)


src/components/dialogs/PremiumWarningDialog.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const modal = useModal()
+    const { translate } = useInternationalization()
+
+    const displayTitle = title || translate('text_63b3155768489ee342482f4f')
+    const displayDescription = description || translate('text_63b3155768489ee342482f51')
+    const displayMailtoSubject = mailtoSubject || translate('text_63b3f676d44671bf24d81411')
+    const displayMailtoBody = mailtoBody || translate('text_63b3f676d44671bf24d81413')
+
+    const handleClose = () => {
+      modal.resolve()
```
ansmonjol: Reject instead?

package.json
```diff
@@ -114,6 +114,7 @@
   },
   "dependencies": {
     "@apollo/client": "3.14.0",
+    "@ebay/nice-modal-react": "^1.2.13",
```
ansmonjol: Can we fix the version?

src/components/dialogs/PremiumWarningDialog.tsx
AllanMichay: I could use reject yes but since there are no "resolve" action, I thought it wasn't that important

AllanMichay: Decided to use reject on all of the cancel button, for design consistency

ansmonjol: Yeah cause I would expect my callback to not fire in case the user presses cancel button

src/components/dialogs/WarningDialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const modal = useModal()
+    const { translate } = useInternationalization()
+
+    const handleCancel = async (): Promise<void> => {
+      modal.reject()
+      modal.hide()
+    }
+
+    const handleContinue = async (): Promise<void> => {
+      await onContinue()
```
domenicofalco: what if `await onContinue()` throws an error? should it be wrapped around a `try ... catch`?

src/components/dialogs/WarningDialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+export default WarningDialog
+
+export const useWarningDialog = (): HookDialogReturnType<WarningDialogProps> => {
+  const modal = useModal(WARNING_DIALOG_NAME)
+
+  return {
+    open: (props?: WarningDialogProps) => modal.show(props),
+    close: () => modal.hide(),
+    resolve: (args?: unknown) => modal.resolve(args),
+    reject: (args?: unknown) => modal.reject(args),
```
domenicofalco: why exporting **resolve** & **reject**? are them handled within the `WarningDialog` already? or am I missing the point here

src/pages/__devOnly/ModalTest.tsx
```diff
@@ -0,0 +1,200 @@
...
+      actions: <ModalTestActions modal={modal} />,
+    })
+  }
+  const handleOpenLongDialogWithDescription = (): void => {
+    modal.open({
+      title: 'Test Modal',
+      description: 'This is a description for the long modal.',
+      headerContent: <LongModalHeaderContent />,
+      children: <LongModalContent />,
+      actions: <ModalTestActions modal={modal} />,
```
domenicofalco: nothing critical but just thinking loud: I am wondering if there's a different way instead of passing the `modal` instance? It looks a bit weird injecting the entire `modal` in a `modal.open` prop.
would a similar pattern as below worth exploring? this also simplify the type over the `<Modal` because now I must aways have a `HookDialogReturnType<...>` type filled

```
actions: (
      <ModalTestActions
        onCancel={() => { modal.reject(); modal.close() }}
        onConfirm={() => { modal.resolve(); modal.close() }}
      />
    ),
```

src/components/dialogs/WarningDialog.tsx
domenicofalco: also, where will the loading state be handles? should we have a default structure for it within the <BaseDialog and pass it as props? or do you think it should be handled from the consumer?

domenicofalco:
thanks for the huge work on this @AllanMichay, it looks good to me. Just left few points to clarify 🙌

src/components/dialogs/WarningDialog.tsx
AllanMichay: you're right, I didn't really think about it completely

src/pages/__devOnly/ModalTest.tsx
AllanMichay: Oh yes clearly. To be fair, this page will not stay, it will be deleted and wasreally used for design QA

src/components/dialogs/WarningDialog.tsx
AllanMichay: loading state is already handled by the button automatically since it's an async function

AllanMichay: Good catch. this does open up a question: 

Should all the modal.reject be resolve with some other parameters? Because I feel like in this case, I'd use a reject but it would end up being treated at the same place where the normal close is treated if I do so

domenicofalco: we could add an optional onError callback to the dialog props, like:

  - by default, if onContinue fails, the dialog stays open so the user can retry
  - the consumer can pass onError to handle the error as needed by the consumer
  - the error handling is encapsulated inside the dialog, so the consumer doesn't need to wrap every call in try/catch which becomes tedious on every async modal
  
  something like:
  
```
  const handleContinue = async (): Promise<void> => {
    try {
      await onContinue()
      modal.resolve()
      modal.hide()
    } catch (error) {
       // dialog stays open, user can retry 
      onError?.(error) // this can be anything from the consumer
    }
  }
```

thoughts?

src/components/dialogs/useDialogActions.ts
```diff
@@ -0,0 +1,67 @@
...
+    } catch (error) {
+      if (closeOnError) {
+        modal.reject({
+          reason: 'error',
+          error: error as Error,
+        })
+        modal.hide()
+      } else {
+        onError?.(error as Error)
+      }
```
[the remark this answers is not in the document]
AllanMichay: Can't do it otherwise cannot keep the dialog open

domenicofalco:
LGTM

PR #2972, opened 2026-01-22
bug(customer-portal): redirect from usage if subscription is not active

ansmonjol:
## Context

We fetch usage of subscription in the customer portal.

If the link is shared and the subscription is terminated in the meantime, the usage call will fail

## Description

This takes care of redirecting the user - with a new error toast - in case one of the two usage queries fails with the `NoActiveSubscription` error. 

Fixes ISSUE-1400


AllanMichay (commented):
Please add tests, coverage is at 0% on this new code

PR #2976, opened 2026-01-23
misc: split main nav layout in multiple components

ansmonjol:
## Context

The main nav sections start to grow and I thought it was a good idea to start splitting it while investing a bug. 

This bug no longer exists but basically what was happening is that we have different sections in the menu and it could happen that all the items within a section are hidden. By the way this was a mistake and this will not happen anymore. While refactoring that it was great also to include a new way to build a section and make sure that you check if all the items are visible or not before displaying a section. 

## Description

So this pool request makes the two changes:
1. The split into multiple chunks
2. Implement this new logic to build the different items within a section

Fixes ISSUE-1409

---

> [!NOTE]
> **Main changes**
> 
> - Split `MainNavLayout` into focused components: `MainNavLayout`, `OrganizationSwitcher`, `MainNavMenuSections`, `BottomNavSection`, and `VerticalMenuSkeleton`
> - Introduced `utils.getNavTabs` to detect when all tabs are hidden and conditionally render/hide entire sections
> - Updated `NavLayout` and `Spinner` to accept `data-test` attributes for improved testability
> - Switched router layout import to `~/layouts/MainNavLayout/MainNavLayout`
> 
> **Quality and tests**
> 
> - Added unit tests for all new components and `utils.getNavTabs`, covering visibility logic and interactions
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c0bbfec1fb3572a0d260bcbaa07b0a579d183e04. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/layouts/MainNavLayout/BottomNavSection.tsx
+++ b/src/layouts/MainNavLayout/BottomNavSection.tsx
@@ -15,8 +15,6 @@ import { NavLayout } from '~/layouts/NavLayout'
 
 import { getNavTabs, NavTab } from './utils'
 
-const { appEnv } = envGlobalVar()
-
 export const BOTTOM_NAV_SECTION_TEST_ID = 'bottom-nav-section'
 
 interface BottomNavSectionProps {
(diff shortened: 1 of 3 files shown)


src/layouts/MainNavLayout/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './MainNavLayout'
```
domenicofalco: not sure if we were against barrel export during our folder structure discussion? I could not find this point in linear but I might be wrong

AllanMichay: I'm pretty sure we said we should avoid it, you're right. I'm gonnna change it directly since I'm in charge of support this week

PR #2987, opened 2026-01-27
refactor(form migration): migrate SignUp form to Tanstack Form

domenicofalco:
## Context

  The SignUp form was using Formik with Yup validation, which is being
  phased out in favor of Tanstack Form with Zod validation for better
  type safety and consistency across the codebase.

  ## Description

  - Migrate SignUp.tsx from Formik/Yup to Tanstack Form/Zod
  - Create Zod validation schema with separate schemas for standard signup and Google registration flows
  - Add usePasswordValidation hook for real-time password strength feedback with single source of truth (validatePassword function)
  - Update TextInputFieldForTanstack to handle multiple errors with proper translation before joining, and add showOnlyErrors prop to filter which validation errors are displayed
  - Export EMAIL_REGEX from zodCustoms for reuse across validation schemas
  - Tests

Fixes ISSUE-1417

### DIFF
--- a/src/pages/auth/SignUp.tsx
+++ b/src/pages/auth/SignUp.tsx
@@ -17,7 +17,6 @@ import {
 import { LagoApiError, useGoogleRegisterMutation, useSignupMutation } from '~/generated/graphql'
 import { useInternationalization } from '~/hooks/core/useInternationalization'
 import { useAppForm } from '~/hooks/forms/useAppform'
-import { useShortcuts } from '~/hooks/ui/useShortcuts'
 import { theme } from '~/styles'
 import { Card, Page, StyledLogo, Subtitle, Title } from '~/styles/auth'
 import { tw } from '~/styles/utils'
@@ -109,7 +108,6 @@ const SignUp = () => {
     },
   })
 
-  const canSubmit = useStore(form.store, (state) => state.canSubmit)
   const password = useStore(form.store, (state) => state.values.password)
   const passwordValidation = usePasswordValidation(password)
 
@@ -144,16 +142,6 @@ const SignUp = () => {
       setIsGoogleRegister(true)
     }
   }, [googleCode])
-
-  useShortcuts([
-    {
-      keys: ['Enter'],
-      disabled: !canSubmit,
-      action: () => {
-        form.handleSubmit()
-      },
-    },
-  ])
 
   const handleSubmit = (e: React.FormEvent) => {
     e.preventDefault()


src/pages/auth/signUpForm/validationSchema.ts
```diff
@@ -0,0 +1,101 @@
...
+  .string()
+  .min(1, { message: PASSWORD_VALIDATION_ERRORS.REQUIRED })
+  .superRefine((val, ctx) => {
+    validatePassword(val).forEach((error) => {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: error,
+      })
+    })
+  })
```
AllanMichay: I feel like those two could live in the zodCustoms since they could be re-used (especially requiredEmail)

src/components/form/TextInput/TextInputFieldForTanstack.tsx
```diff
@@ -1,29 +1,39 @@
...
+    : allErrors
+
+  // Translate each error key and join them before passing to getErrorToDisplay
+  const translatedError = filteredErrors.map((errorKey) => translate(errorKey as string)).join('\n')
 
   const finalError = getErrorToDisplay({
-    error,
-    errorMap,
+    error: translatedError,
+    errorMap: {},
```
AllanMichay: Why do you give an empty object here?
errorMap is used when we have an issue coming from the back. Try to validate it by creating a customer using an already existing external id

src/pages/auth/SignUp.tsx
```diff
@@ -77,30 +66,39 @@ const SignUp = () => {
...
-    email: '',
-    password: '',
-    organizationName: '',
+  const form = useAppForm({
+    defaultValues: signUpDefaultValues,
+    validationLogic: revalidateLogic(),
+    validators: {
+      onDynamic: isGoogleRegister ? googleRegisterValidationSchema : signUpValidationSchema,
+    },
+    onSubmit: async ({ value }) => {
```
AllanMichay, line 87: Don't forget you can use onInvalidSubmit to scroll on the first incorrect input 😄

domenicofalco: great point, done!

src/components/form/TextInput/TextInputFieldForTanstack.tsx
domenicofalco: good point, my bad. fixing this

src/pages/auth/signUpForm/validationSchema.ts
domenicofalco: done

src/formValidation/zodCustoms.ts
```diff
@@ -47,3 +47,67 @@ export const zodOneOfPermissions = z.string().refine((value) => {
 
   return allPermissions.includes(value as PermissionName)
 })
+
+export const zodRequiredEmail = z
+  .string()
+  .min(1, { message: 'text_620bc4d4269a55014d493f3d' })
+  .refine((val) => EMAIL_REGEX.test(val), 'text_620bc4d4269a55014d493f43')
```
AllanMichay: Just saw that we have a translation key for this: 
`text_620bc4d4269a55014d493fc3`

domenicofalco: fixed

PR #2995, opened 2026-01-29
misc: migrate premium warning dialog

AllanMichay:
## Context

Migrate the Dialog for premium

## Description

Move to our new Dialog system

Fixes LAGO-XXX


AllanMichay:
I think we should move all of them one by one

PR #2997, opened 2026-01-29
fix(webhooks): remove retry possibility on webhook success

AllanMichay:
## Context

When a webhook has failed, we can retry it on the UI. Issue is, after a retry, it is still possible to retry since ui isn't updated

## Description

Use polling after a retry to
- stop the possibility of retries for 3s
- update the status automatically after 3s (or less)

Fixes [ISSUE-1366](https://linear.app/getlago/issue/ISSUE-1366/prevent-retry-of-a-succeeded-webhook)

### DIFF
--- a/src/components/developers/webhooks/WebhookLogDetails.tsx
+++ b/src/components/developers/webhooks/WebhookLogDetails.tsx
@@ -102,9 +102,13 @@ export const WebhookLogDetails = ({ goBack }: { goBack: () => void }) => {
   // Launch the webhook retry and then wait for the status to be other than 'PENDING'
   // We check every second if the status has changed or not. Until then, the retry button is disabled
   const retryWebhookAndWait = async () => {
-    await retry()
+    const { data: retryData } = await retry()
 
-    // Abort controller only needed if we launched a retry
+    // Only poll if the retry was accepted by the backend
+    if (!retryData?.retryWebhook) {
+      return
+    }
+
     abortControllerRef.current = new AbortController()
 
     await pollUntilCondition(


src/components/developers/webhooks/WebhookLogDetails.tsx
```diff
@@ -81,7 +109,7 @@ export const WebhookLogDetails = ({ goBack }: { goBack: () => void }) => {
           <>
             {webhookType}
             {hasError && (
-              <Button variant="quaternary" onClick={async () => await retry()}>
+              <Button variant="quaternary" onClick={async () => await retryWebhookAndWait()}>
```
domenicofalco: should the button be disabled while retrying? else users can click multiple times before polling completes (polling may last few seconds)

src/components/developers/webhooks/WebhookLogDetails.tsx
```diff
@@ -68,6 +77,25 @@ export const WebhookLogDetails = ({ goBack }: { goBack: () => void }) => {
 
   const hasError = status === WebhookStatusEnum.Failed
 
+  // Launch the webhook retry and then wait for the status to be other than 'PENDING'
+  // We check every second if the status has changed or not. Until then, the retry button is disabled
+  const retryWebhookAndWait = async () => {
+    await retry()
```
domenicofalco: if retry fails for some reason, the code still proceeds to create an AbortController and start polling. should this be wrapped in a try ... catch instead?

src/components/developers/webhooks/WebhookLogDetails.tsx
AllanMichay: Button is disabled. This behaviour is done in the button. It is disabled until the promise has been resolved

src/components/developers/webhooks/WebhookLogDetails.tsx
AllanMichay: you're right, will do 👍

PR #3324, opened 2026-04-09
feat: Store superset filters in localStorage

stephenlago99:
### Context

Superset embedded dashboards lose user-selected filters on page refresh. When a user selects a filter (e.g., Currency: EUR), refreshing the page resets all filters to their defaults, forcing users to reapply them every time.

### Description

Persist Superset dashboard filter selections in `localStorage` so they survive page refresh.

**How it works:**
- When a user changes filters, the `observeDataMask` callback (debounced 500ms) saves the active native filter state to `localStorage`
- On next page load, saved filters are RISON-encoded and passed to the Superset embedded SDK via the `native_filters` URL parameter
- Only `NATIVE_FILTER-*` entries with non-empty values are stored — chart cross-filters and empty filter states are excluded
- Storage is scoped per organization and dashboard (`superset-filters-{orgId}`) to prevent filter leakage between users or dashboards
- This is hidden behind a frontend feature flag (run `localStorage.setItem('featureFlags', JSON.stringify(['superset_persistent_filters']))` to enable it)

**Key decisions:**
- Inline RISON encoder (~25 lines) instead of adding an unmaintained `rison` npm package
- Filters with cleared values are removed from storage, so the dashboard's defaults apply on reload
- No cleanup of orphan localStorage keys — old entries expire naturally if a dashboard's `embeddedId` changes

Fixes LAGO-1787

### DIFF
--- a/src/pages/dashboards/Dashboards.tsx
+++ b/src/pages/dashboards/Dashboards.tsx
@@ -10,6 +10,7 @@ import {
   ORGANIZATION_LS_KEY_ID,
   SUPERSET_FILTERS_LS_KEY_PREFIX,
 } from '~/core/constants/localStorageKeys'
+import { FeatureFlags, isFeatureFlagActive } from '~/core/utils/featureFlags'
 import { encodeRison } from '~/core/utils/risonEncoder'
 import { extractNativeFilters } from '~/core/utils/supersetFilters'
 import { useSupersetDashboardsQuery } from '~/generated/graphql'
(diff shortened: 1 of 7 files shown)


src/pages/dashboards/Dashboards.tsx
```diff
@@ -55,26 +61,59 @@ const Dashboards = () => {
...
             expanded: true,
           },
+          ...(urlParams && { urlParams }),
         },
         iframeSandboxExtras: ['allow-top-navigation', 'allow-popups-to-escape-sandbox'],
       })
 
+      embedded.observeDataMask((dataMask) => {
+        if (debounceTimer) clearTimeout(debounceTimer)
+        debounceTimer = setTimeout(() => {
```
AllanMichay: Why not use debounce from lodash?

AllanMichay: Move this function into it's own extracted file so you can test it

AllanMichay (approved):

PR #3340, opened 2026-04-13
chore: combobox chevron should also receive disabled props

ansmonjol:
Minor fix, noticed that while navigating using TAB, the chevron of disabled combobox was selectable.

It's now fixed

### DIFF
--- a/src/components/form/ComboBox/ComboBoxInput.tsx
+++ b/src/components/form/ComboBox/ComboBoxInput.tsx
@@ -88,6 +88,7 @@ export const ComboBoxInput = ({
                 variant="quaternary"
                 size="small"
                 icon="chevron-up-down"
+                disabled={restParams.disabled}
                 onClick={restParams.disabled ? undefined : () => inputProps.onMouseDown()}
               />
             )}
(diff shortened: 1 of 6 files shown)


src/components/form/ComboBox/ComboBoxInput.tsx
```diff
@@ -88,6 +88,7 @@ export const ComboBoxInput = ({
                 variant="quaternary"
                 size="small"
                 icon="chevron-up-down"
+                disabled={restParams.disabled}
                 onClick={restParams.disabled ? undefined : () => inputProps.onMouseDown()}
```
domenicofalco, line 92: Not sure the `onClick={restParams.disabled ? ...}` check is needed here, since the `disabled` prop should already prevent the click.

domenicofalco (approved):

src/components/form/ComboBox/ComboBoxInput.tsx
ansmonjol: Without this props we can select the chevron button-icon using tab even tho the container is disabled

PR #3342, opened 2026-04-14
feat(quotes): display quotes

AllanMichay:
## Context

We are currently working on Quotes and how to display them

## Description

Display quotes and quote details in their own pages

Fixes LAGO-1392
Fixes LAGO-1394


src/pages/quotes/common/getQuoteStatusMapping.ts
```diff
@@ -0,0 +1,18 @@
+import { StatusProps, StatusType } from '~/components/designSystem/Status'
+import { StatusEnum } from '~/generated/graphql'
+
+export const getQuoteStatusMapping = (
+  status: StatusEnum,
+  translate: (key: string) => string,
+): Pick<StatusProps, 'type' | 'label'> => {
+  switch (status) {
+    case StatusEnum.Draft:
+      return { type: StatusType.outline, label: 'draft' as const }
```
domenicofalco, line 10: do these labels in the file need translations?

src/pages/quotes/hooks/useQuotes.ts
```diff
@@ -0,0 +1,90 @@
...
+    | ((
+        fetchMoreOptions: FetchMoreQueryOptions<OperationVariables, GetQuotesQuery>,
+      ) => Promise<unknown>)
+    | undefined
+}
+
+export const useQuotes = (
+  variables?: Omit<GetQuotesQueryVariables, 'limit' | 'page'>,
+): UseQuotesReturn => {
+  const [getQuotes, { data, loading, error, fetchMore }] = useGetQuotesLazyQuery({
```
domenicofalco: do you need `getQuotes` at all? if fire immediately on the useEffect below, it would be the same as using `data` directly with `useGetQuotesQuery`

src/pages/quotes/QuoteDetails.tsx
```diff
@@ -0,0 +1,83 @@
...
+import QuoteDetailsVersions from './QuoteDetailsVersions'
+
+const QuoteDetails = (): JSX.Element => {
+  const { translate } = useInternationalization()
+  const navigate = useNavigate()
+  const { quoteId } = useParams()
+  const { quote, loading } = useQuote(quoteId)
+
+  useEffect(() => {
+    if (!loading && !quote) {
```
domenicofalco: consider using an early return

src/pages/quotes/QuoteDetails.tsx
```diff
@@ -0,0 +1,83 @@
...
+      navigate(QUOTES_LIST_ROUTE, { replace: true })
+    }
+  }, [loading, quote, navigate])
+
+  const activeTabContent = useMainHeaderTabContent()
+
+  const getViewSubtitle = () => {
```
domenicofalco: the ternary exp is pretty flat, it might be used directly in the render instead of a separate function

domenicofalco (approved):
LGTM, just some minor comments. Also, noticed that `RolePermissionsForm.test.tsx.snap` is very huge and have +28k line of snapshots. is it really worth to have such test to maintain?

src/pages/quotes/common/getQuoteStatusMapping.ts
AllanMichay: The status component have it's own translation logic for those specific labels

src/pages/quotes/hooks/useQuotes.ts
AllanMichay: Not at all, it seems I have a bit of a regression somewhere. I had corrected it by hand while Claude was working on something else

AllanMichay:
> LGTM, just some minor comments. Also, noticed that `RolePermissionsForm.test.tsx.snap` is very huge and have +28k line of snapshots. is it really worth to have such test to maintain?

Yeah, I also think I will delete it

PR #3346, opened 2026-04-15
feat(org-slug): Phase 1/5 — Reactive var foundation (currentOrganizationVar)

domenicofalco:
## Context

First step of the organization slug feature (LAGO-1344). Replaces scattered localStorage reads for the current org ID with a single Apollo makeVar, following the same pattern as authTokenVar. Zero breaking changes — prepares the reactive foundation for Phase 2 where the URL slug becomes the source of truth.

## Description

- Create currentOrganizationVar (makeVar + getter/setter) in reactiveVars/
- Update init.ts, useCurrentUser.ts, cacheUtils.ts to read/write org ID through the reactive var instead of localStorage directly
- Add slug to CurrentUserInfos GraphQL fragment (needed by Phase 2 for client-side slug → org ID resolution from memberships)

Fixes LAGO-1344


src/core/apolloClient/reactiveVars/currentOrganizationVar.ts
```diff
@@ -0,0 +1,21 @@
+import { makeVar } from '@apollo/client'
+
+import { getItemFromLS, removeItemFromLS, setItemFromLS } from '~/core/apolloClient/cacheUtils'
+import { ORGANIZATION_LS_KEY_ID } from '~/core/constants/localStorageKeys'
+
+/** ----------------- VAR ----------------- */
+export const currentOrganizationVar = makeVar<string | null>(
```
AllanMichay, line 7: I have found usage of `getItemFromLS(ORGANIZATION_LS_KEY_ID)` in
- src/hooks/core/useLocationHistory.ts
- src/pages/Home.tsx (and the test)
- src/pages/dashboards/Dashboards.tsx

Do we want to replace all this places in this PR?

domenicofalco: Those 3 files are intentionally out of scope for Phase 1. They don't just read the org ID, they use it in navigation and routing logic and will be migrated throughout phase 2 and 3

domenicofalco: see here in case you want to get deeper https://www.notion.so/getlago/FE-Organization-slug-333ef63110d2806cb31bcb9fec1ed4c6 🙏

AllanMichay (approved):

PR #3365, opened 2026-04-17
feat(org-slug): Phase 2/5 — Router restructure

domenicofalco:
## Context

Second step of the organization slug feature ([LAGO-1345](https://linear.app/getlago/issue/LAGO-1345)). Nests all authenticated routes under `/:organizationSlug`, introduces `OrganizationLayout` to resolve the URL slug into an org ID via a client-side lookup on `currentUser.memberships`, and makes Home.tsx the slug-aware redirect hub.

## Description

- New `OrganizationLayout` resolves slug → org ID from memberships, gates `<Outlet />` until `currentOrganizationVar` matches, and clears Apollo cache on org switch detection
- Router restructure: all authenticated routes nest under `:organizationSlug` via a `makeRelative()` helper; Home also registered at `/` as redirect hub outside org scope
- `Home.tsx` slug-aware redirect hub: SSO saved path, router saved path (validated against user's org slugs), default permission-based — all `/${slug}`-prefixed
- Two 404 variants: `Error404` full-screen (outside org scope) and new `Error404InApp` inside SideNavLayout with "Go to homepage" CTA. This gives the chance to customaze it as we want to
- Sentry tracking via transitional `LEGACY_APP_PATH_SEGMENTS`: `legacy_url_accessed` (warning, external hits) and `slug_migration_missed_link` (error, missed internal migrations)
- `OrganizationSwitcher` navigates to `/${targetSlug}/` instead of `HOME_ROUTE`

Fixes [LAGO-1345](https://linear.app/getlago/issue/LAGO-1345)


src/layouts/OrganizationLayout.tsx
```diff
@@ -0,0 +1,103 @@
...
+      })
+    } else {
+      // External hit: old bookmark, stale link, or typed URL.
+      Sentry.captureMessage('legacy_url_accessed', {
+        level: 'warning',
+        ...sharedContext,
+        tags: { ...sharedContext.tags, source: 'external' },
+      })
+    }
+  }
```
AllanMichay, line 100: Shouldn't this be in a useEffect?

src/layouts/MainNavLayout/OrganizationSwitcher.tsx
```diff
@@ -84,9 +84,11 @@ export const OrganizationSwitcher = ({
     setIsSwitchingOrg(true)
 
     try {
+      const targetOrg = organizationList?.find((org) => org.id === organizationId)
+
       await switchCurrentOrganization(client, organizationId)
 
-      navigate(HOME_ROUTE)
+      navigate(`/${targetOrg?.slug}${HOME_ROUTE}`)
```
AllanMichay: Should we check if it exist before? Otherwise, we would go to `/undefined/`

src/core/router/utils/__tests__/orgSlug.test.ts
```diff
@@ -0,0 +1,148 @@
...
+      mockGetCurrentOrganizationId.mockReturnValue(undefined)
+
+      const result = resolveOrgSlug(buildCurrentUser([]))
+
+      expect(result).toBeUndefined()
+    })
+  })
+})
+
+describe('pathHasValidSlug', () => {
```
AllanMichay, line 81: Add a test on `/valid-slug` for pathHasValidSlug please. I feel like this case will not work because of the 
```
path.startsWith(`/${m.organization.slug}/`)
```
implem

src/pages/settings/SettingsHomePage.tsx
```diff
@@ -28,13 +29,14 @@ const SettingsHomePage = () => {
...
-        billingEntityCode: defaultBillingEntity.code,
-      }),
-      { replace: true },
-    )
-  }, [billingEntitiesData, navigate])
+    const target = generatePath(BILLING_ENTITY_ROUTE, {
+      billingEntityCode: defaultBillingEntity.code,
+    })
+
+    return navigate(organizationSlug ? `/${organizationSlug}${target}` : target, {
```
AllanMichay, line 36: Should we use `ensureSlugPrefix` here?

src/layouts/OrganizationLayout.tsx
domenicofalco: fair point. fixing it

src/layouts/MainNavLayout/OrganizationSwitcher.tsx
domenicofalco: that should not be the case because `targetOrg?.slug` comes from user membership table from graphql query. if this is missing, it means it's not being return from the API and, in turn, it should not get to the frontend at all. but it is a valid point, I am add a safety net

src/core/router/utils/__tests__/orgSlug.test.ts
domenicofalco: Good catch. added two explicit tests. The implementation looks still correct

src/pages/settings/SettingsHomePage.tsx
domenicofalco: that should be good. this block gets removed entirely in Phase 3 which I am about to start

AllanMichay (approved):

PR #3394, opened 2026-04-22
feat(org-slug): Phase 5/5 — Edit slug in organization settings

domenicofalco:
## Context

Final phase of the organization slug initiative: expose slug editing in a new "General" section under organization settings. BE already ships `updateOrganization(slug)`; the FE wires the UI, validation, and the in-app state propagation needed to keep the SPA in sync after a rename.

## Description

- New `/settings/general` route + "General" sidebar entry (last under "Your organization"), gated by `organizationView` / `organizationUpdate`.
- Slug block follows the `SettingsListWrapper` pattern, with inline edit button opening a NiceModal `FormDialog`.
- `useEditOrganizationSlugDialog` uses the new TanStack Form architecture (`useAppForm` + Zod) with schema + defaults extracted into a sibling `validationSchema.ts`, mirroring existing `*/validationSchema.ts`.
- `beforeChangeFormatter: ['lowercase', 'dashSeparator']` auto-normalizes input; single `superRefine` collapses length + format into one user-facing error. Backend `ValueAlreadyExist` on `slug` is mapped to an inline "not available" message via `formApi.setErrorMap`; expected 422s silenced from Sentry via `silentErrorCodes`.
- SPA stays consistent after rename via two targeted patches:
  - `cache.modify` on `Organization:${id}` inside the mutation's `update`, because `updateOrganization` returns `CurrentOrganization` (separate cache entity) and wouldn't update the membership-side `Organization` that `OrganizationLayout` reads to match URL slug.
  - New `rewriteSlugInLocationHistory` util rewrites the leading slug segment across `locationHistoryVar`, so `goBack(...)` ("Back to app" and siblings) doesn't point to a stale URL post-rename.
- `slug` added to `MainOrganizationInfos` fragment for the page render.

Fixes LAGO-1347

### DIFF
--- a/src/core/router/SettingRoutes.tsx
+++ b/src/core/router/SettingRoutes.tsx
@@ -40,6 +40,9 @@ const CreateInvoiceCustomSection = lazyLoad(
 )
 
 const TaxesSettings = lazyLoad(() => import('~/pages/settings/TaxesSettings'))
+const OrganizationGeneralSettings = lazyLoad(
+  () => import('~/pages/settings/OrganizationGeneralSettings/OrganizationGeneralSettings'),
+)
 const Integrations = lazyLoad(() => import('~/pages/settings/Integrations'))
 
 const AnrokIntegrationDetails = lazyLoad(() => import('~/pages/settings/AnrokIntegrationDetails'))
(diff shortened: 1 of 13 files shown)


src/pages/settings/OrganizationGeneralSettings/OrganizationGeneralSettings.tsx
```diff
@@ -0,0 +1,78 @@
...
+          metadata: translate('text_1776867582729flunw00muqy'),
+        }}
+      />
+
+      <SettingsPaddedContainer>
+        <SettingsListWrapper>
+          {loading && <SettingsListItemLoadingSkeleton count={1} />}
+
+          {!loading && (
+            <SettingsListItem className="!pb-12 !shadow-b">
```
AllanMichay: Remove !

src/core/router/SettingRoutes.tsx
```diff
@@ -236,6 +240,12 @@ export const settingRoutes: CustomRouteObject[] = [
         element: <TaxesSettings />,
         permissions: ['organizationTaxesView'],
       },
+      {
+        path: [GENERAL_SETTINGS_ROUTE],
+        private: true,
+        element: <OrganizationGeneralSettings />,
+        permissions: ['organizationView'],
+      },
```
AllanMichay, line 248: Add it as first element in the nav

src/pages/settings/OrganizationGeneralSettings/OrganizationGeneralSettings.tsx
domenicofalco: done

src/core/router/SettingRoutes.tsx
domenicofalco: we agreed on having it as last item of the menu but it makes more sense to have it as first. changed

AllanMichay (approved):

PR #3400, opened 2026-04-23
fix(subscription): correct downgrade date and add pending sub alert

ansmonjol:
## Context

Fixes [ISSUE-1845 — Wrong downgrade date displayed when downgrade is pending](https://linear.app/getlago/issue/ISSUE-1845/wrong-downgrade-date-displayed-when-downgrade-is-pending).

Depends on: https://github.com/getlago/lago-api/pull/5399 (new GraphQL fields on `Subscription`).

## Problem

On a subscription with a pending downgrade, the active subscription details page displayed the wrong effective date, e.g.:

> Downgrading to `lower_2` on **April 22, 2026**

April 22 is the day the subscription *started* (and the day the downgrade was scheduled). The downgrade actually takes effect on **May 22, 2026** (end of the current billing period + 1 day), which is already what the REST API returns as `downgrade_plan_date`.

The UI was reading `nextSubscriptionAt`, which the back-end resolves from the pending sub's `subscription_at` (the scheduling date), so the displayed date was off.

Additionally, the **pending** subscription view had no indication of when it would activate, so there was no way to see the incoming effective date from that side.

## Changes

- Extend `SubscriptionForSubscriptionInformations` fragment to query `downgradePlanDate`, `previousPlan { id name }` and `previousSubscription { id downgradePlanDate }` (all new fields exposed by the companion API PR).
- Fix the existing `"Downgrading to {planName} on {date}"` alert on the active-sub view to read `downgradePlanDate` instead of `nextSubscriptionAt`.
- Add a symmetric `"Downgrading from {planName} on {date}"` alert on the pending-sub view, using `previousPlan.name` and `previousSubscription.downgradePlanDate`. This addresses Anais' follow-up request in the ticket comments (no way to see when a pending sub activates).
- Regenerate `src/generated/graphql.tsx` via `pnpm codegen` against the updated API schema.
- Add the new translation key `text_1776951742342o96gqg8qg8j` via `pnpm translations:add`.

### DIFF
--- a/src/components/subscriptions/SubscriptionInformations.tsx
+++ b/src/components/subscriptions/SubscriptionInformations.tsx
@@ -28,11 +28,20 @@ gql`
     endingAt
     terminatedAt
     billingTime
+    downgradePlanDate
     nextSubscriptionAt
     nextSubscriptionType
     nextPlan {
       id
       name
+    }
+    previousPlan {
+      id
+      name
+    }
+    previousSubscription {
+      id
+      downgradePlanDate
     }
     customer {
       id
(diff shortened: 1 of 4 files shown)


src/components/subscriptions/SubscriptionInformations.tsx
```diff
@@ -85,21 +95,37 @@ export const SubscriptionInformations = ({
...
+      return translate('text_1776951742342o96gqg8qg8j', {
+        planName: subscription?.previousPlan?.name,
+        dateStartNewPlan: intlFormatDateTimeOrgaTZ(
+          subscription?.previousSubscription?.downgradePlanDate,
+        ).date,
+      })
+    }
+
+    return null
+  }, [subscription, translate, intlFormatDateTimeOrgaTZ])
```
AllanMichay: Should we maybe scope on `subscription.nextplan, subscription.nextSubscriptionType, subscription.downgradePlanDate, subscription.status, subscription.previousPlan, subscription.previousSubscription` ? I feel like using `subscription` here doesn't funnel down enough to allow useMemo to be helpful in terms of performance

AllanMichay (approved):

src/components/subscriptions/SubscriptionInformations.tsx
ansmonjol: Those could reference objects and the trade-off having a memo or alias them in a const above is almost null.

I think here i'll just drop the memo as the gain vs memo bookkeeping is certainly not crazy high (could even be lower, didn't check)

Thanks for your comment

ansmonjol: I've created a component for it, way cleaner

PR #3427, opened 2026-04-27
fix(org slug): QA Phase - prepend org slug when navigating from the devtools panel

domenicofalco:
## Context
Clicking resource links / action buttons in the devtools panel
(API keys, webhooks, activity logs) navigated to slug-less paths
and 404'd under the post-Phase-2 `/:organizationSlug/...` router.

## Description
- `RouteWrapper`: in the MemoryRouter → BrowserRouter bridge, derive
  the slug from the first segment of the current URL and pass it
  with `skipSlugPrepend: true`.
- `activityLogs/utils.tsx`: split into a pure `getResourceLink`
  (path mapping) + a thin `formatResourceObject(resource, onClick?)`
  renderer.
- `ActivityLogDetails`: replace raw `<a href>` with `<Button variant="inline">`
  wired to the existing `setMainRouterUrl + closePanel` bridge.

### DIFF
--- a/src/components/activityLogs/utils.ts
+++ b/src/components/activityLogs/utils.ts
@@ -1,6 +1,5 @@
 import { generatePath } from 'react-router-dom'
 
-import { Button } from '~/components/designSystem/Button'
 import { AvailableFiltersEnum, setFilterValue } from '~/components/designSystem/Filters'
 import { ACTIVITY_LOG_ROUTE } from '~/components/developers/devtoolsRoutes'
 import { ACTIVITY_LOG_FILTER_PREFIX } from '~/core/constants/filters'
@@ -137,26 +136,6 @@ export function getResourceLink(
     default:
       return null
   }
-}
-
-/**
- * Renders a resource as a clickable Button when an `onClick` is provided
- * (caller is responsible for the bridge to the main BrowserRouter), or as
- * plain text otherwise.
- */
-export function formatResourceObject(
-  resource: ActivityLogDetailsFragment['resource'],
-  onClick?: () => void,
-) {
-  if (!resource) return null
-
-  return onClick ? (
-    <Button variant="inline" onClick={onClick}>
-      {resource.id}
-    </Button>
-  ) : (
-    resource.id
-  )
 }
 
 export function buildLinkToActivityLog(activityId: string, filter?: AvailableFiltersEnum): string {
--- a/src/components/activityLogs/utils.tsx
+++ b/src/components/activityLogs/utils.tsx
@@ -1,5 +1,6 @@
 import { generatePath } from 'react-router-dom'
 
+import { Button } from '~/components/designSystem/Button'
 import { AvailableFiltersEnum, setFilterValue } from '~/components/designSystem/Filters'
 import { ACTIVITY_LOG_ROUTE } from '~/components/developers/devtoolsRoutes'
 import { ACTIVITY_LOG_FILTER_PREFIX } from '~/core/constants/filters'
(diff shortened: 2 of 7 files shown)


src/components/activityLogs/utils.tsx
```diff
@@ -77,77 +83,76 @@ export function formatResourceObject(
     resourceType?: keyof typeof ResourceTypeEnum
     activityType?: ActivityTypeEnum
   },
-) {
-  if (!resource) return null
+): string | null {
+  if (!resource || !activityType || isDeletedActivityType(activityType)) return null
```
ansmonjol: Shouldn't that be `&&` ?

domenicofalco: @ansmonjol that should be ok, I just grouped the already existing conditions in one line, so essentially they behave as before.

with && we'd early return only when all three hit at the same time, which shouldn't happen

I agree the mixed positive/negative reads a bit weird tho. happy to split it for clarity:

```
if (!resource) return null
if (!activityType || isDeletedActivityType(activityType)) return null
```

src/components/activityLogs/utils.ts
```diff
@@ -77,77 +83,77 @@ export function formatResourceObject(
     resourceType?: keyof typeof ResourceTypeEnum
     activityType?: ActivityTypeEnum
   },
-) {
+): string | null {
   if (!resource) return null
+  if (!activityType || isDeletedActivityType(activityType)) return null
```
AllanMichay, line 87: Thanks for this 👍

src/components/developers/activityLogs/ActivityLogDetails.tsx
```diff
@@ -226,40 +243,57 @@ export const ActivityLogDetails = ({ goBack }: { goBack: () => void }) => {
...
+                        resourceType: remapResourceTypeNames(resource.__typename),
+                        activityType,
+                      })
+
+                      return formatResourceObject(
+                        resource,
+                        link ? () => handleResourceNavigate(link) : undefined,
+                      )
+                    })()
                   : '-',
```
AllanMichay: Maybe would be better in a separate method

src/components/developers/activityLogs/ActivityLogDetails.tsx
```diff
@@ -24,6 +25,15 @@ import {
 import { useActivityLogsInformation } from '~/hooks/activityLogs/useActivityLogsInformation'
 import { useInternationalization } from '~/hooks/core/useInternationalization'
 import { useFormatterDateHelper } from '~/hooks/helpers/useFormatterDateHelper'
+import { useDeveloperTool } from '~/hooks/useDeveloperTool'
+
+export const ACTIVITY_LOG_DETAILS_LOADING_TEST_ID = 'activity-log-details-loading'
+export const ACTIVITY_LOG_DETAILS_CONTENT_TEST_ID = 'activity-log-details-content'
+export const ACTIVITY_LOG_DETAILS_CLOSE_BUTTON_TEST_ID = 'activity-log-details-close-button'
+export const ACTIVITY_LOG_DETAILS_RESOURCE_LINK_TEST_ID = 'activity-log-details-resource-link'
```
AllanMichay, line 32: `ACTIVITY_LOG_DETAILS_RESOURCE_LINK_TEST_ID` seem unused

src/components/developers/activityLogs/ActivityLogDetails.tsx
domenicofalco: fair. done, ty

src/components/developers/activityLogs/ActivityLogDetails.tsx
domenicofalco: done

AllanMichay (approved):

PR #3441, opened 2026-04-29
fix(org slug): Cover iframe multi-org case to auto-detect org slug

domenicofalco:
## Context

Added iframe slug recovery in `OrganizationLayout` for the post-auth flow. Testing against `getlago/lago-salesforce` surfaced two gaps that broke multi-org users on the "session expired → re-login" path:

1. Auth guard drops search params on `/login` (preexisting issue from #2838), so `useIframeConfig` loses iframe context and renders Google/Okta buttons.
2. Home only prepends the slug for single-membership users — multi-org in iframe fell through to the default redirect after login.

## Description

- `useLocationHistory`: propagate iframe params onto the `/login` URL.
- `Home/utils` (`resolveRedirectTarget`): extend the slug-less legacy branch to fire for any user when the URL carries iframe params
- `useIframeConfig`: extracted `hasIframeParams(search)` shared helper, reused across `OrganizationLayout`, `Home/utils`, `useLocationHistory`.

**Multi-org with LS pointing to a different org's slug still produces a 404, same UX as pre-migration, documented as known limitation. This needs a separated investigation.**

Fixes LAGO-1443

### DIFF
--- a/src/hooks/useIframeConfig.ts
+++ b/src/hooks/useIframeConfig.ts
@@ -13,6 +13,25 @@ type TUseIframeConfigReturn = {
   isRunningInSalesForceIframe: boolean
 }
 
+/**
+ * Pure check for iframe context query params (`?sfdc=true` for Salesforce,
+ * `?ifrm=true` for Hubspot). Use when you have a `search` string in hand
+ * (e.g. inside a pure resolver, or against a saved Location) and don't want
+ * to subscribe to React Router via `useIframeConfig`.
+ *
+ * Strict `=== 'true'` (not truthy) — defensive: the only valid embed values
+ * Salesforce/Hubspot send are `true`. Any other value is treated as "not
+ * iframe context" so we don't trigger LS-based slug auto-recovery on bad
+ * input.
+ */
+export const hasIframeParams = (search: string | undefined): boolean => {
+  if (!search) return false
+
+  const params = new URLSearchParams(search)
+
+  return params.get('sfdc') === 'true' || params.get('ifrm') === 'true'
+}
+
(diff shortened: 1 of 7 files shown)


src/hooks/useIframeConfig.ts
```diff
@@ -13,6 +13,25 @@ type TUseIframeConfigReturn = {
...
+ * `?ifrm=true` for Hubspot). Use when you have a `search` string in hand
+ * (e.g. inside a pure resolver, or against a saved Location) and don't want
+ * to subscribe to React Router via `useIframeConfig`.
+ *
+ * Strict `=== 'true'` (not truthy) — defensive: the only valid embed values
+ * Salesforce/Hubspot send are `true`. Any other value is treated as "not
+ * iframe context" so we don't trigger LS-based slug auto-recovery on bad
+ * input.
+ */
+export const hasIframeParams = (search: string | undefined): boolean => {
```
ansmonjol, line 27: Could be a boolean within the useIframeConfig so you already have the `isRunningInSalesForceIframe` and `isRunningInIframeContext` to rely to?
Not a blocker but felt producing less code, vars could be renamed too

ansmonjol: and it also keep the search param retrival part of the hook and not invocator

domenicofalco: hey @ansmonjol `hasIframeParams` is used in two non-React spots too (`resolveRedirectTarget` and `useLocationHistory`), so the hook can't fully replace it. Kept it as a shared helper and moving it into the hook on top would add code, not remove it

ansmonjol (approved):

PR #3449, opened 2026-04-30
feat(quote): allow quote edition

AllanMichay:
## Context

Last part of edition handling: allow quote edition

## Description

This PR
- adds quote edition page
- adds rte auto save
- adds sidebar element update

Fixes LAGO-1411
Fixes LAGO-1413


src/pages/quotes/hooks/useCreateQuote.ts
```diff
@@ -48,13 +65,26 @@ export const useCreateQuote = (): UseCreateQuoteReturn => {
   })
 
   const onSave = async (values: CreateQuoteValues): Promise<void> => {
+    if (!values.hasCustomerCurrency && values.currency && values.customerExternalId) {
+      await updateCustomerCurrency({
```
domenicofalco: how are errors handled here? does it simply fallback to the error boundaries if any?

src/pages/quotes/EditQuote.tsx
```diff
@@ -1,46 +1,227 @@
...
+    if (!quoteId) return
+    navigate(
+      generatePath(QUOTE_DETAILS_ROUTE, {
+        quoteId,
+        tab: QuoteDetailsTabsOptionsEnum.overview,
+      }),
+    )
+  }
+
+  type SaveStatus = 'idle' | 'saving' | 'saved' | 'error'
```
domenicofalco: this can be put outside the component

src/pages/quotes/editQuote/EditQuoteAside.tsx
```diff
@@ -0,0 +1,306 @@
...
+  onSaveFinished?: () => void
+  onSaveError?: (payload: UpdateQuoteVersionInput) => void
+}) => {
+  const { translate } = useInternationalization()
+  const { updateQuoteVersion } = useUpdateQuote({ onUpdateFinished: onSaveFinished })
+
+  const hasSubscription = !!quote.subscription
+  const versionId = quote.currentVersion.id
+
+  const billingItems = quote.currentVersion.billingItems as
```
domenicofalco: should we have a custom type for `quote.currentVersion.billingItems`? the graphql returns any for this one and we can think of centralize its type locally

domenicofalco (approved):

src/pages/quotes/hooks/useCreateQuote.ts
AllanMichay: You're right, I'll check how to handle them. I think since we have validation, most of the errors will fall into the error boundaries

src/pages/quotes/editQuote/EditQuoteAside.tsx
AllanMichay: I think so, I'll sync with back end to have a typing there

PR #3500, opened 2026-05-14
feat(subscription): Add invoice consolidation control

domenicofalco:
## Context
Adds a per-subscription `consolidate_invoice` boolean (default `true`) that opts a subscription invoice
consolidation. The FE needs to surface this toggle on subscription create/edit and on the subscription detail page.

## Description
- Add `consolidateInvoice` to the subscription form schema, defaults, GraphQL query / fragment / create + update mutations.
- New reusable `SubscriptionInvoiceConsolidationSection` rendering a radio group (Consolidate / Isolate) with explanatory sublabels.
- Restructure the "Billing settings" section so its title + subtitle and the consolidation block are always visible, while `PaymentMethodsInvoiceSettings` stays gated behind the multi-payment feature flag.
- Regenerate GraphQL types; update translations (orphan old section title removed, 7 new keys added).

Fixes ING-130


src/components/subscriptions/SubscriptionInvoiceConsolidationSection.tsx
```diff
@@ -0,0 +1,48 @@
...
+type SubscriptionInvoiceConsolidationSectionProps = {
+  value: boolean
+  onChange: (value: boolean) => void
+  name?: string
+}
+
+const CONSOLIDATE = 'consolidate'
+const ISOLATE = 'isolate'
+
+export const SubscriptionInvoiceConsolidationSection = ({
```
AllanMichay: Why not use the `withForm` HOF?

AllanMichay (approved):

PR #3502, opened 2026-05-15
feat: Presentation group keys

stephenlago99:
## Context

  The backend now exposes per-fee **presentation group keys** (a way to break billable metric units down by dynamic attributes like region, type, etc. — see getlago/lago-api#5492). On the front end we need to:

  - surface those breakdowns wherever a fee or charge is rendered (invoices, current usage, usage detail drawer);
  - give users a richer, read-only way to inspect a fee that isn't tied to the draft-only `EditFeeDrawer`;
  - make the current-usage view more glanceable so customers can see the open billing period and totals without drilling in.

  While in that area we also fixed a long-standing bug where zero-amount fees on finalized invoices rendered as a no-fee placeholder and weren't inspectable.

  ## Description

  ### New `ViewFeeDetailsDrawer`

  - New read-only drawer built on the `useDrawer` hook (same pattern as `SubscriptionFeeDrawer` / `FeatureEntitlementDrawer`), portaled through NiceModal with the canonical 48px content padding.
  - Two variants driven by the data: single-view (Overview only) and tabbed (Overview + Presentation group keys table). Tabs run inside a `MemoryRouter` so `NavigationTab` works outside the app's `BrowserRouter`.
  - Drawer title is the static "Fee details" string; the in-body header shows a fee-type-aware label — subscription fees show *"{{interval}} subscription fee - {{plan}}"*, everything else (charge, fixed_charge, add_on, commitment, credit) falls back to `invoiceDisplayName || itemName`.
  - Body is keyed by `fee.id`, so opening a second fee from the same page fully remounts the content (no stale plan name, no stale `tabIndex`).
  - New `FeeForViewFeeDetailsDrawer` fragment covers all fields shown in the Overview grid (precise_amount_cents, total_amount_cents, taxes, source-item info, etc.).

  ### Invoice fee rows are clickable + action menu

  - Every fee row in `InvoiceDetailsTable` and its five variants (Graduated, GraduatedPercentage, Volume, Percentage, Package) is now clickable and opens the drawer. Hover/cursor state on rows backed by a real fee; placeholder rows stay inert.
  - New `FeeActionsCell` (3-dots) on every clickable row with "Copy fee ID" + "View fee details". The cell `stopPropagation`s click events so the menu doesn't reopen the drawer.
  - **Bug fix**: finalized invoices containing only zero-amount fees previously fell into the no-fee placeholder branch (gate was `!hasAnyPositiveFeeParsed`), which passed `fee={undefined}` to `BodyLine` and broke clickability. Unified the gate to `!hasAnyFeeParsed` for all statuses so zero-amount fees go through the
  regular path. `hasAnyPositiveFeeParsed` is still computed (test fixtures depend on it) but no longer drives UI.
  - Fixed the period-row `colSpan` to account for the new action column; removed the dead `isDraftInvoice` prop from `InvoiceDetailsTableHeader` and the five variants.
  - `viewFeeDetailsDrawerRef` is optional on `InvoiceDetailsTableProps` — the regenerate flow and the integration tests don't wire it, and `openViewFeeDetailsDrawer` already guards on `ref?.current?.openDrawer(...)`.

  ### `SubscriptionUsageDetailDrawer` (presentation breakdowns)

  - Split into two sections: **Usage linked to the charge** (always rendered, includes the charge summary row + presentation breakdowns under it) and **Usage linked to charge filters and pricing groups** (filters / grouped-usage tables, or an info alert when neither exists).
  - Breakdown rows are aggregated client-side by a stable `presentationBy` key, so identical breakdowns across multiple fees don't over-count. On the Projected tab breakdowns are hidden and raw GraphQL units are shown as-is.
  - Drawer title is *"Usage of {{billableMetricName}}"*; main column header is now "Charge name". 48px gap between the tabs and the first section.
  - Charge summary row matches the table format: `charge.invoiceDisplayName || billableMetric.name` as the title with `billableMetric.code` as the subtitle. (`code` added to both `CustomerUsageForUsageDetails` and `CustomerProjectedUsageForUsageDetails` fragments.)

  ### `SubscriptionCurrentUsageTable` (usage page header + row click)

  New summary panel between the section title and the tabs:

  | Current billing period | Usage so far | Projected end of period (premium) |

  - Columns are natural-width, separated by vertical `divide-x` dividers with 32px of horizontal padding on each side. Titles `captionHl/grey600`, values `bodyHl/grey700`. Panel has a bottom shadow.
  - Date range format is plain `"Jul 1, 2026 - Jul 31, 2026"`.
  - "Usage so far" is `usageData.amountCents` (works for both queries since both return current usage as `amountCents`).
  - "Projected end of period" only renders when the org has the premium projected-usage flag **and** `usageData.projectedAmountCents` is populated, which is whatever the existing `useProjectedUsageForSubscriptionUsageQuery` has already loaded. No extra API call is fired from the Current tab — pending alignment with
  Anna on whether the projected value should always be eagerly loaded.

  Other usage-page changes:

  - Removed the "Show breakdown" CTA; rows are now clickable and open the detail drawer directly (added `containerClassName="[&_tbody>tr>td]:px-1"` so hover states don't sit flush against the table edges).
  - Replaced the per-row breakdown link with inline metadata pills under the charge name: `billableMetric.code` + "priced per" filter pills + "split by" grouped-by-key pills.
  - Header bar: subtitle *"Usage data for this subscription in the current open billing period."* under the "Usage" title; reload icon replaced with a right-aligned "Refresh" text button vertically centered between the title and subtitle.
  - Removed the "Total current usage" row above the table (the value now lives in the summary panel) and the unused `showExcludingTaxLabel` prop (incl. customer portal caller).

  ### Tests

  - `FeeActionsCell.test.tsx` — menu rendering, button wiring, click-event stop-propagation (14 tests).
  - `ViewFeeDetailsDrawer.test.tsx` — drawer open/close, single vs. tabbed variants, parent-id row rendering (11 tests).
  - `InvoiceDetailsTableBodyLine.clickable.integration.test.tsx` — row click wiring across body-line variants.
  - `SubscriptionUsageDetailDrawer.test.tsx` — `makeBreakdownRows` aggregation, `sumBreakdownUnits`, `isBreakdownRow` (19 tests).
  - Snapshots updated for the new period-line `colSpan` and action-column widths.
  - Cypress: `cypress/e2e/10-resources/t80-invoice-fee-details.cy.ts` (skipped pending fixtures).


src/components/customers/usage/SubscriptionUsageDetailDrawer.tsx
AllanMichay: What do you think of moving this whole graphQL query into it's own hook?

src/components/customers/usage/SubscriptionUsageDetailDrawer.tsx
```diff
@@ -340,203 +439,472 @@ export const SubscriptionUsageDetailDrawer = forwardRef<
...
+            // resolve the entire ChargeUsage object graph.
+            type ChargeSummaryRow = {
+              id: string
+              units: number | string
+              projectedUnits?: number | string
+              amountCents?: number | string
+              projectedAmountCents?: number | string
+              pricingUnitAmountCents?: number | string
+              pricingUnitProjectedAmountCents?: number | string
+            }
```
AllanMichay: I'd rather have this type at the start of the file / in it's own file instead of right in the middle of the render

src/components/customers/usage/SubscriptionUsageDetailDrawer.tsx
```diff
@@ -340,203 +439,472 @@ export const SubscriptionUsageDetailDrawer = forwardRef<
...
+              ? [chargeSummaryRow]
+              : [chargeSummaryRow, ...makeBreakdownRows('charge', usage?.presentationBreakdowns)]
+
+            return (
+              <Table
+                name="charge-summary-table"
+                containerSize={0}
+                rowSize={!!pricingUnitShortName ? 72 : 48}
+                data={summaryData}
+                columns={[
```
AllanMichay: Could we move this out of the render? You can take inspiration of how to change it in ActivityLogsTable (it uses `TableColumn` type)

src/components/customers/usage/SubscriptionUsageDetailDrawer.tsx
```diff
@@ -340,203 +439,472 @@ export const SubscriptionUsageDetailDrawer = forwardRef<
...
+
+          {hasAnyFilterInGroupUsage && (
+            <div className="[&_table:not(#table-grouped-usage-with-filters-table-0)_thead]:hidden">
+              {/* NOTE: We have to make a copy of the array here, otherwise we got an error after usage reload while opening the Drawer */}
+              {[...(usage?.groupedUsage || [])]
+                ?.sort((a, b) => {
+                  if (a.filters?.some((f) => !f.id)) return -1
+                  if (b.filters?.some((f) => !!f.id)) return 1
+                  return 0
+                })
```
AllanMichay: I think we should move this copy + sort out of the render

src/components/customers/usage/SubscriptionUsageDetailDrawer.tsx
AllanMichay, line 1: Overall I think this file is a bit hard to read with all those Table. The best way to handle this IMO would be to separate seach section into separate files

src/components/invoices/details/ViewFeeDetailsDrawer.tsx
AllanMichay, line 1: I think since it's a new Drawer, we should use our new Drawer system (You can get examples in `src/pages/__devOnly/tabs/DrawerTest.tsx`)

src/components/invoices/details/FeeActionsCell.tsx
```diff
@@ -0,0 +1,125 @@
...
+  FEE_ACTIONS_CELL_TEST_ID,
+  FEE_COPY_ID_BUTTON_TEST_ID,
+  FEE_VIEW_DETAILS_BUTTON_TEST_ID,
+} from './invoiceDetailsTestIds'
+
+type Fee = FeeForViewFeeDetailsDrawerFragment | null | undefined
+
+export const openViewFeeDetailsDrawer = (
+  fee: Fee,
+  ref: RefObject<ViewFeeDetailsDrawerRef> | undefined,
```
AllanMichay: With the new drawer system, we wouldn't need the ref props for the drawer. Check my comment on `ViewFeeDetailsDrawer`

src/components/invoices/details/InvoiceDetailsTable.tsx
```diff
@@ -206,21 +208,27 @@ export const InvoiceTableSection: FC<{
...
     )
   } else if (canHaveUnitPrice) {
     tableStructureClasses = tw(
-      '[&_table>thead>tr>th:nth-child(1)]:w-[45%] [&_table>thead>tr>th:nth-child(2)]:w-[15%] [&_table>thead>tr>th:nth-child(3)]:w-[15%] [&_table>thead>tr>th:nth-child(4)]:w-[10%] [&_table>thead>tr>th:nth-child(5)]:w-[15%]',
-      '[&_table>tbody>tr>td:nth-child(1)]:w-[45%] [&_table>tbody>tr>td:nth-child(2)]:w-[15%] [&_table>tbody>tr>td:nth-child(3)]:w-[15%] [&_table>tbody>tr>td:nth-child(4)]:w-[10%] [&_table>tbody>tr>td:nth-child(5)]:w-[15%]',
+      '[&_table>thead>tr>th:nth-child(1)]:w-[45%] [&_table>thead>tr>th:nth-child(2)]:w-[15%] [&_table>thead>tr>th:nth-child(3)]:w-[15%] [&_table>thead>tr>th:nth-child(4)]:w-[10%] [&_table>thead>tr>th:nth-child(5)]:w-[15%] [&_table>thead>tr>th:nth-child(6)]:w-6 [&_table>thead>tr>th:nth-child(6)]:overflow-visible',
+      '[&_table>tbody>tr>td:nth-child(1)]:w-[45%] [&_table>tbody>tr>td:nth-child(2)]:w-[15%] [&_table>tbody>tr>td:nth-child(3)]:w-[15%] [&_table>tbody>tr>td:nth-child(4)]:w-[10%] [&_table>tbody>tr>td:nth-child(5)]:w-[15%] [&_table>tbody>tr>td:nth-child(6)]:w-6 [&_table>tbody>tr>td:nth-child(6)]:overflow-visible',
+      actionColumnClasses,
```
AllanMichay, line 223: Woaw, this looks so complex. Let's leave it like this right now but I'd love to see if there is a way to handle this more gracefully

src/components/plans/chargeAccordion/ChargeWrapperSwitch.tsx
```diff
@@ -55,14 +59,17 @@ export const ChargeWrapperSwitch = memo(
...
+            <ChargePercentage premiumWarningDialogRef={premiumWarningDialogRef} />
+          )}
           {localCharge?.chargeModel === ALL_CHARGE_MODELS.Volume && <VolumeChargeTable />}
           {localCharge?.chargeModel === ALL_CHARGE_MODELS.Custom && (
             <CustomCharge onExpandCustomCharge={onExpandCustomCharge} />
           )}
           {localCharge?.chargeModel === ALL_CHARGE_MODELS.Dynamic && <DynamicCharge />}
 
           {isUsageCharge && <PricingGroupKeys />}
+          {isUsageCharge && <PresentationGroupKeys />}
```
AllanMichay: Should we regroup those? (Question from Alex on the first PR)

src/components/customers/usage/SubscriptionUsageDetailDrawer.tsx
AllanMichay: In the end, if we decide to have a file for each Table, no need to extract that

src/components/customers/usage/sections/AmountCentsCell.tsx
```diff
@@ -0,0 +1,57 @@
...
+          pricingUnitShortName,
+        },
+      )}
+    </Typography>
+
+    {!!pricingUnitShortName && (
+      <Typography variant="caption" color="grey600">
+        {intlFormatNumber(
+          deserializeAmount(
+            showProjected ? row.amountCents : (row as ProjectedChargeUsage).projectedAmountCents,
```
AllanMichay: Shouldn't it be the other way around?

src/components/customers/usage/sections/GroupedUsageWithFiltersTable.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+// Sort groupedUsage entries so groups whose filters include a filter with no
+// `id` come first — the no-id filter represents the "all other values" bucket
+// and reads more clearly at the top.
+const sortGroupedUsage = (
+  groupedUsage: NonNullable<SubscriptionUsageDetailDrawerUsage['groupedUsage']>,
+) =>
+  [...groupedUsage].sort((a, b) => {
+    if (a?.filters?.some((f) => !f?.id)) return -1
+    if (b?.filters?.some((f) => !!f?.id)) return 1
```
AllanMichay: is it normal to check no filter here?

AllanMichay (approved):

PR #3504, opened 2026-05-15
feat(parent-children-hierarchy): Add multi billing & multi currency flow

domenicofalco:
## Context

  The Lago platform is introducing support for multi-billing entity and
  multi-currency operations within the parent/children hierarchy. These
  capabilities are gated behind two feature flags: `MultiEntityBilling`
  and `MultiCurrency`. When enabled, customers can have invoices, credit
  notes, payments, subscriptions, and wallets scoped to different billing
  entities and currencies simultaneously.

  ## Description

  - Add `BillingEntityFormPicker` and `BillingEntityLabel` components for
    selecting and displaying billing entities across forms and detail views
  - Add `CustomerInvoiceBalancesBreakdown` and `CustomerCreditNotesBreakdown`
    tables that aggregate balances per (currency × billing entity) pair,
    replacing legacy single-value cards when feature flags are enabled
  - Add `CurrencyPicker` and `LockedPickerBox` form components for
    currency selection and premium-gated UI placeholders
  - Add `FiltersItemBillingEntityId` filter element and extend filter
    utilities with customer-scoped query formatters for credit notes,
    invoices, and payments
  - Add `useCustomerFilterDefaults` and `useBillingEntitiesOptions` hooks
    to derive filter configuration and billing entity options from feature
    flags and organization data
  - Update `CustomerOverview`, `CustomerCreditNotesList`,
    `CustomerInvoicesTab`, `CustomerPaymentsTab`, `SubscriptionsPage`,
    and `CustomerRequestOverduePayment` to conditionally render breakdown
    tables vs legacy cards based on feature flags, with scoped filters
  - Integrate billing entity selection into `CreateSubscription`,
    `CreateInvoice`, `CreateWallet`, and `CreateCustomer` forms
  - Add `MonthSelectorDropdown` premium gating with `LockedPickerBox`


src/components/customers/overview/CustomerInvoiceBalancesLegacyCards.tsx
```diff
@@ -0,0 +1,199 @@
...
+    [lastPaymentRequestCreatedAt],
+  )
+  const hasMadePaymentRequestToday = isSameDay(lastPaymentRequestDate, today)
+
+  return (
+    <Stack gap={4}>
+      {hasOverdueInvoices &&
+        !overdueBalancesError &&
+        !isPaymentProcessingStatusLoading &&
+        isCustomerReadyForOverduePayment && (
```
ansmonjol: Could it be a unique boolean bind above the render? Feels bloated here

src/hooks/useIsCustomerReadyForOverduePayment.ts
```diff
@@ -25,7 +25,7 @@ export const useIsCustomerReadyForOverduePayment =
     const { data, loading, error } = useGetCustomerOverdueInvoicesReadyForPaymentProcessingQuery({
       variables: { id: customerId ?? '' },
       skip: !customerId,
-      fetchPolicy: 'network-only',
+      fetchPolicy: 'cache-and-network',
```
ansmonjol: Isn't it the default for queries?

src/components/customers/usage/CustomerUsage.tsx
```diff
@@ -45,12 +58,21 @@ export const CustomerUsage = () => {
...
+            <Filters.Component />
+          </div>
+        </Filters.Provider>
+      )}
+
       <AnalyticsStateProvider>
         <Gross
           className="analytics-graph py-0"
-          currency={data?.customer?.currency || organization?.defaultCurrency}
+          currency={currency ?? CurrencyEnum.Usd}
```
ansmonjol, line 72: I think we should keep `organization?.defaultCurrency` no? in case the feature is off we lose this info.
Yet as it's a filter and only manually triggered it probably don't make sense and the default is just undefined, so what is expected for a filter 🤔 
Just felt too simple and may be worth double checking on usage

src/components/customers/CustomerInvoicesTab.tsx
```diff
@@ -54,30 +72,47 @@ interface CustomerInvoicesTabProps {
...
+  )
+
+  const [getDraftInvoices, { data: dataDraft, error: errorDraft, loading: loadingDraft }] =
+    useGetCustomerInvoicesLazyQuery({
+      variables: {
+        customerId,
+        limit: DRAFT_INVOICES_ITEMS_COUNT,
+        status: [InvoiceStatusTypeEnum.Draft],
+      },
+    })
```
ansmonjol: Could we leverage usage of `skip` on a non-lazy query here?
Lazy queries loses the advantage of caching. Not an issue but worth being sure we're aware of such

src/components/customers/CustomerInvoicesTab.tsx
```diff
@@ -54,30 +72,47 @@ interface CustomerInvoicesTabProps {
...
-      status: [InvoiceStatusTypeEnum.Draft],
-    },
+  const draftFiltersProps = useCustomerFilterDefaults({
+    filtersNamePrefix: CUSTOMER_INVOICES_DRAFT_FILTER_PREFIX,
+    include: ['currency', 'entity'],
   })
+  const finalizedFiltersProps = useCustomerFilterDefaults({
+    filtersNamePrefix: CUSTOMER_INVOICES_FINALIZED_FILTER_PREFIX,
+    include: ['currency', 'entity'],
+  })
```
ansmonjol: What if we had only one call to `useCustomerFilterDefaults`?

This method seems to re-return the `filtersNamePrefix` without any transformation, hence them being available in the return if i'm correct.

So we could have this?
```
const res = useCustomerFilterDefaults({
  include: ['currency', 'entity']
})

const draftFiltersProps = { filtersNamePrefix: CUSTOMER_INVOICES_DRAFT_FILTER_PREFIX, ...res }
const finalizedFiltersProps = { filtersNamePrefix: CUSTOMER_INVOICES_FINALIZED_FILTER_PREFIX, ...res }
...

<Filters.Provider {{...draftFiltersProps}}>

...

<Filters.Provider {...finalizedFiltersProps}>
```

There could be drawback to the refactoring so i'm not seeing this as a hard blocker if the calls are or not too heavy.
Just felt weird having 2 invocations of the same method

src/hooks/useIsCustomerReadyForOverduePayment.ts
domenicofalco: i've put `cache-and-network` intentionally to have a faster first render

src/components/customers/usage/CustomerUsage.tsx
domenicofalco: `CurrencyEnum.Usd` is just a safety net, `currency` is being resolved already with `customerCurrency || organization?.defaultCurrency` in useCustomerFilterDefaults

src/components/customers/CustomerInvoicesTab.tsx
domenicofalco: good catch! fixing

src/components/customers/overview/CustomerInvoiceBalancesLegacyCards.tsx
domenicofalco: done

src/components/customers/CustomerInvoicesTab.tsx
domenicofalco: makes sense!

domenicofalco: although `filtersNamePrefix` is mandatory for `useCustomerFilterDefaults` and therefore I have to pass it with a default and then override as per your above code

src/hooks/useIsCustomerReadyForOverduePayment.ts
ansmonjol: yeah but not setting it will result the same 
check `src/core/apolloClient/init.ts:310`

src/components/customers/usage/CustomerUsage.tsx
ansmonjol: I missed it yeah

ansmonjol (approved):

src/hooks/useIsCustomerReadyForOverduePayment.ts
domenicofalco: removed

PR #3529, opened 2026-05-20
feat(plans): v2 shared scaffolding + cascade hook migration

ansmonjol:
Fixes LAGO-1467

Ships shared primitives for the Plan/Sub v2 inline-edit initiative (LAGO-1269):
- \`PlanDetailsV2\` page composer + \`getPlanForDetailsV2\` query
- \`PlanDetailsV2LeftSidebar\` (sticky, scroll-spy via IntersectionObserver, sub-flow gating)
- \`SectionAccordion\`, \`SectionHeader\`, \`cacheHelpers\`
- \`CascadeUpdatesField\` (TanStack toggle) + \`useCascadeFormDialog\`

Also folds in cleanup C3 (LAGO-1479): migrates \`CreatePlan.tsx\` off the legacy Formik \`ImpactOverriddenSubscriptionsDialog\` onto the new hook + V2 toggle UX, deletes the legacy dialog.

Behind \`EDIT_DETAILS_PAGE\` feature flag. Section bodies are skeleton placeholders — real content lands in LAGO-1468+.

Drift test: \`PlanDetailsV2LeftSidebar.test.tsx\` locks in the sub-flow gating contract (Progressive billing + Entitlements hidden when \`isInSubscriptionForm=true\`).

Cascade UX change: default flips from \`false\` (radio) to \`true\` (toggle) per spec §4.4.


src/components/plans/details-v2/shared/cacheHelpers.ts
```diff
@@ -0,0 +1,60 @@
+import { ApolloCache, Reference } from '@apollo/client'
```
domenicofalco: this helper looks not very specific to the plan-details, even tho it's been using here. I am not sure if this is the correct path for it? we already have some cache utils within `src/core/apolloClient/`, worth checking if something from there can be reused or this `cacheHelpers.ts` can be moved in there for consistency maybe. wdyt?

domenicofalco (approved):

src/components/plans/details-v2/PlanDetailsV2.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+const PLAN_SECTION_IDS: PlanDetailsV2SectionId[] = [
+  PlanDetailsV2SectionId.PlanSettings,
+  PlanDetailsV2SectionId.SubscriptionFee,
+  PlanDetailsV2SectionId.FixedCharges,
+  PlanDetailsV2SectionId.UsageCharges,
+  PlanDetailsV2SectionId.MinimumCommitment,
+  PlanDetailsV2SectionId.ProgressiveBilling,
+  PlanDetailsV2SectionId.Entitlements,
+]
```
AllanMichay: I don't see advanced settings here but I can see it in the left sidebar

src/components/plans/details-v2/PlanDetailsV2.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+const SUB_FLOW_HIDDEN_SECTIONS = new Set<PlanDetailsV2SectionId>([
+  PlanDetailsV2SectionId.ProgressiveBilling,
+  PlanDetailsV2SectionId.Entitlements,
+])
+
+type PlanDetailsV2Props = {
+  planId: string
+  isInSubscriptionForm?: boolean
+  subscriptionId?: string
```
AllanMichay: Seems unused

src/components/plans/details-v2/PlanDetailsV2LeftSidebar.tsx
```diff
@@ -0,0 +1,162 @@
...
+            {item.children.map((child) => renderItem(child, depth + 1))}
+          </div>
+        )}
+      </div>
+    )
+  }
+
+  return (
+    <nav
+      className="sticky top-0 flex h-screen w-64 flex-col gap-1 border-r border-grey-300 pr-4 pt-4"
```
AllanMichay, line 156: should we add auto overflow?

src/components/plans/details-v2/shared/cacheHelpers.ts
ansmonjol: Probably! Will look into that

src/components/plans/details-v2/PlanDetailsV2.tsx
ansmonjol: Not sure to understand the question.

This list is about the "real" sections of the page, while Advanced settings is a group for 3 of those here and is mostly a visual grouping done at invocation.

Does it make sense?

src/components/plans/details-v2/PlanDetailsV2.tsx
ansmonjol: Indeed! removing

src/components/plans/details-v2/PlanDetailsV2LeftSidebar.tsx
ansmonjol: What would be the advantage? The nav is sticky so if i'm correct is should have not impact

src/components/plans/details-v2/shared/cacheHelpers.ts
ansmonjol: Good catch — checked the existing utilities:

- `mergePaginatedCollection` / `createPaginatedFieldPolicy` operate on the paginated `{ metadata, collection }` shape (Apollo field policies, not ad-hoc mutations).
- `evictFromCache` handles deletes from those same paginated collections.

The v2 helpers target plain `Array<T>` fields (`Plan.charges`, `Plan.fixedCharges`, etc.) — non-paginated, granular CRUD. No overlap with what's there.

Moved them to `src/core/apolloClient/cacheHelpers.ts` alongside the existing ones (commit `b9587ec2`). Added JSDoc on both pointing at `createPaginatedFieldPolicy` / `evictFromCache` for the paginated-collection cases.

src/components/plans/details-v2/PlanDetailsV2.tsx
AllanMichay: Oh yes, makes total sense now. I understand :+1:

src/components/plans/details-v2/PlanDetailsV2LeftSidebar.tsx
AllanMichay: I'm just thinking in case the page height isn't high enough

ansmonjol: I tried! All good, the scroll still happen and sidebar behaves the same, just within a smaller window 😆

PR #3545, opened 2026-05-22
feat(plans): v2 Fixed charges section + CRUD

ansmonjol:
Fixes LAGO-1470

Mounts the Fixed charges section in `PlanDetailsV2`. Full CRUD via granular
`createFixedCharge` / `updateFixedCharge` / `destroyFixedCharge` mutations.
Cascade dialog opens per-mutation when `plan.hasOverriddenPlans=true`.

No virtualization — list renders as plain `.map()`. `content-visibility: auto`
added on `SectionAccordion` and `PlanDetailsV2LeftSidebar` rows for off-screen
layout/paint avoidance. Spec §4.8 virtualization deliberately skipped per
locked-in initiative decision (max fixed charges per plan in prod = 27).

Stacks on `lago-1470-extract` (`FixedChargeInfo` extraction).


src/components/plans/details-v2/shared/SectionAccordion.tsx
```diff
@@ -39,65 +36,73 @@ export const SectionAccordion = ({
...
+                          variant="quaternary"
+                          align="left"
+                          fullWidth
+                          onClick={(e) => {
+                            e.stopPropagation()
+                            e.preventDefault()
+                            closePopper()
+                            action.onClick()
+                          }}
+                        >
```
AllanMichay, line 92: We don't use action.danger anymore?

src/components/plans/details-v2/SubscriptionFeeAccordion.tsx
AllanMichay, line 1: Missing tests. It seems they were deleted

src/components/plans/details-v2/PlanDetailsV2FixedChargesSection.tsx
```diff
@@ -0,0 +1,195 @@
...
+      {fixedCharges.map((fixedCharge, index) => (
+        <SectionAccordion
+          key={fixedCharge.id}
+          title={fixedCharge.invoiceDisplayName || fixedCharge.addOn.name}
+          subtitle={fixedCharge.addOn.code}
+          actions={[
+            {
+              label: translate('text_63e51ef4985f0ebd75c212fc'),
+              onClick: () => openEdit(toLocalInput(fixedCharge), index),
+              hidden: isInSubscriptionForm,
```
AllanMichay: Don't we have plansUpdate permission?

src/components/plans/details-v2/shared/SectionAccordion.tsx
ansmonjol: no it was an oversight during initial implementation and adding it now would mean inconsistency across the app so i (sadly) preferred removing it

src/components/plans/details-v2/PlanDetailsV2FixedChargesSection.tsx
ansmonjol: We do not for this.
Please note I've also forgot to rely on `usePermissions` for plan actions in those pages (planCreate, planUpdate). Good Side catch 😆 

I'm gonna add it to the code but worth mentioning:
- Those conditions are "just" about calling the `usePermissions` and I don't want to add another wrapper on a hook call.
- The isInSubscriptionForm are contextual to plan used in subscription form, so not attached to the plan model logic, and not all fields uses it as some can be edited in this context. For the same reason I'm not willing to make a separate logic for this.

Please note I've a followup cleanup task about moving this `isInSubscriptionForm` props drilling into a context and I preffered to delay this as it's already drilled today.

src/components/plans/details-v2/SubscriptionFeeAccordion.tsx
ansmonjol: Added

AllanMichay (approved):

PR #3558, opened 2026-05-26
feat(order-forms): add actions on order forms

AllanMichay:
## Context

We are displaying order forms from their quotes

## Description

This PR
- Adds download order forms action
- Adds void order form action

Fixes LAGO-1427
Fixes LAGO-1589
Fixes LAGO-1571


src/components/designSystem/RichTextEditor/common/printHtmlContent.ts
```diff
@@ -1,11 +1,15 @@
...
+  // Off-screen, sized to A4 so the content lays out at a sane print width.
+  // (Pagination itself is driven by the html/body overflow reset below, not by
+  // these dimensions.)
   iframe.style.position = 'fixed'
   iframe.style.left = '-9999px'
   iframe.style.top = '0'
-  iframe.style.width = '0'
-  iframe.style.height = '0'
+  iframe.style.width = '794px' // ~A4 width @96dpi (210mm)
+  iframe.style.height = '1123px' // ~A4 height @96dpi (297mm)
```
ansmonjol, line 11: If you resize the screen it breaks no? Cannot those be in percent?

AllanMichay: This is an iframe used only when the user asks for download pdf. The idea is to mimic as much as possible the pdf format, thus the page is constricted to a size similar to A4

Not sure I understand the comment

ansmonjol (approved):

PR #3564, opened 2026-05-27
fix(invoice): Change Regenerate invoice components to retrieve preview invoice

tiagolupepic:
This commit updates the CustomerInvoiceRegenerate and CreateInvoice to use a draft invoice query.

Recently, we detected a issue with a customer.

> We have a current problem related to taxes applied in Invoice, Fee and its relations, applied_taxes.
These taxes are generated during the invoice generation and they keep persisted, they're based
on customer data, when the customer data changes, the previous applied taxes are not affected.

However, we have a scenario when customer changes their taxes and they want to regenerate invoices.
During the process to display the invoice information, frontend ask the invoice details using the invoice GraphQL query and as expected, we're still showing the applied taxes based on previous customer information.

We have a https://github.com/getlago/lago-api/pull/5237 and introducing the GraphQL query and,
in this PR we're changing the CustomerInvoiceRegenerate and CreateInvoice to use the "preview" state.

Also, we're removing the extra query of fees because the fees are nested in the invoice draft query
and we don't need an extra query.

Dependency: https://github.com/getlago/lago-api/pull/5237


src/pages/CustomerInvoiceRegenerate.tsx
```diff
@@ -82,6 +81,19 @@ gql`
...
+      ...AllInvoiceDetailsForCustomerInvoiceDetails
+
+      fees {
+        ...FeeDetailsForInvoiceOverview
+        ...FeeForInvoiceDetailsTable
+        ...FeeForInvoiceDetailsTableFooter
+      }
+    }
+  }
+
```
AllanMichay: This query seem to be used in both pages.

Please
- Create a hook containing this query
- use this hook on the two pages

You can find an example on how to handle and use it in `src/pages/creditNoteDetails/common/useDownloadCreditNote.ts` or `src/pages/createCustomers/common/useTaxProviders.ts`

tiagolupepic: thanks! I've moved to a hook

AllanMichay (approved):

PR #3568, opened 2026-05-27
feat(plans): v2 Advanced settings section

ansmonjol:
Fixes LAGO-1472

Adds the v2 Plan-tab "Advanced settings" section (Minimum commitment, Progressive billing, Entitlements) — read-only accordions with drawer-based inline edit, cascade-to-children, and non-premium paywalls. Behind the `EDIT_DETAILS_PAGE` flag. Mirrors the Subscription-fee pattern from #3538.

## Hook change worth flagging

`useUpdatePlanWithCascade` gains `includeAdvancedFields?: boolean` (default `false`). When unset (PlanSettings + SubFee callers), the produced `UpdatePlanInput` is byte-identical to before. When set, the hook persists `minimumCommitment` / `usageThresholds` / `entitlements`.

`buildUpdatePlanFormDefaults` now also hydrates `trialPeriod` / `payInAdvance` / `invoiceDisplayName` from the plan (previously hardcoded `0` / `false` / `undefined`). Required so per-section saves don't wipe sibling fields — and incidentally fixes a latent clobber where saving Plan settings was resetting `trialPeriod` / `payInAdvance` on the plan.

## Drawer change

`MinimumCommitmentDrawer` / `ProgressiveBillingDrawer` / `FeatureEntitlementDrawer` widen `onSave` to `void | boolean | Promise<void | boolean>` and close only when the result `!== false` (the same backward-compatible tweak #3538 made to `SubscriptionFeeDrawer`). Existing plan-form callers return `void` so they still close.

## Pattern per accordion

- Self-contained sub-section: paywall / read accordion / add. Owns its `useUpdatePlanWithCascade` instance and reuses the existing inner drawer for edit. Delete reachable from both the accordion menu and the drawer's delete button.
- `applyAndSubmit` resets the form from `buildUpdatePlanFormDefaults(plan)` before mutating + submitting — prevents stale cross-section state across the three independent hook instances editing one plan.
- Edit pre-fill `deserializeAmount`s plan amounts (raw cents) to major units for the drawer's `AmountInputField`. Guard tests assert (e.g. `'50'` from `'5000'`).
- Min-commit paywall reuses `<PremiumFeature>`; Progressive billing mirrors the create-plan form's `premiumIntegrations.includes(ProgressiveBilling)` gate. Every action gated on `hasPermissions(['plans{Create,Update,Delete}']) && !isInSubscriptionForm`.

## Serializer

`serializeMinimumCommitment` / `serializeUsageThresholds` / `serializeEntitlements` extracted from `serializePlanInput` with byte-identical output, reused by the hook. The full plan-form serializer now delegates to them.


src/components/plans/details-v2/accordions/EntitlementAccordion.tsx
```diff
@@ -0,0 +1,130 @@
...
+        <SectionAccordion
+          key={`entitlement-${entitlement.code}`}
+          title={entitlement.name || entitlement.code}
+          subtitle={entitlement.code}
+          actions={[
+            {
+              label: translate('text_63e51ef4985f0ebd75c212fc'),
+              onClick: () =>
+                drawerRef.current?.openDrawer({
+                  featureId: '',
```
AllanMichay, line 98: Is it normal for it to be empty?

AllanMichay (approved):

src/components/plans/details-v2/accordions/EntitlementAccordion.tsx
ansmonjol: Mixed concerns. The zod schema requires it initially but the BE iddentify features by code so it's an non-necessary oversight and I've added a cleanup task to get rid of this. 
Didn't did it now as it has broader impact than just removing it's invocation here.

ansmonjol: Thanks for flagging!

PR #3575, opened 2026-05-29
feat: Payment gated subscriptions

stephenlago99:
## Context

Today a subscription becomes active as soon as it's created, regardless of whether the customer has actually paid. For flows where access should be granted only after payment succeeds, we need to gate subscription activation on a successful payment, with an optional timeout after which an unpaid subscription is cancelled.

This PR implements the frontend for payment-gated subscription activation (project: Payment gated subscription activation), including the new Incomplete subscription state that a payment-gated subscription sits in until payment clears.

### Subscription form — activation rule

- New "Subscription activation rule" section in the create/edit subscription form (SubscriptionActivationRuleSection) letting the user choose:
- Activate immediately (default, unchanged behavior), or
- Activate on successful payment, with an optional timeout (hours).
- Payment-gated activation is only offered when the effective payment method can be charged automatically. When it resolves to manual (explicitly selected, or inherited from a customer with no chargeable method), the whole radio group is disabled and the selection falls back to Activate immediately.
- The section only renders for customers on the multi-payment flow / with a payment provider, and is read-only outside creation/pending-edition.

### Timeout semantics (aligned with backend)

- An empty timeout means "no timeout": the timeoutHours field is omitted from the payload (the BE expects the field absent, not null). 0 is also treated as "no timeout".
- Validation allows an empty timeout and only rejects negative / non-integer values when one is provided.
- Round-trips correctly: a rule returned with timeoutHours: null hydrates back to an empty field.

### Incomplete subscription state

- Status badge + mapping (incomplete → warning), subscriptions filter option, and list row actions updated.
- Incomplete subscriptions are not editable (useSubscriptionPermissionsActions), and opening one in the form redirects to the subscription overview.

### Subscription detail view

- Payment-gated alerts: pending/incomplete (awaiting payment), payment failed, and timeout/expired.
- New info rows for the activation rule and a human-readable timeout value (e.g. "In 12 hours", "Expired", "No timeout defined").

### Plumbing

  - serializeActivationRules / deserializeActivationRules, subscriptionUtils helpers, form-schema validation, and default-value hydration (buildSubscriptionDefaultValues). GraphQL fragments/mutations now fetch activationRules + cancelationReason.

### Tests

- Unit tests for the activation section, serializer, schema, utils, default-value hydration, status mapping, alerts, and permissions (new code ≥95% branch / 100% line coverage on the key files). No new e2e — the payment-gated path can't be reliably exercised without a provider-backed customer in the e2e env; the existing add-subscription spec still covers the create/cancel happy path.

  Fixes LAGO-1512
  Fixes LAGO-1513


ansmonjol (commented):
Can you add a description?

src/core/constants/statusSubscriptionMapping.ts
```diff
@@ -13,6 +13,11 @@ export const subscriptionStatusMapping = (status?: StatusTypeEnum | null): Statu
         type: StatusType.default,
         label: 'pending',
       }
+    case StatusTypeEnum.Incomplete:
```
AllanMichay, line 16: Since it's warning here, why not put it in the type `StatusLabelWarning` in `Status.tsx`?

src/core/utils/subscriptionUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+    return translate('text_1779882021466x423uayjorq')
+  }
+
+  if (paymentActivationRule.timeoutHours === 0) {
+    return translate('text_17798820214660s59bjuztra')
+  }
+
+  if (!paymentActivationRule.expiresAt) {
+    return translate('text_17798820214660s59bjuztra')
+  }
```
AllanMichay, line 60: Still works because we have the same text but IIRC we can have timeoutHours undefined / null. This means it would fall under expiresAt and not `timeoutHours === 0` right?

src/pages/subscriptions/CreateSubscription.tsx
AllanMichay, line 110: You have a sonarqube comment here. Can you let claude check it out?

src/core/utils/subscriptionUtils.ts
stephenlago99: Yes, this was intended (it was weird to have an expired subscription shown with a No timeout defined error)

AllanMichay (approved):

PR #3588, opened 2026-06-01
feat(plans): charge codes, sub settings edit & v2 UX fixes

ansmonjol:
Fixes LAGO-1499

UX fixes on the new plan/subscription v2 details/edition page, found while building it. Commits are split per concern.

- **Single-open action poppers** — opening one accordion `…` menu now closes any other open menu in the same group. Implements the previously-unused `popperGroupName` coordination in the design-system `Popper` (module-level registry); wired into `SectionAccordion` + the subscription entitlements tab (`Table` already passed a group name).
- **Edit plan settings in subscription context** — the plan-settings accordion now shows **Edit** in the subscription plan tab. Reuses `PlanSettingsDrawer` with non-overridable fields disabled (existing `isInSubscriptionForm` + `subscriptionFormType`) and routes the save through `updateSubscription(planOverrides)` (description + tax codes), never `updatePlan` (R3).
- **Editable, unique charge `code` (fixed + usage)** — required **Code** field in both charge drawers, shown only in the v2 details/edition UI (temporary `showCode` gate, cleanup tracked in LAGO-1498). On create it is prefilled from the add-on / billable-metric code with a numeric suffix for local uniqueness; the backend enforces final uniqueness and a duplicate is surfaced under the input (same pattern as plan-settings code). Read-only in subscription context, where the code is the override lookup key.
- **Advanced-settings separators** — minimum commitment / progressive billing / entitlement now render as visually separate sections with the same separator + spacing as the other top-level sections.


src/components/plans/drawers/planSettings/PlanSettingsDrawer.tsx
```diff
@@ -32,12 +35,34 @@ export const PlanSettingsDrawer = forwardRef<PlanSettingsDrawerRef, PlanSettings
...
+    // mutate the shared base plan (R3). Plan mode keeps the cascade submit.
+    const handleSubmit = async () => {
+      if (subscriptionId) {
+        const values = form.state.values
+        const success = await updatePlanOverride({
+          description: values.description || undefined,
+          taxCodes: values.taxes?.map((tax) => tax.code) ?? [],
+        })
+
+        if (success) drawer.close()
```
AllanMichay, line 54: updatePlanOverride always returns true (unless it breaks but we don't try catch)

src/components/plans/drawers/fixedCharge/FixedChargeDrawerContent.tsx
```diff
@@ -254,7 +258,24 @@ export const FixedChargeDrawerContent = withForm({
                 />
 
                 {showCode && (
-                  <form.AppField name="code">
+                  <form.AppField
+                    name="code"
+                    listeners={{
+                      // Clear the server "code already exists" error once the user
+                      // edits the code so the submit button re-enables. Gated by the
+                      // message so the zod required-check isn't wiped.
```
AllanMichay: Strange, it should be checked again if you use revalidateLogic but okay

src/components/plans/drawers/planSettings/PlanSettingsDrawer.tsx
ansmonjol: Good catch. You're right — `updatePlanOverride` hardcoded `return true`, and since the client runs with `errorPolicy: 'all'`, a backend error resolves into `result.errors` instead of throwing, so `success` was always truthy and the drawer closed even on a failed override (discarding the user's edits).

Fixed in 04748d08: the hook now returns `!errors?.length`, matching the pattern used by the charge cascade hooks. That makes the `if (success) drawer.close()` guard meaningful — the drawer stays open on error. Added tests for both the success and error branches.

I left the `drawer.close()` line itself unchanged since it was never the problem — the bug was in the hook it depends on, so that's where the fix belongs.

AllanMichay (approved):

PR #3604, opened 2026-06-03
feat(wallet): Link wallet breadcrumb back to the customer

domenicofalco:
## Context

On the wallet detail page there was no way to navigate back to the
owning customer. The second-level breadcrumb showed a static "Wallets"
label pointing at the customer's Wallets tab, but never surfaced who
the customer was.

## Description

- Replace the static "Wallets" breadcrumb label with the customer name,
  resolved as name → "firstname lastname" → externalId (mandatory final
  fallback, so the old translation fallback is removed as dead code).
- The label loads asynchronously with the wallet query, so the
  breadcrumb item shows a skeleton loader while it resolves.
- Add an optional `loading` flag to BreadcrumbItem and skeleton support
  to the Breadcrumb component.
- Cover both the name-resolution fallback chain and the loading state
  with tests.

Fixes ING-268


src/pages/wallet/WalletDetails.tsx
```diff
@@ -139,6 +146,11 @@ const WalletDetails = () => {
 
   const wallet = data?.wallet
 
+  const customerName =
```
stephenlago99: I think we should extract this into an util

domenicofalco: done

src/pages/wallet/WalletDetails.tsx
```diff
@@ -294,11 +307,12 @@ const WalletDetails = () => {
             path: CUSTOMERS_LIST_ROUTE,
           },
           {
-            label: translate('text_62d175066d2dbf1d50bc937c'),
+            label: customerName ?? '',
```
AllanMichay: Is it necessary since we have the fallback defined in `getCustomerDisplayName`?

AllanMichay (approved):

src/pages/wallet/WalletDetails.tsx
domenicofalco: going to remove it. thanks

PR #3619, opened 2026-06-05
feat(subs): plan/sub v2 details polish + drawer fixes

ansmonjol:
Fixes LAGO-1515 BIL-163 BIL-161 BIL-162 BIL-166

Final polish pass on the plan/subscription v2 details/edition page, tightening details found while reviewing and fixing drawer save UX, ahead of the QA pass and flag rollout.

## Details page / tabs
- Promote the v2 **Overview** and **Subscription plan** tabs as the primary tabs (behind `EDIT_DETAILS_PAGE`), replacing the legacy overview when the flag is on.
- Sub Overview now shows **name / status / external ID**.

## Subscription plan tab
- Move the "plan has no overrides, editing charges creates an overridden plan" alert out of the Subscription information section to the **top of the plan tab's right column** (full-width, flush with the sidebar separator) so it scopes the whole plan section.
- Left sidebar: **Fixed charges** and **Usage-based charges** folders now list their charges (label = `invoiceDisplayName || addOn.name || code` / `|| billableMetric.name`); clicking a charge scrolls to its accordion. The sidebar defines its own GraphQL fragments, spread by the parent query (no extra fetch).
- Subscription fee accordion shows the **formatted fee amount** in its summary.

## Bug fixes
- **Clear subscription name**: emptying the name now sends `null` (not `undefined`) so the BE actually clears it.
- **Drawer save UX** (plan settings, subscription fee, fixed charge, usage charge, minimum commitment): save button shows a spinner while submitting and is no longer disabled during submission (`disabled={!canSubmit && !isSubmitting}` + `loading={isSubmitting}`); the Button's internal loading guard blocks double-submit. Plan settings sub-mode routes through the form so `isSubmitting` is accurate, and gains Enter-to-submit.
- **Subscription fee drawer prefill**: deserialize `amountCents` before opening (was passing raw serialized cents).
- **Keep drawer open on save error**: sub-mode charge/override handlers now report success via data-presence (`!!data?.<root>?.id`), unified across the three handlers. Previously they returned `true` unconditionally, so a failed mutation (e.g. a 500) still closed the drawer.

## Misc
- Minimum commitment: show the add action without the create-permission gate.
- i18n copy + new keys; regenerated GraphQL types.


src/components/plans/drawers/usageCharge/UsageChargeDrawer.tsx
```diff
@@ -392,12 +392,15 @@ export const UsageChargeDrawer = forwardRef<UsageChargeDrawerRef, UsageChargeDra
...
+              <form.Subscribe
+                selector={({ canSubmit, isSubmitting }) => ({ canSubmit, isSubmitting })}
+              >
+                {({ canSubmit, isSubmitting }) => (
                   <Button
                     data-test="usage-charge-drawer-save"
                     onClick={handleFormSubmit}
-                    disabled={!canSubmit}
+                    disabled={!canSubmit && !isSubmitting}
+                    loading={isSubmitting}
```
AllanMichay: Just looking at this, why not use our SubmitButton? you should find a lot of example with `form.SubmitButton`

Because with the loading now, it should behave exactly like our SubmitButton

AllanMichay (commented):
The SubmitButton should be usable everywhere now that you've added all those behaviour. Unless I'm missing something (I don't remember if we already had this conversation before or not)

src/components/plans/drawers/usageCharge/UsageChargeDrawer.tsx
ansmonjol: indeeeeed my bad, good catch
Updating them all

src/pages/subscriptions/SubscriptionDetails.tsx
```diff
@@ -146,20 +146,76 @@ const SubscriptionDetails = () => {
...
+              getPlanSubscriptionDetailsRoute(
+                CustomerSubscriptionDetailsTabsOptionsEnum.editOverview,
+              ),
+            ],
+            content: (
+              <DetailsPage.Container>
+                <SubscriptionDetailsV2Overview subscriptionId={subscriptionId as string} />
+              </DetailsPage.Container>
+            ),
+            hidden: !isFeatureFlagActive(FeatureFlags.EDIT_DETAILS_PAGE),
```
domenicofalco, line 172: I think this is redundant because there's already  `isFeatureFlagActive(FeatureFlags.EDIT_DETAILS_PAGE)` as ternary expression

domenicofalco (approved):

src/pages/subscriptions/SubscriptionDetails.tsx
ansmonjol: Not idea indeed, this will be all crushed when feature releases so i guess I won't make it perfect now WDYT?

domenicofalco: sure, your call

src/components/plans/drawers/usageCharge/UsageChargeDrawer.tsx
ansmonjol: Updated them all

AllanMichay (approved):

PR #3638, opened 2026-06-10
feat(subs): display per-subscription fixed-charge units override

ancorcruz:
Subscription detail page now reads `units` via `Subscription.fixedCharges` (override-aware) instead of `Plan.fixedCharges`. A dedicated no-cache query isolates the override from the plan-scope cache entry, and the `updateSubscriptionFixedCharge` mutation result is narrowed to `id` to avoid polluting the same normalized FixedCharge.


ancorcruz:
`Subscription.fixedCharges[*]` and `Plan.fixedCharges[*]` both return FixedCharge, so Apollo's id-normalized cache shares a single entry per FixedCharge id. Naively fetching the subscription's override units would overwrite the plan-default units field for every plan-scope reader of the same charge.

stephenlago99:
I will close this - I will open a separate PR

PR #3639, opened 2026-06-10
feat(plans): refine v2 details sidebar (folders, jump-to, focus)

ansmonjol:
Refines the Plan/Subscription v2 details left sidebar.

- **BIL-159** — clicking a folder row expands/collapses it instead of navigating; only leaf items navigate.
- **BIL-160** — clicking an item opens its accordion first, then smooth-scrolls to it and shows the focus ring.
- **BIL-164** — flattened hierarchy: no "Advanced settings" folder; Minimum commitment & Progressive billing sit at the root; Entitlements is a folder with an add action listing its entitlements (`name || code`).

Also included:
- Full-width sidebar rows so the hover background is consistent across folders/leaves, with a continuous tree guide line on indented items.
- Whole-card focus ring on accordions (open or closed), reusing the existing `focus-visible` ring.
- Jump-to scroll opens the target instantly (no expand animation) and waits for the layout to settle, so it lands fully in view.

Fixes BIL-159 BIL-160 BIL-164


domenicofalco (approved):

src/components/plans/details-v2/PlanDetailsV2LeftSidebar.tsx
```diff
@@ -45,6 +51,19 @@ type SidebarItem = {
...
+// guide line is an absolute overlay (it must sit ON TOP of the hover background, like
+// GitHub's PR file tree), never a container that insets the row.
+const SIDEBAR_INDENT_STEP = 20
+const SIDEBAR_LEAF_BASE_PADDING = 24
+
+const getLeafPaddingLeft = (depth: number): number =>
+  SIDEBAR_LEAF_BASE_PADDING + depth * SIDEBAR_INDENT_STEP
+
+// x of the vertical guide line for a children block whose items sit at `childDepth`.
+const getGuideLineLeft = (childDepth: number): number => 10 + (childDepth - 1) * SIDEBAR_INDENT_STEP
```
AllanMichay: Let's update 10 to a named const just like the other const

AllanMichay (commented):
Hmmm should we remove all the snapshots testing for complex elements? (I would keep it for Accordion for example)

AllanMichay (approved):

src/components/plans/details-v2/PlanDetailsV2LeftSidebar.tsx
ansmonjol: Ah yep missed it thanks!

PR #3653, opened 2026-06-12
feat(quotes): create pricing drawer for subscriptions

AllanMichay:
## Context

We've created a `/pricing` command in RTE for the one-off type quotes. We must now link this command to it's own drawer for the subscription amendment / creation quote type

## Description

This PR
- Renames the hook for the one off pricing drawer
- Creates the hook for the subscription pricing drawer
- Reuses most of the sections from plan creation
- Refactor usePlanForm to be able to reuse the sections

Fixes LAGO-1491


src/components/designSystem/RichTextEditor/PricingBlock/useQuotePlanSettingsDrawer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const openDrawer = useCallback(() => {
+    drawer.open({
+      title: translate('text_642d5eb2783a2ad10d67031a'),
+      children: <PlanSettingsSection form={planForm} />,
+      actions: (
+        <div className="flex items-center justify-end gap-3">
+          <Button variant="quaternary" onClick={() => drawer.close()}>
+            {translate('text_6411e6b530cb47007488b027')}
+          </Button>
+          <Button data-test="plan-settings-drawer-save" onClick={() => drawer.close()}>
```
ansmonjol: is it fine that the save just closes?

src/components/designSystem/RichTextEditor/PricingBlock/SubscriptionPricingContent.tsx
```diff
@@ -0,0 +1,424 @@
...
+  subscriptionId,
+}: SubscriptionPricingContentProps) {
+  const { translate } = useInternationalization()
+
+  // Plan selection
+  const [selectedPlanId, setSelectedPlanId] = useState(initialState?.planId || '')
+
+  const { data: plansData, loading: plansLoading } = usePlansQuery({
+    variables: { limit: 100 },
+  })
```
ansmonjol: It's not a lazy query so how next plans are fetched? If user have more than 100 plans they won't find them all correct?

src/components/designSystem/RichTextEditor/PricingBlock/SubscriptionPricingContent.tsx
```diff
@@ -0,0 +1,424 @@
...
+
+  // Sync to stateRef — compute overrides from form state vs original plan
+  useEffect(() => {
+    if (!formReady || !selectedPlanId) {
+      stateRef.current = null
+      return
+    }
+
+    // Always serialize the full form state as overrides — the backend applies them on top of the plan
+    const overrides: PlanOverrides = {}
```
ansmonjol: Dual serialization path / drift risk: this block hand-builds the snake_case `overrides` from form state, while `toPlanBillingItems` separately serializes `formValues` into `payload`. Two hand-maintained form→snake_case mappers for the same form will drift over time (a new charge field added in one path but not the other silently desyncs override vs payload).

Consider deriving `overrides` inside the serializer (pass form state to `toPlanBillingItems` and compute both there), so there is a single source of truth for the field mapping.

src/core/serializers/serializeQuotePlanBillingItems.ts
```diff
@@ -0,0 +1,521 @@
...
+      nonRecurringUsageThresholds: (payload.non_recurring_usage_thresholds ?? []).map((t) => ({
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        amountCents: t.amount_cents as any,
+        thresholdDisplayName: t.threshold_display_name ?? undefined,
+        recurring: t.recurring,
+      })) as PlanFormInput['nonRecurringUsageThresholds'],
+      recurringUsageThreshold: payload.recurring_usage_threshold
+        ? ({
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            amountCents: payload.recurring_usage_threshold.amount_cents as any,
```
ansmonjol: `as any` (×4 here) goes against the no-`any` convention. These reconstruct a loosely-typed payload into `PlanFormInput`, so the intent is understandable, but the casts can be tightened to the real field types instead of `any`, maybe using `PlanFormInput['minimumCommitment']` or similar

src/components/designSystem/RichTextEditor/PricingBlock/useQuotePlanSettingsDrawer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const { translate } = useInternationalization()
+  const drawer = useDrawer()
+
+  const openDrawer = useCallback(() => {
+    drawer.open({
+      title: translate('text_642d5eb2783a2ad10d67031a'),
+      children: <PlanSettingsSection form={planForm} />,
+      actions: (
+        <div className="flex items-center justify-end gap-3">
+          <Button variant="quaternary" onClick={() => drawer.close()}>
```
domenicofalco: cancel doesn't actually cancel the form here, or?

AllanMichay: Changing how the whole section work. This way save will really save (just like the other drawers) and cancel just closes (since it was never saved to the main form)

Wanted to use section here but not sure it is possible

src/components/designSystem/RichTextEditor/PricingBlock/useQuotePlanSettingsDrawer.tsx
AllanMichay: Changing how the whole section work. This way save will really save (just like the other drawers) and cancel just closes.

Wanted to use section here but not sure it is possible

domenicofalco (approved):

ansmonjol (approved):

PR #3668, opened 2026-06-15
perf: Improve translation performance by removing Object.keys check

stephenlago99:
## Description

Performance win: translate() was ~380× slower than it needed to be

Every `translate()` call was running an Object.keys() over the full ~4,300-key translations object just to check "is this loaded yet?" — on a hot path we hit hundreds of times per render.

### Root cause
translateKey() started with Object.keys(translations).length === 0. Our translations object (4,308 keys) is in V8 dictionary mode, so Object.keys materializes and sorts the entire key array on every single call — purely to test for emptiness. A normal lookup that should cost ~12ns was costing hundreds of µs.

Measured in-browser (React DevTools + a temp counter), same form session:

one small editing session (before): 239ms of CPU
one small editing session (after): 1ms of CPU

That's ~380× per call and 239ms of pure waste removed from a single editing session. Because translate fires hundreds of times per commit, on interactive paths (e.g. typing in the subscription form) clusters of these calls were landing in single keystroke renders — enough to blow the 16ms frame budget and drop frames. So this isn't just a synthetic number: it's smoother typing and snappier re-renders on translate-heavy screens.

### The fix
Switched the "not loaded yet" sentinel from {} to undefined, so the check becomes an O(1) !translations and the per-call enumeration is gone entirely. Tiny blast radius — only translateKey reads that field, type was already ... | undefined. Added characterization tests for translateKey; all green, lint/types clean.

### DIFF
--- a/src/core/apolloClient/reactiveVars/internationalizationVar.ts
+++ b/src/core/apolloClient/reactiveVars/internationalizationVar.ts
@@ -12,7 +12,9 @@ interface InternationalizationVar {
 
 const internationalizationVar = makeVar<InternationalizationVar>({
   locale: getItemFromLS(LOCALE_LS_KEY) ?? LocaleEnum.en,
-  translations: {},
+  // `undefined` (not `{}`) is the "translations not loaded yet" sentinel so `translateKey`
+  // can detect it in O(1) without enumerating the multi-thousand-key translations object.
+  translations: undefined,
 })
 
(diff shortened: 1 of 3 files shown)


domenicofalco (approved):

AllanMichay (approved):

ansmonjol (approved):
👏

PR #3673, opened 2026-06-15
refactor(form): migrate invoice document locale dialog to TanStack Form

domenicofalco:
## Context

Migrate EditBillingEntityDocumentLocaleDialog from Formik + Yup to
TanStack Form + Zod. The legacy ref-based Dialog is retained and driven
via its native formId/formSubmit form wrapper.

Reuse the shared "Value is mandatory to move forward" key
(text_624ea7c29103fd010732ab7d) so submitting with the locale cleared
shows a proper required-field message instead of Zod's raw type error.

### DIFF
--- a/src/components/settings/invoices/EditBillingEntityDocumentLocaleDialog.tsx
+++ b/src/components/settings/invoices/EditBillingEntityDocumentLocaleDialog.tsx
@@ -90,7 +90,6 @@ export const EditBillingEntityDocumentLocaleDialog = forwardRef<
   })
 
   const isDirty = useStore(form.store, (state) => state.isDirty)
-  const canSubmit = useStore(form.store, (state) => state.canSubmit)
 
   const handleFormSubmit = (e: React.FormEvent) => {
     e.preventDefault()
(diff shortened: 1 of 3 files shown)


src/components/settings/invoices/EditBillingEntityDocumentLocaleDialog.tsx
```diff
@@ -53,73 +66,76 @@ export const EditBillingEntityDocumentLocaleDialog = forwardRef<
...
+    e.preventDefault()
+    form.handleSubmit()
+  }
+
+  const actions = ({ closeDialog }: { closeDialog: () => void }) => (
+    <>
+      <Button variant="quaternary" onClick={closeDialog}>
+        {translate('text_63e51ef4985f0ebd75c21313')}
+      </Button>
+      <Button variant="primary" type="submit" disabled={!canSubmit || !isDirty}>
```
ansmonjol: We have a submit button component for tanstack forms.
I feel claude is missing it every time, maybe worth checking in the skill where to mention this

ansmonjol (approved):

src/components/settings/invoices/EditBillingEntityDocumentLocaleDialog.tsx
domenicofalco: good point. updated both the component and the skill

PR #3685, opened 2026-06-16
refactor(org): post release | make URL slug the single source of truth for current organization

domenicofalco:
## Context

Today the current organization is bootstrapped from localStorage on every page load, and that value is what drives the org used across the app. We want to flip this: the URL slug should be the source of truth for the current org (per tab), and localStorage should be demoted to a single job — remembering the last organization the user was in.

## Description

- New RootRedirect (at `/`) resolves the landing slug from saved `from`, SSO redirect, persisted last-used slug, or first accessible membership without firing any org-scoped query.
- currentOrganizationVar is in-memory only; localStorage now stores a `lastUsedOrganizationSlug` hint consumed exclusively by RootRedirect.
- Org-scoped queries (UserIdentifier, useOrganizationInfos) skip until the org var is set, so no request is sent without an org header.
- onLogIn no longer selects/sets a current org — RootRedirect owns the post-login landing decision.
- OrganizationSwitcher only navigates; OrganizationLayout detects the slug change and runs switchCurrentOrganization once.

Fixes LAGO-1551


src/core/router/utils/orgSlug.ts
```diff
@@ -1,23 +1,38 @@
...
+ *      while navigating inside an org. Correct mid-session and multi-tab safe,
+ *      but `null` after a hard reload (the var is not persisted).
+ *   2. The persisted "last used" slug (`getPersistedOrganizationSlug()`) — the
+ *      only signal that survives a hard reload, so a cold load of a slug-less
+ *      path recovers to the org the user was last on instead of `memberships[0]`.
+ *   3. Fallback to the first membership's org slug.
  *
  * Returns `undefined` only if the user has no memberships.
  */
 export const resolveOrgSlug = (currentUser: CurrentUser): string | undefined => {
```
ansmonjol, line 38: Kinda out of scope of your changes but i feel this file should define it's own fragment to make sure the data shape matches.
Worth adding it now?

ansmonjol (approved):

src/core/apolloClient/reactiveVars/currentOrganizationVar.ts
```diff
@@ -1,21 +1,51 @@
...
+  return localStorage.getItem(LAST_USED_ORGANIZATION_LS_KEY) || null
+}
+
+export const setPersistedOrganizationSlug = (slug: string | null): void => {
+  if (typeof window === 'undefined') return
 
-  if (id) {
-    setItemFromLS(ORGANIZATION_LS_KEY_ID, id)
+  if (slug) {
+    localStorage.setItem(LAST_USED_ORGANIZATION_LS_KEY, slug)
```
AllanMichay: Why not use `setItemFromLS` ?

AllanMichay (approved):

src/core/router/utils/orgSlug.ts
domenicofalco: fair point. done!

src/core/apolloClient/reactiveVars/currentOrganizationVar.ts
domenicofalco: good point Allan! as this is most likely due to a circular dependency that was already in there and it touches quite few files, I am going to restore as it was before and open a dedicated PR for that soon.

domenicofalco: @AllanMichay this has been fixed separately here https://github.com/getlago/lago-front/pull/3693

PR #3697, opened 2026-06-17
feat(copy): inline click-to-copy in list cells

ansmonjol:
Adds inline click-to-copy (`TypographyWithCopy`) on identifier values across list cells and the customer details header, so users can copy a code / number / external id without leaving the list. Clicking the copy button stops propagation, so it never triggers row navigation.

## Places touched

Main lists:
- Customers list > email
- Billable metrics list > code
- Plans list > code
- Features list > code
- Add-ons list > code (rendered as `code • amount`)
- Coupons list > code (rendered as `code • caption`)
- Subscriptions list > external id (under name, taller rows)
- Invoices list > invoice number
- Payments list > invoice number
- Credit notes list > credit note number

Customer details:
- Customer details > header > external id (below name, non-compact)
- Customer details > subscriptions > external id (under name, taller rows)
- Customer details > applied coupons > coupon code
- Customer details > wallets > wallet code (rendered as `code • rate`)
- Customer details > invoices > invoice number
- Customer details > payments > invoice number

Shared / infra:
- `TypographyWithCopy`: new `compact` prop (default off, keeps existing look e.g. API keys)
- `MainHeader` `metadata` now accepts `string | ReactNode`; snapshot now strips React elements
- Added `code` to `AddOnItem`, `CouponItem`, `WalletAccordion` fragments (+ codegen)

## Drawbacks / trade-offs

- ~4px left gap before the button text in tables: Table wraps cells in `<Typography noWrap>` (`overflow:hidden`), so the negative margin had to be dropped (`!ml-0` + `px-1`), otherwise the rounded hover corner was cropped.
- `compact` relies on several `!important` utilities to beat the MUI theme (fixed 32px height, padding, 80px min-width) - brittle if the button theme changes.
- Magic `!leading-[26px]` on the in-button text to fix baseline alignment in the `code • text` line (caption is 24px, +2px). Hardcoded in the shared component; also applies to body-variant standalone copies (harmless, single line, no sibling).
- `!items-baseline` on the compact button to expose the text baseline; the duplicate icon then baseline-aligns and sits slightly below center.
- In the flex `code • text` line the button needs `shrink-0` and the sibling `min-w-0`, else `noWrap` collapses the code to icon-only.
- `WalletAccordion` fragment is duplicated in `CustomerWalletList.tsx` and `WalletAccordion.tsx` (pre-existing); `code` had to be added to both or codegen fails on the name collision.
- Wallet `code` is nullable - the `code •` prefix only shows when a code exists.
- A `ReactNode` passed as header `metadata` is stripped from the change-detection snapshot, so it relies on `viewName` to detect changes. Edge case: two customers with no display name and different external ids, navigated without any other config change, could show a stale copy value.
- Payments: copy only on a single-invoice payment; a multi-invoice payment request still shows the plain "N invoices" count.
- Subscriptions / coupons / wallets list rows bumped to a taller row size (72px).

Fixes BIL-171

### DIFF
--- a/src/components/MainHeader/EntitySection.tsx
+++ b/src/components/MainHeader/EntitySection.tsx
@@ -61,9 +61,13 @@ export const EntitySection: FC<{ entity?: MainHeaderEntityConfig }> = ({ entity 
           )}
         </div>
         {entity.metadataLoading && <Skeleton variant="text" className="w-24" />}
-        {!entity.metadataLoading && entity.metadata && (
-          <Typography data-test={ENTITY_SECTION_METADATA_TEST_ID}>{entity.metadata}</Typography>
-        )}
+        {!entity.metadataLoading &&
+          entity.metadata &&
+          (typeof entity.metadata === 'string' ? (
+            <Typography data-test={ENTITY_SECTION_METADATA_TEST_ID}>{entity.metadata}</Typography>
+          ) : (
+            <div data-test={ENTITY_SECTION_METADATA_TEST_ID}>{entity.metadata}</div>
+          ))}
       </div>
     </div>
   )
--- a/src/components/MainHeader/MainHeaderConfigure.tsx
+++ b/src/components/MainHeader/MainHeaderConfigure.tsx
@@ -18,6 +18,12 @@ function configSnapshot(config: MainHeaderConfig): string {
     if (key === 'icon' && typeof value !== 'string') return undefined
     if (key === 'snapshotKey') return value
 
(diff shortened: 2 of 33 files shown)


AllanMichay (approved):
Too bad for the importants but seems like we have no choice

> Subscriptions / coupons / wallets list rows bumped to a taller row size (72px).

I only see it twice: SubscriptionPage and CustomerSubscription. Did you forget some?

PR #3716, opened 2026-06-22
refactor(type filters): return query-specific types from formatFiltersForQuery

domenicofalco:
## Context

`formatFiltersForQuery` returned a loose `Record`, agnostic of its target query, so it never warned when a filter (or `keyMap` remap) produced a key the query doesn't accept. This hid a bug: Forecasts emitted `accountType`, absent from its query, silently dropping the customer-account-type filter.

## Description

- Make `formatFiltersForQuery<T>` generic (default keeps old callers/tests working) and constrain `keyMap` targets to `keyof T` → invalid keys are now compile errors.
- Type each wrapper's return as `Pick<…QueryVariables>` of its query.
- Fix the surfaced bug: remap `customerAccountType → customerType` in Forecasts (only runtime change).
- Consumer touch-ups: generic `formatAmountCurrency` + export cast, align `getDefaultStaticTimeGranularityFilter` to `TimeGranularityEnum`, drop the redundant security-logs guard.

No behavioral change except the Forecasts fix; types, tests and lint pass.

Fixes LAGO-1559


stephenlago99 (approved):
Nice 👍

PR #3723, opened 2026-06-22
chore(sonar): suppress S6679 isNaN rule project-wide

ansmonjol:
Adds `javascript:S6679` and `typescript:S6679` to the SonarCloud ignore list so the \"Number.isNaN() should be used\" warning no longer fires across the codebase.


sonar-project.properties
AllanMichay, line 1: Should we do the same for Number.parseInt and number.parseFloat?

ansmonjol: Does it complain about them? Let's merge this and open when it does wdyt?

AllanMichay (approved):

sonar-project.properties
AllanMichay: it did. We will merge this one and add another yes

PR #3727, opened 2026-06-23
feat(quotes): add download and approve button on edit

AllanMichay:
## Context

While testing, we decided to add other action buttons on the edit quote flow

## Description

This PR adds download pdf and approve quote button on the edit quote flow

Fixes LAGO-1574

### DIFF
--- a/src/pages/quotes/editQuote/EditQuoteAside.tsx
+++ b/src/pages/quotes/editQuote/EditQuoteAside.tsx
@@ -194,6 +194,12 @@ const EditQuoteAsideForm = ({
 
   const gridClassName = 'grid grid-cols-[7.5rem_1fr] items-center gap-0 gap-y-2'
 
+  const handleDownloadPdf = () => {
+    download(buildQuotePreviewProps(quote.currentVersion, quote.customer, pdfHeader)).catch(
+      () => undefined,
+    )
+  }
+
   return (
     <div className="flex min-h-full flex-col">
       <div className="flex flex-col gap-3 px-3 py-4">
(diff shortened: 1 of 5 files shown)


src/pages/quotes/editQuote/EditQuoteAside.tsx
```diff
@@ -304,7 +331,33 @@ const EditQuoteAsideForm = ({
...
       </div>
-    </>
+      <div className="sticky bottom-0 mt-auto flex justify-end gap-3 border-t border-grey-200 bg-white p-4">
+        <Button
+          variant="secondary"
+          data-test={EDIT_QUOTE_ASIDE_DOWNLOAD_PDF_TEST_ID}
+          loading={isSaving}
+          disabled={!!isSaving}
+          onClick={() =>
+            download(buildQuotePreviewProps(quote.currentVersion, quote.customer, pdfHeader)).catch(
```
domenicofalco: perhaps this can be brought outside the render just for readability

src/pages/quotes/editQuote/EditQuoteAside.tsx
```diff
@@ -46,6 +58,7 @@ const EditQuoteAside = ({
   return (
     <EditQuoteAsideForm
       quote={quote}
+      isSaving={isSaving}
```
domenicofalco, line 61: this is prop drilling all the way down. And I am wondering what is `<EditQuoteAsideForm` for and why is it a separate component since it not exported either? perhaps having just one component, it removes the overhead of drilling so many props

domenicofalco (approved):

src/pages/quotes/editQuote/EditQuoteAside.tsx
AllanMichay: I feel like merging both of these would create a way too complex to read component. Not against discussing it a bit more if you want. I think having it separated makes it clearer for the next time we will have to work on this.

domenicofalco: fair

PR #3728, opened 2026-06-23
refactor(payments): remove multiple_payment_methods feature flag

domenicofalco:
## Context

The feature is now enabled for all customers, so the flag is no longer needed. Remove every `hasFeatureFlag(FeatureFlagEnum.MultiplePaymentMethods) ` check and enable the previously gated payment-method behavior unconditionally, dropping the dead legacy retry-collect path in the invoice lists and the flag-off branches and tests.

Fixes ING-342


stephenlago99 (approved):

domenicofalco:
This will be merged once the feature is fully open for all customers

AllanMichay (approved):

PR #3734, opened 2026-06-24
refactor(EditInvoiceDisplayName): Migrate to TanStack Form + new dialog system

domenicofalco:
## Context

The dialog used the deprecated Formik/Yup stack and the legacy ref-based `Dialog`. Both are being phased out in favor of TanStack Form (+ Zod) and the hook-based dialog system in `~/components/dialogs`.

## Description

- Migrate the form from Formik/Yup to `useAppForm` + Zod.
- Replace the ref-based `Dialog` with a `useEditInvoiceDisplayNameDialog()` hook backed by `useFormDialog`; rename the file to `useEditInvoiceDisplayName.tsx`.
- Update `CreateInvoice` (the only real caller) to use the hook; remove dead, never-opened dialog instances from `CreatePlan` and `CreateSubscription`.
- Rewire `EditInvoiceDisplayNameButton` onto the hook (no more ref prop).
- Update tests (new hook + button tests, fix stale mocks).

Fixes LAGO-1209


src/components/invoices/useEditInvoiceDisplayName.tsx
```diff
@@ -0,0 +1,100 @@
...
+              )}
+            </form.AppField>
+          </div>
+        ),
+        closeOnError: false,
+        mainAction: (
+          <form.AppForm>
+            <form.Subscribe selector={(state) => state.isDirty}>
+              {(isDirty) => (
+                <form.SubmitButton disabled={!isDirty}>
```
AllanMichay: Should we keep the old behaviour? Most of our new form uses a simple submit button that is disabled only after the first submit.

the user can click at any moment and errors will appear. I think we should move to this behaviour

AllanMichay (approved):

PR #3739, opened 2026-06-24
refactor(EditInvoiceItemDescriptionDialog): Migrate to TanStack Form + new dialog system

domenicofalco:
## Context

The dialog used the deprecated Formik/Yup stack and the legacy ref-based `Dialog`.
Both are being phased out in favor of TanStack Form (+ Zod) and the hook-based dialog
system in `~/components/dialogs`.

## Description

- Migrate the form from Formik/Yup to `useAppForm` + Zod (`description`, optional, max 255).
- Replace the ref-based `Dialog` with a `useEditInvoiceItemDescriptionDialog()` hook backed
  by `useFormDialog`; closes via callback, char counter preserved.
- Validate on change (`revalidateLogic({ mode: 'change' })`) so the submit button disables
  live when the 255-char limit is exceeded (matching the previous Formik behavior).
- Update `CreateInvoice` (the only caller) to use the hook.
- Update tests (new hook test + fix stale mock).

Fixes [LAGO-1210](https://linear.app/getlago/issue/LAGO-1210)


src/components/invoices/EditInvoiceItemDescriptionDialog.tsx
```diff
@@ -1,108 +1,106 @@
...
-      }),
-      validateOnMount: true,
-      enableReinitialize: true,
-      onSubmit: (values, formikBag) => {
-        !!values.description && data?.callback(values.description)
+  const form = useAppForm({
+    defaultValues: {
+      description: '',
+    },
+    validationLogic: revalidateLogic({ mode: 'change' }),
```
ansmonjol, line 32: isn't it too agressive?

ansmonjol (approved):

src/components/invoices/EditInvoiceItemDescriptionDialog.tsx
domenicofalco: I made it this way because the only validation on this form is the limit at 255 chars for the textarea. therefore it disables the button as soon at the limit is exceeded

PR #3745, opened 2026-06-25
refactor(EditFeeBillingPeriod): Migrate to TanStack Form + new dialog system

domenicofalco:
## Context

The dialog used the deprecated Formik/Yup stack and the legacy ref-based `Dialog`.
Both are being phased out in favor of TanStack Form (+ Zod) and the hook-based dialog
system in `~/components/dialogs`.

## Description

- Migrate the form from Formik/Yup to `useAppForm` + Zod (`fromDatetime` required, plus the
  cross-field rule `toDatetime` must be a valid date and later than `fromDatetime`).
- Replace the ref-based `Dialog` with a `useEditFeeBillingPeriodDialog()` hook backed by
  `useFormDialog`; closes via callback.
- Validate on change (`revalidateLogic({ mode: 'change' })`) so the submit button disables
  live when the period is invalid; reset to the opened values so it stays disabled until edited.
- Keep the raw `DatePicker` inside `AppField` to preserve the start/end-of-day transforms.
- Update `CreateInvoice` (the only caller) to use the hook.
- Update tests (new hook test + fix stale mock).

Fixes [LAGO-1208](https://linear.app/getlago/issue/LAGO-1208)


src/components/invoices/EditFeeBillingPeriod.tsx
```diff
@@ -1,151 +1,148 @@
...
-  const formikProps = useFormik<Omit<EditFeeBillingPeriodProps, 'callback'>>({
-    initialValues: {
-      fromDatetime: data?.fromDatetime || '',
-      toDatetime: data?.toDatetime || '',
+  const form = useAppForm({
+    defaultValues: {
+      fromDatetime: '',
+      toDatetime: '',
+    },
+    validationLogic: revalidateLogic({ mode: 'change' }),
```
AllanMichay: Why did you change the revalidateLogic?

domenicofalco: `mode: 'change'` gives live validation so the toDatetime inline error shows onChange straightaway, not only on submit and it matches the old Formik behaviour as Formik validates onChange by default. if we compare the previous UI, it indeed validates these fields onChange as well

AllanMichay: Yeah on the behaviour side, date time was on change validation since the old ui used to have submit disabled until "everything is okay" whereas now we let the user submit first. 

The default revalidateLogic is `mode: onSubmit` then `afterSubmit: onChange`

Thus my question. On the technical side it's a bit better. I'd love to have validation from product but in the mean time, i'll validate

AllanMichay (approved):

AllanMichay (approved):

PR #3746, opened 2026-06-25
fix(quotes): use customer displayName in quote/order form displays

AllanMichay:
## Context

The quote and order form screens displayed the customer's raw `name` field. This switches them to the customer's **`displayName`**

## Description

This PR
- adds to graphQL fragments `displayName`
- use `displayName` everywhere

Fixes LAGO-1603

### DIFF
--- a/src/pages/quotes/hooks/useQuotes.ts
+++ b/src/pages/quotes/hooks/useQuotes.ts
@@ -21,6 +21,7 @@ gql`
     customer {
       id
       name
+      displayName
     }
   }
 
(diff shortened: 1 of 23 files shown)


src/pages/quotes/hooks/useQuotes.ts
```diff
@@ -21,6 +21,7 @@ gql`
     customer {
       id
       name
```
ansmonjol: Do we still need the name on all those fragments / pages you switched to display name?

ansmonjol (approved):

src/pages/quotes/hooks/useQuotes.ts
AllanMichay: Hmm you're right, I'll check this

PR #3747, opened 2026-06-25
feat(multiple billing): Warn when switching a customer's billing entity affects EU tax

domenicofalco:
## Context

EU Tax Management is set per billing entity. Since a customer's billing entity is now mutable, switching it can silently change how taxes are calculated on future invoices, with no warning to the operator.

## Description

Adds an info banner below the "Affiliated entity" picker on the customer edit form, shown in two cases (edit mode only, behind the `multi_entity_billing` flag):

1. the current entity uses EU tax and a different entity is selected;
2. the current entity does not use EU tax but the selected one does.

Logic lives in an isolated, unit-tested `BillingEntityTaxAlerts` component; `euTaxManagement` is now exposed on the billing-entity options and the customer fragment. Includes 2 translation keys and tests.

Fixes ING-350


ansmonjol (approved):
Claude is adding a ton of comments 😅

PR #3750, opened 2026-06-25
feat(orders): display orders on quotes list and quote details

AllanMichay:
## Context

Adds a read-only **Orders** list to the Quotes feature, completing the Quote → Order Form → **Order** hierarchy alongside the existing Quotes and Order Forms lists.

## Description

A single `OrdersList` component, reused in two places (mirroring how `OrderFormsList` is reused):

Fixes LAGO-1430

### DIFF
--- a/src/pages/quotes/OrdersList.tsx
+++ b/src/pages/quotes/OrdersList.tsx
@@ -1,0 +1,61 @@
+import { InfiniteScroll } from '~/components/designSystem/InfiniteScroll'
+import { Table, TableColumn } from '~/components/designSystem/Table/Table'
+import { DetailsPage } from '~/components/layouts/DetailsPage'
+import { OrderListItemFragment } from '~/generated/graphql'
+import { useInternationalization } from '~/hooks/core/useInternationalization'
+import { useOrganizationInfos } from '~/hooks/useOrganizationInfos'
+
+import { createOrdersPaginationHandler } from './common/ordersPaginationHandler'
+import {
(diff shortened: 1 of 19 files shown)


src/pages/quotes/OrdersList.tsx
```diff
@@ -0,0 +1,61 @@
...
+    quoteNumber ? { quoteNumber: [quoteNumber] } : undefined,
+  )
+
+  const columns: Array<TableColumn<OrderListItemFragment>> = [
+    ordersNumberColumn(translate),
+    ordersStatusColumn(translate),
+    ...(quoteNumber ? [] : [ordersSourceQuoteColumn(translate)]),
+    ordersOrderFormColumn(translate),
+    ordersExecutionModeColumn(translate),
+    ordersExecutionDateColumn(translate, intlFormatDateTimeOrgaTZ),
```
domenicofalco, line 36: not sure to understand why columns are separated in such modular way? also, having a hook instead of a function like here, allows you to fetch both the `useOrganizationInfos` and `translate` directly inside the hook without the need of passing them down to these functions

domenicofalco (approved):

src/pages/quotes/OrdersList.tsx
AllanMichay: I started doing this because I had to use the table display with similar column at multiple places. This one will have the same issue. But you're right, I'll move everything to a hook based design in another PR 👍

ansmonjol (approved):

PR #3770, opened 2026-06-29
refactor(subscriptions): split PaymentMethodsInvoiceSettings into standalone components

domenicofalco:
## Context

Prerequisite for the subscription drawer migration. `PaymentMethodsInvoiceSettings` bundles the payment-method and invoice-custom-section selectors, but they are independent (disjoint fields, separate guards, own dialogs). They must be standalone before moving into separate drawers (ING-361, ING-362).

## Description

- Extracted `PaymentMethodSettings` (field `paymentMethod`, guard `externalId`) and `InvoiceCustomSectionSettings` (field `invoiceCustomSection`, guard `id`), each owning its `formFieldBasePath` adapter.
- `PaymentMethodsInvoiceSettings` is now a thin composite of the two — identical props and output, so the 6 consumers are unchanged.
- Added unit tests for both (gating, prop forwarding, nested field-path).

No behavior change.

Fixes ING-360

### DIFF
--- a/src/components/paymentMethodsInvoiceSettings/types.ts
+++ b/src/components/paymentMethodsInvoiceSettings/types.ts
@@ -34,20 +34,11 @@ export interface PaymentMethodsForm<T extends ViewTypeEnum = ViewTypeEnum> {
   setFieldValue(field: string, value: unknown): unknown
 }
 
-export interface PaymentMethodsSettingsBaseProps<T extends ViewTypeEnum = ViewTypeEnum> {
+// Shared by both single-purpose settings components and the composite — they
+// only differ in which child (and customer field) they render.
+export interface SettingsComponentProps<T extends ViewTypeEnum = ViewTypeEnum> {
   customer: CustomerForPaymentMethods
   form: PaymentMethodsForm<T>
   viewType: T
   formFieldBasePath?: string
 }
-
-// The single-purpose settings components and the composite share the exact same
-// props contract — they only differ in which child (and customer field) they render.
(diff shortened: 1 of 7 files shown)


src/components/paymentMethodsInvoiceSettings/types.ts
```diff
@@ -34,9 +34,20 @@ export interface PaymentMethodsForm<T extends ViewTypeEnum = ViewTypeEnum> {
...
+// The single-purpose settings components and the composite share the exact same
+// props contract — they only differ in which child (and customer field) they render.
+export type PaymentMethodSettingsProps<T extends ViewTypeEnum = ViewTypeEnum> =
+  PaymentMethodsSettingsBaseProps<T>
+
+export type InvoiceCustomSectionSettingsProps<T extends ViewTypeEnum = ViewTypeEnum> =
+  PaymentMethodsSettingsBaseProps<T>
+
+export type PaymentMethodsInvoiceSettingsProps<T extends ViewTypeEnum = ViewTypeEnum> =
+  PaymentMethodsSettingsBaseProps<T>
```
AllanMichay: Not sure I understand why we create this

AllanMichay (approved):

src/components/paymentMethodsInvoiceSettings/types.ts
domenicofalco: good catch. i'd split them per-component for self-documenting names + future divergence but they're identical today so I collapsed into one shared type. thanks

PR #3786, opened 2026-06-30
refactor(subscriptions): payment settings drawer + atomic selector fields

domenicofalco:
## Context

Subscription create form is moving inline sections into focused drawers.
Invoicing settings already landed; this adds the Payment settings drawer and
unifies how the payment-method / invoice-custom-section selectors are built so
the drawers, the legacy dialogs and the details-v2 edit drawer share one
implementation.

## Description

- Extract atomic `PaymentMethodFields` (radios + combobox, header-less); rewrite
  `EditPaymentMethodDialog` to reuse it (−130 lines) and switch details-v2 to it.
- Add `PaymentSettingsDrawer` + `PaymentSettingsSection` entry card, wired into
  `CreateSubscription`.
- Make `InvoiceCustomSectionFields` + consolidation atomic too; align all drawers
  to the canonical title hierarchy (`PageTitle` 24/16 + `PageSectionTitle` 18/14).
- Validation: block Save + inline error when "specific selection" is empty
  (payment & custom sections), via zod schema + `revalidateLogic`.
- No change to `SelectedPaymentMethod`, the `paymentMethod` field, or the
  `PaymentMethodComboBox` export — the 7 other consumers are untouched.

Fixes Fixes [ING-361](https://linear.app/getlago/issue/ING-361)


src/components/subscriptions/form/PaymentSettingsDrawer.tsx
```diff
@@ -0,0 +1,153 @@
...
+        </CenteredPage.PageSection>
+      </CenteredPage.SectionWrapper>
+    )
+  },
+})
+
+export interface PaymentSettingsDrawerRef {
+  openDrawer: (values: { paymentMethod?: SelectedPaymentMethod }) => void
+  closeDrawer: () => void
+}
```
AllanMichay, line 85: Do we use a ref still because of the same issue Alex had? (IIRC Alex hd to make it work with pages on tanstack and others on formik)

src/components/subscriptions/form/InvoicingSettingsDrawer.tsx
```diff
@@ -90,10 +142,6 @@ interface InvoicingSettingsDrawerProps {
   onSave: (values: InvoicingSettingsValues) => void
 }
 
-// Invoicing settings drawer for the subscription form: holds a local draft of
-// `consolidateInvoice` + `invoiceCustomSection` (seeded on open from the parent
-// form) and commits it back through `onSave` on "Save edits". Mirrors the
-// FixedChargeDrawer local-draft pattern — no live binding to the parent form.
 export const InvoicingSettingsDrawer = forwardRef<
```
AllanMichay, line 145: Do we use a ref still because of the same issue Alex had? (IIRC Alex hd to make it work with pages on tanstack and others on formik)

AllanMichay (approved):

src/components/subscriptions/form/PaymentSettingsDrawer.tsx
domenicofalco: nope, this is all TanStack. The ref is just the shared drawer convention we use everywhere (FixedChargeDrawer, UsageChargeDrawer, MinimumCommitmentDrawer...). We could drop the ref by moving the draft + drawer.open() into the Section, but it'd only diverge from the other 6 drawers, so better as a separate refactor if we wanna de-ref the whole pattern. Kept it consistent for now

src/components/subscriptions/form/InvoicingSettingsDrawer.tsx
domenicofalco: same as above

PR #3803, opened 2026-07-02
feat(quotes): upload images in the rich-text editor

AllanMichay:
## Context

Replaces the Quote rich-text-editor's **paste-a-URL** image flow with **file upload**. Clicking the toolbar image button now opens a `DocumentUploader` (drag/drop or pick) that uploads to the quote as an ActiveStorage blob via the existing `AddQuoteImage` mutation, then inserts the image.

## Description

This PR
- Uses the already existing backend mutation to allow image upload.
- Backend sends back a JSON map `{ blobId: signedUrl }`
- We store it in the content as `` and resolve the signedUrl when we find this content

Fixes LAGO-1402


src/pages/quotes/common/buildQuotePreviewProps.ts
```diff
@@ -21,12 +21,14 @@ export interface QuotePreviewProps {
...
   customerCurrency?: CurrencyEnum
   mentionValues: Record<string, string>
+  images: Record<string, string>
   header?: QuotePdfHeaderData
 }
 
 export const buildQuotePreviewProps = (
   version: QuotePreviewVersionFragment | null | undefined,
   customer: QuotePreviewCustomerFragment | null | undefined,
+  images: Record<string, string> = {},
```
stephenlago99: `images` is inserted here as the 3rd positional argument, *before* the optional `header`. Every current caller was updated correctly, but since both `images` and `header` are plain objects, a future caller copying the old `(version, customer, header)` shape would silently pass `header` into `images` — losing the PDF header and breaking image resolution — with **no type error** to catch it.

Consider either appending `images` after `header`, or switching to an options object (`buildQuotePreviewProps(version, customer, { images, header })`) so argument order can't be transposed.

src/components/designSystem/RichTextEditor/common/printHtmlContent.ts
```diff
@@ -99,7 +99,46 @@ export const printHtmlContent = (html: string, options?: { title?: string }): vo
...
+    ),
+  )
+
+  // Bound the wait: a stalled resource must never hang the download. Whichever
+  // settles first — all resources loaded, or the timeout — triggers printing.
+  const RESOURCE_WAIT_TIMEOUT_MS = 3000
+  const waitForResources = Promise.race([
+    Promise.all([waitForStylesheets, waitForImages]),
+    new Promise<void>((resolve) => {
+      globalThis.setTimeout(resolve, RESOURCE_WAIT_TIMEOUT_MS)
```
stephenlago99: Minor: when `Promise.all([...])` wins this race, the 3s timer keeps running (nothing cancels it), so it fires up to 3s later. It's harmless here since `resolve` on an already-settled race is a no-op, but capturing the id and `clearTimeout`-ing it in a `.finally()` (or after the race settles) avoids a dangling timer — and keeps fake-timer tests from having to drain it.

src/pages/quotes/common/buildQuotePreviewProps.ts
AllanMichay: Good catch, will handle it

AllanMichay: Updated it so it uses an object as params so argument order is no longer an issue

domenicofalco (approved):

stephenlago99 (approved):

PR #3808, opened 2026-07-02
feat(pagination): replace infinite scroll with numbered pagination

domenicofalco:
## Context

Every list/table used infinite scroll (shared `InfiniteScroll`) plus two "Load more" buttons in the customer portal. Infinite scroll fired a query on every scroll-to-bottom, kept every fetched page mounted in the DOM, gave no sense of total size, and discouraged real search. The backend was already page/offset-ready (`page`, `limit` + `metadata { currentPage, totalPages, totalCount }`), so this is a frontend presentation change: move every list to standard numbered pagination.

## Description

**Foundations (new)**
- `Pagination` — table-footer pager: `‹ ›` prev/next + "start-end of N results" label + a rows-per-page menu (20 / 50 / 100, default 20). Stays visible during loading with its own loader (skeleton label + disabled arrows) so it doesn't flicker in/out. Renders nothing when there's a single page.
- `PaginatedContent` — the `InfiniteScroll` counterpart: renders the list then a `Pagination` at the bottom. `sticky` (default true); pass `false` for lists inside a scrolling section.
- `createSinglePageFieldPolicy` — same `keyArgs` as the paginated policy (page/limit/offset excluded) but a **replace** merge instead of append, so the rendered collection is exactly the page navigated to. Page nav stays on `fetchMore({ variables: { page } })`.
- `~/core/constants/pagination` — `DEFAULT_PAGE_SIZE`, `PAGE_SIZE_OPTIONS`.

**Migration** — every list/table swapped from `InfiniteScroll` / "Load more" to `PaginatedContent`: customers, plans, invoices, payments, credit notes, coupons, add-ons, billable metrics, features, taxes, dunning, API keys, members, invites, quotes, wallets, subscriptions, activity logs, API logs, events, webhook logs, security logs, premium `dataApi*` analytics, and the customer portal (invoices + wallets). `InfiniteScroll` deleted.

**Cache** — each migrated field flipped to a replace merge (`createSinglePageFieldPolicy`, or an in-place replace merge for the bespoke `keyArgs: false` fields). Guard test (`cache.test.ts`) stays green — every paginated field still exposes a `merge`.

**Loading** — the list shows a full-page skeleton on load (blank the rows so skeletons replace instead of appending); the pager stays put with its own loader. Fixes the dev-section bug where skeletons were appended below the rows on page change.

**Details**
- Non-sticky pager inside the single-customer detail tabs and the customer portal; full-page lists keep the sticky-bottom pager.
- Queries select `metadata.totalCount` so the footer shows "X-Y of N results".
- The rows-per-page menu renders at `z-dialog` so it isn't hidden behind the devtools console (`z-console`).
- Translations: added the range + rows-per-page keys, removed the orphaned "Load more" key.

Fixes LAGO-1663


src/components/wallets/CustomerWalletList.tsx
```diff
@@ -306,21 +309,17 @@ export const CustomerWalletsList = ({ customerId }: CustomerWalletListProps) =>
...
+        <PaginatedContent
+          metadata={data?.wallets?.metadata}
+          loading={loading}
+          onPageChange={(page) => fetchMore({ variables: { page } })}
+          sticky={false}
         >
           <Table
             name="customer-wallet-list"
-            data={walletsCollection}
+            data={loading ? [] : walletsCollection}
```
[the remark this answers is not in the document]
domenicofalco: False positive here: when switching pages the list is briefly still filled while it's loading, so this check does real work — it clears the old rows so the loading placeholders show in their place instead of stacking underneath. Keeping as is.

src/pages/settings/Dunnings/Dunnings.tsx
```diff
@@ -170,25 +175,20 @@ const Dunnings = () => {
...
+                        onPageChange={(page) => fetchMore({ variables: { page } })}
+                        onPageSizeChange={setPageSize}
                       >
                         <Table
                           name="dunnings-settings-list"
                           containerSize={{ default: 0 }}
                           rowSize={72}
                           isLoading={loading}
-                          data={sortedTable}
+                          data={loading ? [] : sortedTable}
```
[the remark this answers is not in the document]
domenicofalco: False positive here: when switching pages the list is briefly still filled while it's loading, so this check does real work — it clears the old rows so the loading placeholders show in their place instead of stacking underneath. Keeping as is.

src/components/activityLogs/ActivityLogsTable.tsx
```diff
@@ -40,6 +40,7 @@ export const ActivityLogsTable: FC<ActivityLogsTableProps> = ({
   isLoading,
   containerSize = 16,
   onRowActionLink,
+  loadingRowCount,
```
ansmonjol, line 43: What about defaulting to `DEFAULT_PAGE_SIZE`?

ansmonjol: Same for the Table component

src/components/analytics/mrr/MrrBreakdownSection.tsx
```diff
@@ -125,32 +127,32 @@ export const MrrBreakdownSection = () => {
...
+          loadingRowCount={DEFAULT_PAGE_SIZE}
           isLoading={mrrPlanBreakdownLoading}
           hasError={!!mrrPlanBreakdownError}
           data={
-            mrrPlanBreakdownData?.dataApiMrrsPlans.collection.map((p) => ({
-              id: p.planId,
-              ...p,
-            })) || []
+            mrrPlanBreakdownLoading
+              ? []
```
ansmonjol: Why do we have to compute here now? seems we're doing that a lot

ansmonjol (approved):

src/components/analytics/mrr/MrrBreakdownSection.tsx
domenicofalco: that is a fair point and it breaks the loading state if not done this way by appending the loading state at the loaded list, that is why the cleaning is needed...I am looking into changing the <Table component itself to mitigate this.

src/components/activityLogs/ActivityLogsTable.tsx
domenicofalco: checking

src/components/analytics/mrr/MrrBreakdownSection.tsx
domenicofalco: this should have been cleaned up with the latest commit. have a look. i feel it is much better. thanks!

src/components/activityLogs/ActivityLogsTable.tsx
domenicofalco: that should be addressed! ty

PR #3816, opened 2026-07-03
fix(superset): renew guest token on expiry

ancorcruz:
## Context

The analytics dashboards embed Apache Superset in an iframe. Superset only authorizes an embedded viewer through a short-lived guest token (5 minutes by default). The embed SDK re-invokes the `fetchGuestToken` callback shortly before the current token expires so the session can continue without interruption.

Our callback returned the single token captured at page load, so the SDK kept handing back an already-expired token. After a few minutes every chart and filter request failed, and the only way to recover was a full page reload, which also reset the active tab and filters. This is the "leave analytics open, change a filter, get an error" behaviour customers reported.

## Description

Extract a `fetchGuestToken` helper that mints a fresh, organization-scoped token on every call through the new `createSupersetGuestToken` mutation, and wire it into the dashboard embed. Each renewal now returns a valid token, keeping long analytics sessions alive without a reload. The mutation runs with a `no-cache` fetch policy so the token is never written into the Apollo cache, which this app persists to IndexedDB — a short-lived credential has no business being stored there.

The failure and teardown paths are where the SDK's behaviour dictates ours:

- **A failed mint must not reject.** The SDK re-arms its timer only from the resolution of our callback, so a rejection kills renewal permanently until a full page reload — the exact bug this PR fixes.
- **A failed mint must not simply return the stale token either.** The SDK derives the next refresh delay from whatever we hand back, flooring at 5s once that token is expired. Returning it unconditionally turns an outage into ~720 mutations an hour per tab, each making the backend re-authenticate against Superset. So the fetcher retries up to 3 times with a 1s/2s backoff, then applies a per-streak cooldown (5s, doubling to a 5 minute cap) before minting again, and only falls back to the last known good token in the meantime.
- **Teardown must stop the loop without stranding the embed.** `unmount()` only clears the iframe, never the pending refresh timer, so the effect cleanup calls a `cancel()` that makes subsequent calls return a never-settling promise — the only way to halt a chain that ignores rejections. A call already in flight still settles: the SDK awaits our callback inside `Promise.all([fetchGuestToken(), mountIframe()])` and puts its iframe in the DOM before that resolves, so hanging there would leave `embedDashboard` unresolved and the iframe orphaned with its Switchboard port open. `Dashboard.tsx` also unmounts an embed that lands after cleanup already ran.
- **Failures are reported.** The first failure of each streak is captured with an `errorType: 'SupersetGuestTokenMintError'` tag; the rest of the streak stays quiet so a sustained outage doesn't bury the signal. GraphQL errors are left to the errorLink, which already captures them; network errors and token-less responses (a non-null schema contract violation) are captured here.

`pnpm codegen` has been run and the regenerated types committed, so the mutation is consumed through `CreateSupersetGuestTokenDocument` with generated type parameters rather than a hand-rolled document and an `any` payload.

### DIFF
--- a/src/pages/dashboards/Dashboard.tsx
+++ b/src/pages/dashboards/Dashboard.tsx
@@ -1,4 +1,4 @@
-import { gql } from '@apollo/client'
+import { gql, useApolloClient } from '@apollo/client'
 import { embedDashboard, EmbeddedDashboard } from '@superset-ui/embedded-sdk'
 import { debounce } from 'lodash'
 import { useEffect, useMemo, useRef } from 'react'
@@ -15,6 +15,7 @@ import { useSupersetDashboardsQuery } from '~/generated/graphql'
 import { useInternationalization } from '~/hooks/core/useInternationalization'
 import { useCurrentUser } from '~/hooks/useCurrentUser'
 import '~/main.css'
+import { createFetchSupersetGuestToken } from '~/pages/dashboards/fetchSupersetGuestToken'
 import ErrorImage from '~/public/images/maneki/error.svg'
 import { PageHeader } from '~/styles'
 
@@ -42,6 +43,7 @@ export type DashboardProps = {
 const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: DashboardProps) => {
   const { translate } = useInternationalization()
   const { currentMembership } = useCurrentUser()
+  const client = useApolloClient()
 
   const dashboardRef = useRef<string>('')
 
@@ -107,7 +109,7 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
         id: dashboard.embeddedId,
         supersetDomain: lagoSupersetUrl,
         mountPoint,
-        fetchGuestToken: async () => dashboard?.guestToken,
+        fetchGuestToken: createFetchSupersetGuestToken(client, dashboard.id),
         dashboardUiConfig: {
           hideTitle: true,
           emitDataMasks: persistFilters,
@@ -133,7 +135,7 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
       embedded?.unmount()
       dashboardRef.current = ''
     }
-  }, [dashboard, currentMembership?.organization.id, dashboardTitle, mountId])
+  }, [dashboard, currentMembership?.organization.id, client, dashboardTitle, mountId])
 
   return (
     <>
--- a/src/pages/dashboards/Dashboard.tsx
+++ b/src/pages/dashboards/Dashboard.tsx
@@ -89,6 +89,12 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
         }, 500)
       : null
 
+    const fetchGuestToken = createFetchSupersetGuestToken(
+      client,
+      dashboard.id,
+      dashboard.guestToken,
+    )
+
     const mount = async () => {
       const mountPoint = document.getElementById(mountId)
 
@@ -109,7 +115,7 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
         id: dashboard.embeddedId,
         supersetDomain: lagoSupersetUrl,
         mountPoint,
-        fetchGuestToken: createFetchSupersetGuestToken(client, dashboard.id),
+        fetchGuestToken,
         dashboardUiConfig: {
           hideTitle: true,
           emitDataMasks: persistFilters,
@@ -131,6 +137,7 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
     mount()
 
     return () => {
+      fetchGuestToken.cancel()
       debouncedSaveFilters?.cancel()
       embedded?.unmount()
       dashboardRef.current = ''
--- a/src/pages/dashboards/Dashboard.tsx
+++ b/src/pages/dashboards/Dashboard.tsx
@@ -129,11 +129,8 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
         iframeSandboxExtras: ['allow-top-navigation', 'allow-popups-to-escape-sandbox'],
       })
 
-      // Cleanup may have run while the embed was in flight — StrictMode's double
-      // mount, a dashboard tab switch, a fast navigate. The SDK puts its iframe in
-      // the DOM before `embedDashboard` resolves, so the cleanup below had no
-      // `embedded` to unmount: do it here, otherwise the iframe stays orphaned in
-      // the DOM with its Switchboard port open.
+      // The SDK mounts its iframe before `embedDashboard` resolves, so cleanup that
+      // ran while this was in flight had no `embedded` to unmount.
       if (disposed) {
         embedded.unmount()
 
@@ -147,10 +144,8 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
       dashboardRef.current = dashboard.id
     }
 
-    // `embedDashboard` rejects if Superset hands back a token it can't decode
-    // (`jwtDecode` throws inside `getGuestTokenRefreshTiming`). Nothing else
-    // observes this promise, so without a catch that surfaces as an unhandled
-    // rejection and a silently blank dashboard.
+    // Nothing else observes this promise: uncaught, a bad token from Superset is an
+    // unhandled rejection and a silently blank dashboard.
     mount().catch((mountError) => {
       captureException(mountError, {
         tags: { errorType: 'SupersetDashboardMountError', component: 'Dashboard' },
--- a/src/pages/dashboards/Dashboard.tsx
+++ b/src/pages/dashboards/Dashboard.tsx
@@ -1,4 +1,5 @@
 import { gql, useApolloClient } from '@apollo/client'
+import { captureException } from '@sentry/react'
 import { embedDashboard, EmbeddedDashboard } from '@superset-ui/embedded-sdk'
 import { debounce } from 'lodash'
 import { useEffect, useMemo, useRef } from 'react'
@@ -65,6 +66,7 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
     }
 
     let embedded: null | EmbeddedDashboard = null
+    let disposed = false
 
     const persistFilters = isFeatureFlagActive(FeatureFlags.SUPERSET_PERSISTENT_FILTERS)
     // Filter persistence key is scoped to the org from the URL slug (resolved
@@ -127,6 +129,17 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
         iframeSandboxExtras: ['allow-top-navigation', 'allow-popups-to-escape-sandbox'],
       })
 
+      // Cleanup may have run while the embed was in flight — StrictMode's double
+      // mount, a dashboard tab switch, a fast navigate. The SDK puts its iframe in
+      // the DOM before `embedDashboard` resolves, so the cleanup below had no
+      // `embedded` to unmount: do it here, otherwise the iframe stays orphaned in
+      // the DOM with its Switchboard port open.
+      if (disposed) {
+        embedded.unmount()
+
+        return
+      }
+
       if (debouncedSaveFilters) {
         embedded.observeDataMask(debouncedSaveFilters)
       }
@@ -134,9 +147,19 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
       dashboardRef.current = dashboard.id
     }
 
-    mount()
+    // `embedDashboard` rejects if Superset hands back a token it can't decode
+    // (`jwtDecode` throws inside `getGuestTokenRefreshTiming`). Nothing else
+    // observes this promise, so without a catch that surfaces as an unhandled
+    // rejection and a silently blank dashboard.
+    mount().catch((mountError) => {
+      captureException(mountError, {
+        tags: { errorType: 'SupersetDashboardMountError', component: 'Dashboard' },
+        extra: { dashboardId: dashboard.id },
+      })
+    })
 
     return () => {
+      disposed = true
       fetchGuestToken.cancel()
       debouncedSaveFilters?.cancel()
       embedded?.unmount()
--- a/src/pages/dashboards/__tests__/Dashboard.test.tsx
+++ b/src/pages/dashboards/__tests__/Dashboard.test.tsx
@@ -84,9 +84,7 @@ const successMock: TestMocksType = [
       query: CreateSupersetGuestTokenDocument,
       variables: { input: { dashboardId: 'dash-1' } },
     },
-    // Deliberately different from the `guestToken` seeded by the dashboards query
-    // above: if the two matched, an assertion on the fetched token would pass even
-    // when the mutation never ran or threw and fell back to the seed.
+    // Must differ from the seed above, or the assertion passes on the fallback too.
     result: { data: { createSupersetGuestToken: { guestToken: 'refreshed-token' } } },
   },
 ]
@@ -116,8 +114,6 @@ const renderRevenue = (mocks: TestMocksType = successMock) =>
   )
 
 // Whether `promise` has settled once microtasks and 0ms timers have flushed.
-// Racing against an already-resolved sentinel instead would always pick the
-// sentinel, so such an assertion could never fail.
 const hasSettled = async (promise: Promise<unknown>): Promise<boolean> => {
   let settled = false
 
@@ -168,7 +164,6 @@ describe('Dashboard', () => {
       expect(config.supersetDomain).toBe('https://localhost:8089')
       expect(config.mountPoint).toBe(document.getElementById('superset-lago-dashboard'))
       expect(config.dashboardUiConfig.hideTitle).toBe(true)
-      // Proves the token came from the mutation, not from the query's seed.
       await expect(config.fetchGuestToken()).resolves.toBe('refreshed-token')
     })
   })
@@ -262,10 +257,8 @@ describe('Dashboard', () => {
       expect(mockUnmount).toHaveBeenCalled()
     })
 
-    // The SDK's `unmount()` only clears the iframe (`index.js:159-163`); it never
-    // cancels the refresh `setTimeout` it scheduled, so cleanup has to cancel the
-    // fetcher too. Without that, the chain keeps minting tokens for the rest of the
-    // SPA session — every revisit or dashboard switch leaving another one behind.
+    // The SDK's `unmount()` clears the iframe but not its refresh `setTimeout`
+    // (`index.js:159-163`), so cleanup has to cancel the fetcher itself.
     it('THEN cancels the guest token fetcher so the refresh chain stops', async () => {
       const { unmount } = renderAnalytics()
 
@@ -275,16 +268,12 @@ describe('Dashboard', () => {
 
       unmount()
 
-      // A cancelled fetcher deliberately never settles — that is the only way to
-      // halt a chain that re-arms from our resolution. Settling here means cleanup
-      // left the loop alive: the mutation mock would have resolved it.
+      // A cancelled fetcher never settles; the mutation mock would have resolved it.
       expect(await hasSettled(fetchGuestToken())).toBe(false)
     })
 
-    // The SDK puts its iframe in the DOM before `embedDashboard` resolves, so an
-    // embed still in flight at cleanup time has to be unmounted once it lands —
-    // otherwise the iframe is orphaned with its Switchboard port open. StrictMode
-    // hits this on every dev mount.
+    // The SDK mounts its iframe before `embedDashboard` resolves, so an embed still
+    // in flight at cleanup time is orphaned unless it is unmounted once it lands.
     it('THEN tears down an embed that only resolves after cleanup ran', async () => {
       let resolveEmbed: (value: unknown) => void = () => {}
 
@@ -304,7 +293,6 @@ describe('Dashboard', () => {
       resolveEmbed({ unmount: mockUnmount, observeDataMask: mockObserveDataMask })
 
       await waitFor(() => expect(mockUnmount).toHaveBeenCalledTimes(1))
-      // The late embed must not wire up filter observation either.
       expect(mockObserveDataMask).not.toHaveBeenCalled()
     })
   })
--- a/src/pages/dashboards/__tests__/Dashboard.test.tsx
+++ b/src/pages/dashboards/__tests__/Dashboard.test.tsx
@@ -115,6 +115,28 @@ const renderRevenue = (mocks: TestMocksType = successMock) =>
     { mocks },
   )
 
+// Whether `promise` has settled once microtasks and 0ms timers have flushed.
+// Racing against an already-resolved sentinel instead would always pick the
+// sentinel, so such an assertion could never fail.
+const hasSettled = async (promise: Promise<unknown>): Promise<boolean> => {
+  let settled = false
+
+  promise.then(
+    () => {
+      settled = true
+    },
+    () => {
+      settled = true
+    },
+  )
+
+  await new Promise<void>((resolve) => {
+    setTimeout(resolve, 0)
+  })
+
+  return settled
+}
+
 describe('Dashboard', () => {
   beforeEach(() => {
     jest.clearAllMocks()
@@ -238,6 +260,25 @@ describe('Dashboard', () => {
       unmount()
 
       expect(mockUnmount).toHaveBeenCalled()
+    })
+
+    // The SDK's `unmount()` only clears the iframe (`index.js:159-163`); it never
+    // cancels the refresh `setTimeout` it scheduled, so cleanup has to cancel the
+    // fetcher too. Without that, the chain keeps minting tokens for the rest of the
+    // SPA session — every revisit or dashboard switch leaving another one behind.
+    it('THEN cancels the guest token fetcher so the refresh chain stops', async () => {
+      const { unmount } = renderAnalytics()
+
+      await waitFor(() => expect(mockEmbedDashboard).toHaveBeenCalledTimes(1))
+
+      const { fetchGuestToken } = mockEmbedDashboard.mock.calls[0][0]
+
+      unmount()
+
+      // A cancelled fetcher deliberately never settles — that is the only way to
+      // halt a chain that re-arms from our resolution. Settling here means cleanup
+      // left the loop alive: the mutation mock would have resolved it.
+      expect(await hasSettled(fetchGuestToken())).toBe(false)
     })
 
     // The SDK puts its iframe in the DOM before `embedDashboard` resolves, so an
--- a/src/pages/dashboards/__tests__/Dashboard.test.tsx
+++ b/src/pages/dashboards/__tests__/Dashboard.test.tsx
@@ -2,8 +2,10 @@ import { screen, waitFor } from '@testing-library/react'
 
 import { GENERIC_PLACEHOLDER_TEST_ID } from '~/components/designSystem/GenericPlaceholder'
 import { getItemFromLS, setItemFromLS } from '~/core/utils/localStorage'
-import { SupersetDashboardsDocument } from '~/generated/graphql'
-import { CREATE_SUPERSET_GUEST_TOKEN } from '~/pages/dashboards/fetchSupersetGuestToken'
+import {
+  CreateSupersetGuestTokenDocument,
+  SupersetDashboardsDocument,
+} from '~/generated/graphql'
 import { render, TestMocksType } from '~/test-utils'
 
 import Dashboard, { DASHBOARD_MOUNT_TEST_ID } from '../Dashboard'
@@ -82,7 +84,7 @@ const successMock: TestMocksType = [
   { request: { query: SupersetDashboardsDocument }, result: { data: dashboardsData } },
   {
     request: {
-      query: CREATE_SUPERSET_GUEST_TOKEN,
+      query: CreateSupersetGuestTokenDocument,
       variables: { input: { dashboardId: 'dash-1' } },
     },
     result: { data: { createSupersetGuestToken: { guestToken: 'token-1' } } },
--- a/src/pages/dashboards/__tests__/Dashboard.test.tsx
+++ b/src/pages/dashboards/__tests__/Dashboard.test.tsx
@@ -84,7 +84,10 @@ const successMock: TestMocksType = [
       query: CreateSupersetGuestTokenDocument,
       variables: { input: { dashboardId: 'dash-1' } },
     },
-    result: { data: { createSupersetGuestToken: { guestToken: 'token-1' } } },
+    // Deliberately different from the `guestToken` seeded by the dashboards query
+    // above: if the two matched, an assertion on the fetched token would pass even
+    // when the mutation never ran or threw and fell back to the seed.
+    result: { data: { createSupersetGuestToken: { guestToken: 'refreshed-token' } } },
   },
 ]
 
@@ -143,7 +146,8 @@ describe('Dashboard', () => {
       expect(config.supersetDomain).toBe('https://localhost:8089')
       expect(config.mountPoint).toBe(document.getElementById('superset-lago-dashboard'))
       expect(config.dashboardUiConfig.hideTitle).toBe(true)
-      await expect(config.fetchGuestToken()).resolves.toBe('token-1')
+      // Proves the token came from the mutation, not from the query's seed.
+      await expect(config.fetchGuestToken()).resolves.toBe('refreshed-token')
     })
   })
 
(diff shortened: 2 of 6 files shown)


src/pages/dashboards/Dashboard.tsx
```diff
@@ -97,7 +99,7 @@ const Dashboards = () => {
         id: dashboard.embeddedId,
         supersetDomain: lagoSupersetUrl,
         mountPoint,
-        fetchGuestToken: async () => dashboard?.guestToken,
+        fetchGuestToken: createFetchSupersetGuestToken(client, dashboard.id),
```
osmarluz: **Issue 1 — one failed refresh permanently kills renewal**

The SDK's refresh loop re-arms after awaiting our callback, and doesn't observe the promise:

```javascript
async function refreshGuestToken() {
  const newGuestToken = await fetchGuestToken();          // if this rejects…
  ourPort.emit('guestToken', { guestToken: newGuestToken });
  setTimeout(refreshGuestToken, getGuestTokenRefreshTiming(newGuestToken)); // …this never runs
}
```

`createFetchSupersetGuestToken` does `await client.mutate(...)` with no `try/catch` . Any rejection — a network blip, an expired session, or a server error when the backend re-authenticates to Superset on every renewal and gets rate-limited — rejects refreshGuestToken, so the next refresh is never scheduled. The dashboard silently reverts to the exact "leave analytics open → error → reload" behaviour we're trying to fix, except now it's intermittent and only shows up under load. Once the loop dies, only a full page reload recovers it.

src/pages/dashboards/Dashboard.tsx
```diff
@@ -123,7 +125,7 @@ const Dashboards = () => {
       embedded?.unmount()
```
osmarluz, line 160: **Issue 2 — unmount leaks a refresh loop that keeps calling the mutation forever**

The SDK's `unmount()` only clears the iframe; it never cancels the refresh timer chain. After the effect cleanup runs embedded.unmount(), the last-scheduled refresh still fires, calls the mutation again, and re-arms itself — and the closure keeps client and dashboardId alive, so it runs for the rest of the SPA session. Since we're an SPA, navigating away from analytics doesn't unload the page, and every revisit or dashboard switch (which remounts the embed at `Dashboards.tsx:102`) leaves behind another perpetual loop. Each one re-authenticates against Superset every ~5 minutes indefinitely, which compounds the rate-limiting that triggers Issue 1.

src/pages/dashboards/Dashboard.tsx
ancorcruz: Both spot on — confirmed against the embed SDK source and pushed a fix.

The fetcher no longer rejects. On a failed mint it returns the last known token — seeded with the initial token from the dashboards query — so the SDK keeps scheduling the next refresh and retries, instead of dying until a full reload.

src/pages/dashboards/Dashboard.tsx
ancorcruz: The fetcher now exposes a cancel() that the effect cleanup calls alongside unmount(). Once cancelled it returns a never-settling promise, so the SDK's refresh chain stops re-arming — no more mutations after unmount, and no unhandled rejection (which is why I return a pending promise rather than throwing).

Added helper tests covering all of it: fallback-on-failure (resolves, doesn't reject), keeps-latest-good-token across a later failure, and cancel → stops hitting the mutation.

ansmonjol:
Reviewed the hardening commits. Direction is right and both of osmarluz's issues are genuinely addressed. Backend contract verified against `getlago/lago-api` main: `CreateSupersetGuestTokenInput { dashboardId: ID! }` -> `SupersetGuestToken { guestToken: String! }`, permission `analytics:view`, so the wire shape is correct. Also worth noting the never-settling promise is *not* a leak: the suspended frame and the promise reference each other and are otherwise unreachable, so V8 can collect the cycle.

---

### 1. HIGH: the failure path becomes a 5-second retry storm, with no backoff

`src/pages/dashboards/fetchSupersetGuestToken.ts:39`

On failure the fetcher returns `lastToken`, which is an already-expired JWT. The SDK then derives the next refresh delay from that very token:

```js
// node_modules/@superset-ui/embedded-sdk/lib/guestTokenRefresh.js:37-39
const ttl = isValidDate ? Math.max(MIN_REFRESH_WAIT_MS /* 10000 */, exp.getTime() - Date.now()) : DEFAULT_TOKEN_EXP_MS
return ttl - REFRESH_TIMING_BUFFER_MS // 5000
```

For an expired token `exp - now` is negative, so `ttl` clamps to `MIN_REFRESH_WAIT_MS` (10000) and the timing comes out at **5000 ms**. Every failed mint therefore schedules another mint 5 seconds later, indefinitely, for as long as the analytics tab stays open: roughly 720 mutations per hour per tab, each one making the backend re-authenticate against Superset.

That is exactly the rate-limit pressure flagged as the trigger for Issue 1. As written, the change trades "the loop dies permanently" for "the loop hammers the endpoint permanently", and the user still sees broken charts in the meantime because the token we keep emitting is expired. Only a genuinely transient failure benefits from the current fallback.

Suggested fix: retry with exponential backoff *inside* the callback (the SDK awaits it, so that is the only timing lever we control), cap the number of attempts, and only then resolve with `lastToken`.

### 2. MEDIUM: mint failures are completely silent

`fetchSupersetGuestToken.ts:38` is a bare `catch {}`. No Sentry, no console, no toast. Combined with #1 a permanently broken mint is invisible: the dashboard shows Superset errors, the network tab shows a 5 second poll, and nothing reaches Sentry.

The repo already has the pattern (`src/App.tsx:63`, `src/layouts/MainNavLayout/OrganizationSwitcher.tsx:108`). Capture on the first failure with a tag such as `errorType: 'SupersetGuestTokenMintError'`, and avoid re-reporting on every retry.

### 3. MEDIUM: codegen was not run, so the mutation is untyped `any`

`src/generated/graphql.tsx` contains no `createSupersetGuestToken` and no `CreateSupersetGuestTokenInput` (grep comes back empty). CLAUDE.md is explicit: "After any GraphQL schema/query/fragment changes, run `pnpm codegen`".

The consequence is not cosmetic: `data` off `client.mutate` is `any`, so `data?.createSupersetGuestToken?.guestToken` is type-checked against nothing. A field rename or schema drift compiles cleanly and fails at runtime straight into the silent catch from #2. The `Run Codegen` CI job regenerates the file in place and then runs `tsc`, it never diffs, so it cannot catch this either.

It also leaves the same test file internally inconsistent: the query mock uses the generated `SupersetDashboardsDocument` while the mutation mock uses the hand-rolled `CREATE_SUPERSET_GUEST_TOKEN`.

Suggested fix: run `pnpm codegen`, commit the result, then use `client.mutate<CreateSupersetGuestTokenMutation, CreateSupersetGuestTokenMutationVariables>({ mutation: CreateSupersetGuestTokenDocument, ... })` and drop the hand-rolled constant.

### 4. MEDIUM: `fetchSupersetGuestToken.test.ts:66-78` cannot fail

`settledOrPending` races the subject against `Promise.resolve(PENDING)`, which is already fulfilled and therefore always wins the race, regardless of whether the subject settles a microtask later.

Verified locally: deleting the post-await guard on line 37 (`return cancelled ? haltRefreshLoop() : lastToken` -> `return lastToken`) leaves all 5 tests passing. The same weakness makes the resolve assertion in the preceding test vacuous too, although its `mutate` not-called assertion does carry real signal.

```ts
let settled = false

pending.then(() => { settled = true }, () => { settled = true })
await new Promise((resolve) => setTimeout(resolve, 0))

expect(settled).toBe(false)
```

### 5. MEDIUM: the `Dashboard.test.tsx:147` mutation mock proves nothing

The mock returns `guestToken: 'token-1'` and the seed `dashboard.guestToken` is *also* `'token-1'`, so `await expect(config.fetchGuestToken()).resolves.toBe('token-1')` passes identically whether the mutation ran or threw and fell back to the seed.

Verified locally: forcing every mint to throw still leaves all 9 Dashboard tests green, so the assertion added in d321404 does not currently exercise the refresh flow. Making the mutation return a distinct token (`'refreshed-token'`) and asserting that would restore the signal.

src/pages/dashboards/Dashboard.tsx
```diff
@@ -129,11 +137,12 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
     mount()
 
     return () => {
+      fetchGuestToken.cancel()
```
AllanMichay, line 158: **`cancel()` during the initial embed leaves an orphaned iframe + open MessageChannel forever.**

The SDK does `await Promise.all([fetchGuestToken(), mountIframe()])` (`index.js:146`). Cleanup order here is `fetchGuestToken.cancel()` → `embedded?.unmount()`. If cleanup fires before `mount()` resolves, `fetchGuestToken()` never settles → `embedDashboard` never resolves → `embedded` stays `null` → `unmount()` never runs. The iframe was already placed via `replaceChildren` (`index.js:142`), the Switchboard port is open, and the closure is pinned permanently.

`React.StrictMode` is on (`src/main.tsx:116`), so this fires on **every dev mount**. In prod the window is now a full network round-trip (~100-500ms) wide, so a dashboard tab switch or a fast navigate hits it.

The orphan race pre-exists (`embedded` is assigned after cleanup already ran), but this upgrades it from "resolves, GC eventually collects" to "pinned forever".

Suggested fix — let `cancel()` settle an in-flight call and hang only *subsequent* ones, plus a disposed guard in the effect:

```ts
// factory: only the `if (cancelled)` entry guard halts.
// Post-await, return `lastToken` regardless so embedDashboard can finish.

// Dashboard.tsx
let disposed = false

const mount = async () => {
  embedded = await embedDashboard({ /* … */ })
  if (disposed) {
    embedded.unmount()
    return
  }
  // …
}

return () => {
  disposed = true
  fetchGuestToken.cancel()
  embedded?.unmount()
  // …
}
```

Note this inverts the `'never resolves if cancelled while a mutation is in flight'` test, which currently locks in the hang.

src/pages/dashboards/fetchSupersetGuestToken.ts
```diff
@@ -0,0 +1,48 @@
...
+  let lastToken = initialToken
+  let cancelled = false
+
+  const fetchGuestToken = async (): Promise<string> => {
+    if (cancelled) {
+      return haltRefreshLoop()
+    }
+
+    try {
+      const { data } = await client.mutate({
```
AllanMichay: **`pnpm codegen` wasn't run — the mutation is untyped and the generated file is stale.**

`createSupersetGuestToken` and `CreateSupersetGuestTokenInput` appear **zero times** in the committed `src/generated/graphql.tsx`. CLAUDE.md requires `pnpm codegen` after any GraphQL change, and every recent PR adding an operation commits the regenerated file (#4121, #4118, #4114).

Green CI isn't evidence here: the codegen job regenerates against a live lago-api and then runs `tsc` (`.github/workflows/codegen.yml:72-79`) — it never diffs against the checked-in file. So the mutation does exist server-side (that's why it passes), but the repo file is stale for everyone running `tsc` locally, and `data` is `any`, making `data?.createSupersetGuestToken?.guestToken` completely unchecked.

Run codegen, commit the result, and type the call the way `cacheUtils.ts:56-72` does:

```ts
const { data } = await client.mutate<
  CreateSupersetGuestTokenMutation,
  CreateSupersetGuestTokenMutationVariables
>({
  mutation: CreateSupersetGuestTokenDocument,
  variables: { input: { dashboardId } },
  fetchPolicy: 'no-cache',
})
```

Then the hand-rolled `CREATE_SUPERSET_GUEST_TOKEN` export can go, and `Dashboard.test.tsx` imports the generated Document instead.

src/pages/dashboards/fetchSupersetGuestToken.ts
```diff
@@ -0,0 +1,48 @@
...
+        mutation: CREATE_SUPERSET_GUEST_TOKEN,
+        variables: { input: { dashboardId } },
+        fetchPolicy: 'no-cache',
+      })
+
+      lastToken = data?.createSupersetGuestToken?.guestToken ?? lastToken
+
+      return cancelled ? haltRefreshLoop() : lastToken
+    } catch {
```
AllanMichay: **Bare `catch {}` gives zero observability.**

If the mutation starts failing, dashboards silently regress to exactly the stale-token bug this PR fixes — no signal anywhere, and the next report is again "leave analytics open, change a filter, get an error".

At minimum log it; ideally report it. Same applies to the `?? lastToken` branch on line 35 when the API returns a null token — that's an API contract violation worth surfacing, not swallowing.

src/pages/dashboards/fetchSupersetGuestToken.ts
```diff
@@ -0,0 +1,48 @@
...
+`
+
+export type FetchSupersetGuestToken = (() => Promise<string>) & { cancel: () => void }
+
+const haltRefreshLoop = (): Promise<string> => new Promise<string>(() => {})
+
+export const createFetchSupersetGuestToken = (
+  client: ApolloClient<object>,
+  dashboardId: string,
+  initialToken = '',
```
AllanMichay: **`initialToken = ''` is a footgun.**

The SDK calls `jwtDecode(guestToken)` (`guestTokenRefresh.js:34`), which throws on a non-JWT. `mount()` in `Dashboard.tsx` has no `try/catch`, so that's an unhandled rejection and a blank dashboard with no error UI.

`Dashboard.tsx` always passes `dashboard.guestToken`, so it isn't reachable today — but the default invites a caller that omits it, and this PR just put a network call in front of the decode. Either drop the default and make the param required, or wrap `mount()` in a `try/catch` that renders the existing error placeholder.

src/pages/dashboards/fetchSupersetGuestToken.ts
```diff
@@ -0,0 +1,48 @@
...
+  mutation createSupersetGuestToken($input: CreateSupersetGuestTokenInput!) {
+    createSupersetGuestToken(input: $input) {
+      guestToken
+    }
+  }
+`
+
+export type FetchSupersetGuestToken = (() => Promise<string>) & { cancel: () => void }
+
+const haltRefreshLoop = (): Promise<string> => new Promise<string>(() => {})
```
AllanMichay, line 37: **Worth a why-comment.**

A deliberately never-resolving promise is the most surprising line in the PR, and it reads like a bug at a glance. One line explaining that the SDK's `unmount()` never clears its refresh `setTimeout` (`@superset-ui/embedded-sdk/lib/index.js:156-163`), so declining to resolve is the only way to break the chain, saves the next reader a trip through `node_modules`.

src/pages/dashboards/__tests__/Dashboard.test.tsx
```diff
@@ -79,6 +80,13 @@ const dashboardsData = {
 
 const successMock: TestMocksType = [
   { request: { query: SupersetDashboardsDocument }, result: { data: dashboardsData } },
+  {
+    request: {
+      query: CREATE_SUPERSET_GUEST_TOKEN,
+      variables: { input: { dashboardId: 'dash-1' } },
+    },
```
AllanMichay, line 86: **This assertion can't fail.**

The mutation mock returns `'token-1'`, and `dashboard.guestToken` in `dashboardsData` is also `'token-1'`. So `await expect(config.fetchGuestToken()).resolves.toBe('token-1')` (line 139) passes whether the token was freshly minted or the mutation blew up and the fallback kicked in — i.e. it passes even if this PR's entire mechanism is broken.

Return `'refreshed-token-1'` from the mutation mock so the assertion actually proves the mutation path ran.

Also missing: a case asserting `cancel()` runs on unmount.

AllanMichay (commented):
## Review — guest token refresh

Diagnosis is correct and the approach is sound. The SDK's `unmount()` only does `mountPoint.replaceChildren()` (`@superset-ui/embedded-sdk/lib/index.js:159-163`) — it never clears its own `setTimeout(refreshGuestToken, …)`, so a never-resolving promise really is the only way to stop the loop. Good catch, that isn't obvious without reading the SDK.

Six inline comments. Two I'd want addressed before merge:

**Blocking-ish**
1. `Dashboard.tsx` cleanup — `cancel()` during the initial embed hangs `embedDashboard` forever, leaving an orphaned iframe and an open MessageChannel. Fires on every dev mount under StrictMode.
2. `fetchSupersetGuestToken.ts` — `pnpm codegen` wasn't run, so the mutation is untyped and `src/generated/graphql.tsx` is stale. Green CI doesn't prove otherwise (see the comment).

**Should fix**
3. Bare `catch {}` — silently regresses to the exact bug this PR fixes.
4. `initialToken = ''` default → `jwtDecode` throws inside the SDK.
5. `haltRefreshLoop` needs a why-comment.
6. Test can't distinguish the fix from the fallback.

**Nit, description only:** "runs with a no-cache fetch policy so a token is never served stale from the Apollo cache" — mutations never read the cache. `no-cache` here only prevents *writing* the token into it, which is a fine reason, just not the stated one.

**Checked, all fine:** `ApolloClient<object>` matches convention · `client` in the effect deps is safe (`useApolloClient` is stable) · org scoping is handled by the auth link's `x-lago-organization` at call time · inline `gql` + `client.mutate` in a `.ts` file has precedent (`cacheUtils.ts:56`) and `codegen.yml` globs `src/**/*.ts`, so the document is picked up.

The factory's unit tests are genuinely good — the cancellation cases in particular. Fix 1 and 2 and this is ready.

ancorcruz:
@ansmonjol @AllanMichay I have addressed your comments, could you take another look please?. Thanks.

ansmonjol (commented):
Reviewed the diff against the `@superset-ui/embedded-sdk@0.3.0` internals and the app's Apollo link chain. The core fix (mint on demand, never reject, cancel on teardown) is right, and the reasoning in the description matches what the SDK actually does.

One blocking issue: the new `disposed` teardown calls `embedded.unmount()`, which is `mountPoint.replaceChildren()` on a node that a later effect run may already own. Details inline.

Three other things worth resolving before merge: the `catch` branch in `fetchSupersetGuestToken` is unreachable under the app's `errorPolicy: 'all'`, the mutation is not marked `silentError` so every failed mint raises a danger toast, and the local retry loop stacks on top of Apollo's `RetryLink`. The rest are non-blocking.

src/pages/dashboards/Dashboard.tsx
```diff
@@ -119,21 +129,38 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
         iframeSandboxExtras: ['allow-top-navigation', 'allow-popups-to-escape-sandbox'],
       })
 
+      // The SDK mounts its iframe before `embedDashboard` resolves, so cleanup that
+      // ran while this was in flight had no `embedded` to unmount.
+      if (disposed) {
+        embedded.unmount()
```
ansmonjol, line 135: The intent here is right, but `unmount()` is not scoped to this embed. In the SDK it is literally:

```js
function unmount() {
  mountPoint.replaceChildren();   // lib/index.js
}
```

a blind wipe of whatever the mount node currently holds. And `mountIframe()` does `mountPoint.replaceChildren(iframe)` on that same node. Both effect runs get the same `document.getElementById(mountId)`, so a late-landing embed clears its successor's iframe.

Concretely: the effect re-runs while embed #1 is still awaiting its mint (`currentMembership?.organization.id` flips from `undefined` to the id once `getCurrentUserInfos` resolves; a `supersetDashboards` refetch returning a fresh `guestToken` does the same, see the deps comment below). Cleanup sets `disposed = true`, run #2 mounts iframe2. When embed #1 finally resolves, this branch runs `unmount()` and iframe2 is gone. The user gets a permanently blank dashboard until a full reload. Before this PR the stale embed was merely orphaned and the visible one kept working.

Note that `dashboardRef` does not protect against this: cleanup sets `dashboardRef.current = ''` before every re-run, so `dashboard?.id === dashboardRef?.current` is never true on re-entry.

Worth spelling out what the branch buys in each case, because only one of the three needs it:

| Case | Mount node state when the late unmount fires | Result |
|---|---|---|
| Component unmount (navigate away) | React already removed the div | no-op |
| Effect re-run, successor mounted | successor already did `replaceChildren(iframe2)` | destroys iframe2 |
| Effect re-run, new run early-returns on `!dashboard` | iframe1 still attached, orphaned | correct, the real justification |

(The refresh chain itself is already fully handled by `fetchGuestToken.cancel()`: the next `refreshGuestToken()` hits the entry guard and gets `haltRefreshLoop()`. `unmount()` never touched the timer or the Switchboard port.)

Suggested fix, a generation guard so only the run that still owns the node may clear it:

```tsx
const runIdRef = useRef(0)

useEffect(() => {
  if (!dashboard || dashboard?.id === dashboardRef?.current) {
    return
  }

  runIdRef.current += 1

  const runId = runIdRef.current

  // ...

  // `unmount()` is `mountPoint.replaceChildren()` on the shared node, so only
  // clear it if no later run has taken it over.
  if (disposed) {
    if (runIdRef.current === runId) {
      embedded.unmount()
    }

    return
  }
```

The increment has to sit after the early return, so case 3 (new run bails on `!dashboard`) still unmounts.

The alternative is giving each run its own child container so the node is never shared, which is immune by construction, but it changes the DOM structure under `#mountId` and the generation guard is the smaller change.

src/pages/dashboards/fetchSupersetGuestToken.ts
```diff
@@ -0,0 +1,146 @@
...
+        const token = await mintToken()
+
+        if (token) {
+          failureStreak = 0
+          lastToken = token
+
+          return lastToken
+        }
+
+        // `guestToken` is non-null in the schema: an empty payload is a contract
```
ansmonjol, line 113: This branch is unreachable, and as a result every real backend error takes the "contract violation" path.

`src/core/apolloClient/init.ts:312` sets `defaultOptions.mutate.errorPolicy: 'all'`, `ApolloClient.mutate` merges it in, and `QueryManager` only throws an `ApolloError` for GraphQL errors when `errorPolicy === 'none'`. So when the backend answers `{ data: { createSupersetGuestToken: null }, errors: [...] }`, `mintToken()` resolves with `undefined`, the `catch` below never runs, and `reportMintError`'s `error instanceof ApolloError && graphQLErrors.length > 0` de-dup guard never fires. Execution lands here instead, on `captureMessage(..., { level: 'error' })`, which is exactly the second Sentry report the PR set out to avoid, tagged as a schema violation.

The comment is also not accurate: `createSupersetGuestToken` is `Maybe<SupersetGuestToken>` in the generated schema, so a null payload is the schema's normal error shape rather than a violation.

And since the resolved `errors` array is never inspected, permanent failures (bad Superset config, revoked access) get retried three times as if they were transient.

Suggest reading `errors` off the mutation result and branching on it: GraphQL errors present, stay quiet and let the errorLink report; no errors and no token, then it really is a contract violation.

src/pages/dashboards/fetchSupersetGuestToken.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  const mintToken = async (): Promise<string | undefined> => {
+    const { data } = await client.mutate<
+      CreateSupersetGuestTokenMutation,
+      CreateSupersetGuestTokenMutationVariables
+    >({
+      mutation: CreateSupersetGuestTokenDocument,
+      variables: { input: { dashboardId } },
+      fetchPolicy: 'no-cache',
```
ansmonjol, line 81: This needs `context: { silentError: true }`.

`errorLink` (`src/core/apolloClient/init.ts:261`) fires `addToast({ severity: 'danger', translateKey: 'text_622f7a3dc32ce100c46a5154' })` for every non-silenced GraphQL error; only `AUTH_ERRORS` and `Forbidden` are auto-silenced. So a Superset outage returning a generic error produces three danger toasts per invocation, then three more after each 5s, 10s, 20s cooldown, indefinitely, on a page the user is only reading. Before this PR a token failure was silent.

The established pattern is `context: { silentError: true }` (or `silentErrorCodes`), used at roughly 15 call sites, e.g. `src/components/customers/CustomerActivityLogs.tsx:53`.

src/pages/dashboards/fetchSupersetGuestToken.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (failureStreak > 0) {
+      await sleep(streakCooldown(failureStreak))
+
+      if (cancelled) return haltRefreshLoop()
+    }
+
+    const isFirstFailureOfStreak = (attempt: number): boolean =>
+      failureStreak === 0 && attempt === 1
+
+    for (let attempt = 1; attempt <= MAX_MINT_ATTEMPTS; attempt += 1) {
```
ansmonjol, line 102: This loop stacks on top of Apollo's `RetryLink`, so the effective request count is higher than the three the description reasons about.

`init.ts:151-179` installs `new RetryLink({ delay: { initial: 150, max: 5000, jitter: true }, attempts: { max: 3, retryIf: isNetworkError } })`, which is four attempts per `client.mutate` for exactly the network-error class this loop also retries. During an outage each `fetchGuestToken()` issues 3 x 4 = 12 requests over roughly 15 to 20 seconds, which works against the stated goal of not hammering the backend.

Either pass `context: { disableRetry: true }` (the link honours it) and keep this loop, or drop the loop and let the link do the retrying.

src/pages/dashboards/Dashboard.tsx
```diff
@@ -119,21 +129,38 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
...
 
     return () => {
+      disposed = true
+      fetchGuestToken.cancel()
       debouncedSaveFilters?.cancel()
       embedded?.unmount()
       dashboardRef.current = ''
     }
-  }, [dashboard, currentMembership?.organization.id, dashboardTitle, mountId])
+  }, [dashboard, currentMembership?.organization.id, client, dashboardTitle, mountId])
```
ansmonjol, line 163: Depending on the whole `dashboard` object makes the effect re-run on every `supersetDashboards` response. The backend mints a fresh JWT per response, so `guestToken` differs, the normalized cache entry changes, `data.supersetDashboards` gets a new identity, and the `useMemo` yields a new `dashboard`. With `cache-and-network` that happens on every cold load with a warm IndexedDB cache: the whole iframe is torn down and re-embedded, the user loses their active tab and filters, and a second mint fires.

The effect only needs `dashboard.id` and `dashboard.embeddedId`; `guestToken` is read once as a seed and does not need to be fresh. Narrowing to primitives stops the churn and also shrinks the race window for the teardown issue above:

```tsx
}, [
  dashboard?.id,
  dashboard?.embeddedId,
  currentMembership?.organization.id,
  client,
  dashboardTitle,
  mountId,
])
```

The `currentMembership?.organization.id` flip still re-runs the effect once per cold load, so this does not remove the need for the guard, but it removes the recurring case.

src/pages/dashboards/Dashboard.tsx
```diff
@@ -87,6 +91,12 @@ const Dashboard = ({ contentTitle, dashboardTitle, dashboardTitleTestKey }: Dash
         }, 500)
       : null
 
+    const fetchGuestToken = createFetchSupersetGuestToken(
+      client,
+      dashboard.id,
+      dashboard.guestToken,
```
ansmonjol, line 97: Two things about the seed token.

First, the description says the `no-cache` policy keeps the token out of a cache the app persists to IndexedDB. That is true of the mutation, but `supersetDashboards` still selects `guestToken` and `SupersetDashboard` has an `id`, so the token is normalized and `setupCachePersistor` (`src/core/apolloClient/cachePersistor.ts:47`) writes it to IndexedDB anyway. The `no-cache` policy is still worth having, but the claim in the description does not hold as written.

Second, `initialToken` is currently only a fallback. The SDK does `await Promise.all([fetchGuestToken(), mountIframe()])` before emitting to the iframe, so the very first render now blocks on a network mutation even though a valid token is already in hand. `init.ts:38` sets `TIMEOUT = 300000` and `timeoutLink` sits after `retryLink`, so each link-level attempt gets its own five-minute budget; multiplied by this fetcher's three attempts, a hung backend can leave the analytics page blank for a very long time where the old code rendered instantly.

Returning `initialToken` directly on the first invocation and minting only from the first refresh onward keeps the fix and takes a round trip off first paint.

src/pages/dashboards/fetchSupersetGuestToken.ts
```diff
@@ -0,0 +1,146 @@
...
+            level: 'error',
+            tags: SENTRY_TAGS,
+            extra: { dashboardId },
+          })
+        }
+      } catch (error) {
+        if (isFirstFailureOfStreak(attempt)) reportMintError(error)
+      }
+
+      if (cancelled) break
```
ansmonjol, line 126: These two `break`s fall through to `return lastToken` at the bottom, which is inconsistent with the entry and cooldown guards above that return `haltRefreshLoop()`.

If the user navigates away while attempt 1 has failed and the fetcher is in the 1s backoff, `cancel()` wakes the sleeper, this breaks, and the call resolves with the expired `lastToken`. The SDK then does `ourPort.emit('guestToken', ...)` on a Switchboard whose iframe is already gone and schedules another `setTimeout(refreshGuestToken, 5000)` (`getGuestTokenRefreshTiming` floors at 5s for an expired token), keeping the embed closure alive for another five seconds past teardown.

All three cancel checks should route to `haltRefreshLoop()`.

src/pages/dashboards/fetchSupersetGuestToken.ts
```diff
@@ -0,0 +1,146 @@
...
+): FetchSupersetGuestToken => {
+  let lastToken = initialToken
+  let cancelled = false
+  let failureStreak = 0
+  let wakeSleeper: null | (() => void) = null
+
+  const sleep = (ms: number): Promise<void> =>
+    new Promise<void>((resolve) => {
+      const timer = setTimeout(() => {
+        wakeSleeper = null
```
ansmonjol, line 55: `wakeSleeper` is a single shared slot and is unconditionally nulled by whichever timer fires first.

Today the SDK's refresh chain is strictly serial so no two sleeps overlap, but nothing here enforces that. If two `fetchGuestToken()` calls ever overlap (a second embed sharing the fetcher, or any future caller), sleep A's callback nulls the slot while sleep B is pending, so a later `cancel()` sees `null`, never clears B's timer, and B's retry loop keeps running after teardown. Starting B while A is pending leaks A's timer the same way.

A per-sleep local (a `Set` of wakers, or an `AbortSignal` threaded through) removes the hazard without changing the current behaviour.

src/pages/dashboards/__tests__/fetchSupersetGuestToken.test.ts
```diff
@@ -0,0 +1,274 @@
...
+
+      const second = fetchGuestToken()
+
+      await jest.advanceTimersByTimeAsync(FIRST_STREAK_COOLDOWN_MS)
+      await drainFailedInvocation(second)
+
+      expect(mockCaptureException).toHaveBeenCalledTimes(1)
+    })
+
+    it('THEN leaves GraphQL errors to the errorLink rather than reporting them twice', async () => {
```
ansmonjol, line 159: This test asserts a rejection the real client cannot produce, so it green-lights the unreachable guard flagged in `fetchSupersetGuestToken.ts`.

`makeClient` swaps in a bare `jest.Mock`, so `mockRejectedValue(new ApolloError({ graphQLErrors }))` works here. With the app's real `errorPolicy: 'all'`, `mutate` resolves with `{ data: { createSupersetGuestToken: null }, errors }` and never rejects. The branch under test is dead in production while the suite reports it as covered.

The test that would catch it does not exist: a resolved result carrying `errors`. Worth noting that the "token-less response" case at line 203 is silently exercising the real GraphQL-error path while asserting contract-violation semantics.

src/pages/dashboards/__tests__/Dashboard.test.tsx
```diff
@@ -228,5 +256,44 @@ describe('Dashboard', () => {
...
+
+      unmount()
+
+      // A cancelled fetcher never settles; the mutation mock would have resolved it.
+      expect(await hasSettled(fetchGuestToken())).toBe(false)
+    })
+
+    // The SDK mounts its iframe before `embedDashboard` resolves, so an embed still
+    // in flight at cleanup time is orphaned unless it is unmounted once it lands.
+    it('THEN tears down an embed that only resolves after cleanup ran', async () => {
```
ansmonjol, line 277: This test only covers the full-component-unmount path, where React removes the `#mountId` div anyway, so `unmount()` cannot do damage. The case that matters for the teardown issue is an effect re-run: embed #1 in flight, effect re-runs, embed #2 mounts into the same node, then embed #1 resolves.

Rough shape, driving the re-run through `currentMembership?.organization.id`, which is the trigger that fires on every cold load:

```tsx
it('THEN does not clear the mount node when a later run already took it over', async () => {
  let resolveFirstEmbed: (value: unknown) => void = () => {}
  const firstUnmount = jest.fn()

  mockEmbedDashboard
    .mockReturnValueOnce(
      new Promise((resolve) => {
        resolveFirstEmbed = resolve
      }),
    )
    .mockResolvedValue({ unmount: mockUnmount, observeDataMask: mockObserveDataMask })

  const { rerender } = renderAnalytics()

  await waitFor(() => expect(mockEmbedDashboard).toHaveBeenCalledTimes(1))

  // Effect re-run: the second embed takes over the mount node.
  mockCurrentMembership.mockReturnValue({ organization: { id: 'org-2' } })
  rerender(/* same element */)

  await waitFor(() => expect(mockEmbedDashboard).toHaveBeenCalledTimes(2))

  resolveFirstEmbed({ unmount: firstUnmount, observeDataMask: jest.fn() })
  await waitFor(() => expect(mockEmbedDashboard).toHaveBeenCalledTimes(2))

  // The stale embed must not clear the node the live embed now owns.
  expect(firstUnmount).not.toHaveBeenCalled()
})
```

Using a distinct `firstUnmount` mock per embed is what makes the assertion meaningful, since the current shared `mockUnmount` cannot tell the two embeds apart. Keep the existing test as the case-3 coverage.

src/pages/dashboards/__tests__/Dashboard.test.tsx
```diff
@@ -105,6 +113,26 @@ const renderRevenue = (mocks: TestMocksType = successMock) =>
     { mocks },
   )
 
+// Whether `promise` has settled once microtasks and 0ms timers have flushed.
+const hasSettled = async (promise: Promise<unknown>): Promise<boolean> => {
```
ansmonjol, line 117: This helper is byte-identical to the one in `fetchSupersetGuestToken.test.ts:34`, apart from `setTimeout(resolve, 0)` versus `jest.advanceTimersByTimeAsync(0)`. Any future change to the settle-detection logic has to be applied twice or the two suites drift. Worth extracting one helper that takes the flush strategy as an argument.

src/pages/dashboards/__tests__/fetchSupersetGuestToken.test.ts
```diff
@@ -0,0 +1,274 @@
...
+
+const mockCaptureException = jest.fn()
+const mockCaptureMessage = jest.fn()
+
+jest.mock('@sentry/react', () => ({
+  captureException: (...args: unknown[]) => mockCaptureException(...args),
+  captureMessage: (...args: unknown[]) => mockCaptureMessage(...args),
+}))
+
+const makeClient = (mutate: jest.Mock) => ({ mutate }) as unknown as ApolloClient<object>
```
ansmonjol, line 15: `makeClient` here and `tokenResponse` at line 29 have inferred return types. `.agents/docs/typescript-conventions.md` asks for explicit ones, and every other new helper in this PR (`hasSettled`, `drainFailedInvocation`, `streakCooldown`, `haltRefreshLoop`) has them. Typing `tokenResponse` against `CreateSupersetGuestTokenMutation` in particular would stop the fake payload drifting from the generated shape.

domenicofalco:
No real owner of this PR. Closing

PR #3857, opened 2026-07-07
feat: Allow wallet code editing

stephenlago99:
## Context

The wallet code can be set and updated through the API, but the UI wallet form did not expose it at all — there was no code input on creation or edition. This creates an API parity gap: users managing wallets from the app cannot set or change a wallet's code.

Fixes ING-366

## Description

Adds a `code` input to the wallet create/edit form, following the same name/code pattern used in the add-on, feature, and alert forms.

**No API change was needed.** The API already fully supports this:
- `Wallets::UpdateService` applies the code unconditionally (`wallet.code = params[:code] if params[:code]`) — there is no restriction on editing it, so the input is enabled in both creation and edition modes. A blank value is ignored by the API (the code is kept, never cleared), so the form omits the field when left empty.
- The only constraint is uniqueness: the code must be unique among the customer's **active** wallets (`unique_code_per_customer` validation + partial unique index). A violation returns a `value_already_exist` error on the `code` field.
- On creation, when no code is provided, the API auto-generates one from the wallet name.

Form behavior:
- On creation, the code is auto-filled from the name as the user types (via `updateNameAndMaybeCode`), and can be overridden manually. If left blank, nothing is sent and the API generates it.
- On edition, the existing code is loaded and freely editable.
- A `ValueAlreadyExist` mutation error is surfaced as a field error on the code input ("Code already existing, please type a new one"), matching the alert form behavior.

Also regenerated GraphQL types after adding `code` to the `WalletForUpdate` fragment.


src/pages/wallet/CreateWallet.tsx
```diff
@@ -380,6 +382,8 @@ const CreateWallet = () => {
       if (formType === FORM_TYPE_ENUM.edition) {
         const input = {
           ...values,
+          // Blank code is ignored to keep the current one, as the API would store it as is
+          code: values.code || undefined,
```
ansmonjol: What situation can lead to have the code not send?

ansmonjol: We don't enforce it on FE so BE fills it? Don't we want to change this behaviour to do as other forms in the app, so the code is always send by FE + mandatory?

stephenlago99: Good point, we could make it mandatory in the frontend

ansmonjol (approved):

PR #3861, opened 2026-07-07
fix(quotes): serialize pricing amounts as cents

AllanMichay:
## Context

While debugging, we saw that we were not sending amounts in cents, both in plans and add-ons. Coupons are okay

## Description

This PR serializes correctly the amounts sent

Fixes LAGO-1685

### DIFF
--- a/src/core/serializers/serializeQuoteBillingItems.ts
+++ b/src/core/serializers/serializeQuoteBillingItems.ts
@@ -110,7 +110,7 @@ interface FromBillingItemsResult {
 
 export const fromBillingItems = (
   billingItems: BillingItemsPayload,
-  currency: CurrencyEnum = CurrencyEnum.Usd,
+  currency?: CurrencyEnum,
 ): FromBillingItemsResult => {
   const entities: Record<string, EntityData> = {}
   const addOnItems: AddOnItem[] = []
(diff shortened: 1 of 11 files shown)


src/core/serializers/serializeQuoteBillingItems.ts
```diff
@@ -174,8 +186,9 @@ export const fromBillingItems = (billingItems: BillingItemsPayload): FromBilling
  */
 export const buildPreviewEntities = (
   billingItems: BillingItemsPayload,
+  currency: CurrencyEnum = CurrencyEnum.Usd,
```
domenicofalco: does it make sense to have USD as default? I mean what if the currency isn't passed along for any reason and the amount refers to a different currency but it forces USD?

domenicofalco (approved):

src/core/serializers/serializeQuoteBillingItems.ts
AllanMichay: Yeah you're right. 

We should always have a currency but in case we don't have one, we might not want to display any value there

AllanMichay: Decided to use org currency as default

PR #3806, opened 2026-07-02
feat: PO component / add PO to one-off invoices

stephenlago99:
### Context

Purchase order (PO) numbers are a common B2B/procurement requirement — customers need to reference their internal PO on the invoices they receive. This is the first slice of that feature, shipped incrementally: it introduces a reusable PO component and wires it into one-off invoices only. Subsequent slices (subscriptions, wallets, credit notes, invoice-list filtering) build on the same base component and land in separate PRs.

### Description

Reusable PO component (src/components/purchaseOrder/)
- A composable compound component: <PO> root (holds value / onChange / description context) with PO.Title, PO.Description, PO.Number, PO.AddButton, PO.EditButton, PO.TrashButton, and PO.DynamicInputButton sub-parts, so each surface can lay the PO field out however it needs.
- Editing goes through a TanStack + zod form dialog (usePurchaseOrderNumberDialogs) that validates the number is ≤ 255 characters.
- normalizePurchaseOrderNumber util trims input and coerces empty strings to null. Shared constants/types keep the translation keys and field name in one place.

#### One-off invoice creation (CreateInvoice.tsx)
- Adds a "Purchase order number" row directly under Issue date (200px label column, + Add number when empty; number + edit/trash when set).
- Client-side validation caps the value at 255 chars; the normalized value is sent on the createInvoice mutation.
- When regenerating from a voided invoice, the existing PO number is prefilled.

#### Invoice display (InvoiceCustomerInfos.tsx)
- Shows the PO number as an overview line on the invoice details page.

PR #3886, opened 2026-07-09
feat: PO Number - Wallet / Subscriptions

stephenlago99:
## Context

Purchase order (PO) numbers are a common B2B/procurement requirement — customers need to reference their internal PO on the invoices they receive. PR #3806 shipped the first slice: the reusable PO component, wired into one-off invoices. This is the second slice, extending PO numbers to the remaining invoice-generating objects: wallets, recurring transaction rules, wallet transactions and subscriptions.

⚠️ **Depends on** getlago/lago-api#5874, which introduces `purchaseOrderNumber` on the corresponding GraphQL objects and mutation inputs — this PR must be merged/deployed after it. Generated types were produced from that PR's `schema.graphql`.

### Shared component (`src/components/purchaseOrder/PurchaseOrderFormBlock.tsx`)

A vertical composition of the existing `<PO>` compound parts, shared by all four surfaces below: title + description, then a "+ Add purchase order number" button that reveals an inline input (focused on reveal) with a trash button that clears the value and collapses back to the button. Prefilled values (async edit-form loads) reveal the input without stealing focus. The input shows an inline error above 255 characters, and each form schema blocks submission past that limit. The input row mirrors the metrics of sibling input+trash rows (medium trash button), and `PO.TrashButton` gained a `size` prop for this. The one-off invoice flow from #3806 (horizontal row, dialog editing) is unchanged.

### Wallets

- **Create/edit wallet**: PO block in "Wallet settings" (under transaction top-up limits) and inside the recurring top-up rule's payment & invoicing settings. Prefilled on edit via `WalletForUpdate`; normalized on submit (`null` clears).
- **Wallet overview**: PO number displayed in the wallet settings grid.

### Wallet transactions

- **Top-up page**: PO block for **prepaid credits only** (free credits never generate an invoice); the value is cleared when switching to free credits.
- **Transaction details drawer**: PO number displayed after the transaction ID.

### Subscriptions

- **Form**: PO block in the subscription information section, after the start/end dates — it therefore appears in both the full create/edit page and the details-v2 "Subscription settings" edit drawer. Wired through `CreateSubscription` (create/upgrade/downgrade) and `useUpdateSubscriptionInformation` (drawer save), normalized with `null` to clear.
- **Details overview**: PO number displayed in the subscription information grid.


ansmonjol (commented):
Nice extension of the #3806 groundwork, the shared block keeps the four surfaces consistent. A few things worth addressing, all inline. Nothing blocking beyond the top-up prefill race (which can silently drop the PO).

Also: the description only covers wallets / subscriptions / transactions, but the diff also ships the PO filter on Invoices + Credit notes, makes the PO row show for every invoice type (was one-off only), and makes it editable on the regenerate page. Worth adding to the body so QA covers those paths.

src/components/purchaseOrder/PurchaseOrderFormBlock.tsx
```diff
@@ -0,0 +1,96 @@
...
+  // not when it shows up because a prefilled value loaded.
+  const [focusOnReveal, setFocusOnReveal] = useState(false)
+
+  // Prefilled values load asynchronously (e.g. the edit-wallet query resolving
+  // after mount) — reveal the input once a value lands.
+  useEffect(() => {
+    if (normalizePurchaseOrderNumber(value)) {
+      setShowInput(true)
+    }
+  }, [value])
```
ansmonjol: `showInput` is state plus an effect that mirrors `value`, so one boolean has two sources of truth. Deriving it removes the effect and both desync modes:

```tsx
const [revealed, setRevealed] = useState(false)
const showInput = revealed || !!normalizePurchaseOrderNumber(value)
```

(`setRevealed(true)` on add click, `setRevealed(false)` in the trash handler, `focusOnReveal` unchanged.)

The two desyncs it fixes:
1. Trash click sets `showInput = false` unconditionally. If the parent does not actually apply `onChange(null)`, the input disappears while a non-empty value is still in the form and gets submitted.
2. If the parent resets the form while the block stays mounted (drawer reopened), `showInput` stays `true` and you get a revealed empty input instead of the add button.

src/components/purchaseOrder/PurchaseOrderFormBlock.tsx
```diff
@@ -0,0 +1,96 @@
...
+            setFocusOnReveal(false)
+          }}
+        />
+      </div>
+    )
+  }
+
+  return (
+    <PurchaseOrder
+      className="gap-3"
```
ansmonjol: `className="gap-3"` is dead code here: `PurchaseOrderRoot` already applies `tw('flex flex-col gap-3', className)`, and `{...props}` (which still carries `className`, since only `value` / `onChange` / `disabled` are destructured) spreads after it, so a caller-provided `className` replaces it anyway. Can be dropped.

src/pages/wallet/topUp/formInitialization/validationSchema.ts
```diff
@@ -81,6 +85,17 @@ export const getTopUpFormValidationSchema = ({
...
+    // purchaseOrderNumber — optional, bounded length. PurchaseOrderFormBlock
+    // renders its own inline max-length error (plain value/onChange input,
+    // not an AppField), so this issue only gates the submit.
+    if ((data.purchaseOrderNumber?.length ?? 0) > PURCHASE_ORDER_NUMBER_MAX_LENGTH) {
+      ctx.addIssue({
+        code: 'custom',
+        message: PURCHASE_ORDER_TRANSLATIONS.maxLength,
+        path: ['purchaseOrderNumber'],
+      })
+    }
```
ansmonjol: This max-length rule now exists in four places: inline in `PurchaseOrderFormBlock` (the `error` prop), `subscriptionFormSchema.ts`, `wallet/formInitialization/validationSchema.ts` (as `addPurchaseOrderNumberIssue`), and here as a copy of that helper. Suggest exporting the single helper from `~/components/purchaseOrder/` (next to `PURCHASE_ORDER_NUMBER_MAX_LENGTH`) and importing it in the three schemas, so the bound and the message key live in one place.

Related: the filter input hard-caps typing with `inputProps={{ maxLength: PURCHASE_ORDER_NUMBER_MAX_LENGTH }}` while the form block lets you overflow and shows an error. Worth picking one behaviour; capping the form input too would make most of this validation unreachable.

src/pages/wallet/CreateWalletTopUp.tsx
```diff
@@ -178,8 +179,13 @@ const CreateWalletTopUp = () => {
...
     // Recomputed inline on every render: TanStack re-seeds an untouched form
-    // when defaults deep-change as the wallet query resolves.
-    defaultValues: mapFromApiToForm({ wallet }),
+    // when defaults deep-change as the wallet query resolves — the same
+    // mechanism prefills the PO number once the voided invoice resolves
+    // (regenerate flow: carry the voided invoice's PO over to the top-up).
+    defaultValues: mapFromApiToForm({
+      wallet,
+      purchaseOrderNumber: voidedInvoice?.invoice?.purchaseOrderNumber,
+    }),
```
ansmonjol, line 191: The prefill only lands while the form is still untouched, so there is a race: if the user types into `paidCredits` before `GetInvoiceStatus` resolves, the voided invoice's PO number is silently dropped and the regenerated invoice ships without it. No error, nothing visible.

Suggest either seeding it explicitly when the query resolves and the field is still pristine (`form.setFieldValue('purchaseOrderNumber', ...)`), or gating the section on that query's loading state. The current test passes because it waits on the prefill before typing.

src/pages/wallet/CreateWalletTopUp.tsx
```diff
@@ -241,6 +247,12 @@ const CreateWalletTopUp = () => {
 
     form.setFieldValue('grantedCredits', '')
     form.setFieldValue('paidCredits', '')
+
+    // Free credits never generate an invoice, so a PO number set while in
+    // prepaid mode must not survive the switch.
+    if (type === WalletTransactionType.FreeCredits) {
+      form.setFieldValue('purchaseOrderNumber', undefined)
+    }
```
ansmonjol, line 258: This clear-on-switch is advertised in the PR description but not covered by a test. Worth one case in `CreateWalletTopUp.test.tsx`: set a PO number in prepaid mode, switch to free credits, assert the mutation input has no PO number (and that switching back does not resurrect it).

Same gap on the mapper side: `src/pages/wallet/mappers/__tests__/mappers.test.ts` and `topUp/mappers/__tests__/mappers.test.ts` are untouched, so nothing pins `normalizePurchaseOrderNumber` sending `null` (not `undefined`) on clear for the wallet, the recurring rule, and the top-up. That is exactly the bug class the code comments guard against, and it is cheap to cover.

src/components/purchaseOrder/PurchaseOrderButtons.tsx
```diff
@@ -74,7 +74,8 @@ export const PurchaseOrderTrashButton = ({
   className,
   disabled,
   onClick,
-}: Omit<PurchaseOrderButtonProps, 'children'>) => {
+  size = 'small',
+}: Omit<PurchaseOrderButtonProps, 'children'> & { size?: 'small' | 'medium' | 'large' }) => {
```
ansmonjol: Rather than retyping the size union, reuse the Button's own type so it cannot drift:

```tsx
size?: ComponentProps<typeof Button>['size']
```

ansmonjol (approved):

PR #3889, opened 2026-07-10
feat(catalog): product drawer, create more and details page

ansmonjol:
Fixes BIL-256

Third step of the Products & rate cards initiative, stacked on #3809 (`feat/catalog-pricing-nav`): the product create/edit drawer with its real form, a reusable "Create more" drawer system, and the Product details page. Also carries the catalog Create menu, Products list and product item/filter drawer scaffolds from the previous step (they were only in the working tree).

### Product drawer (create + edit)

- Real TanStack form in `useProductDrawer`: name/code (required, zod), description reveal pattern, invoice display name section. `NameAndCodeGroup` auto-generates the code from the name in create mode only (`disableAutoGenerateCode` in edit).
- Create: on success the drawer closes, navigates to the new product's details and toasts.
- **Create more** toggle (footer left slot, create mode only): the drawer stays open, the form resets with a fade-in + scroll-to-top + Name refocus, and the success toast embeds a router link to the created product (slug-prefixed via `prependOrgSlug`, quotes escaped).
- Edit (`openDrawer(product)`): prefilled values, "Save edits", code locked when `attachedToPlanOrSubscription`, cleared optional fields sent as `null` so they actually clear.
- Duplicate code (`ValueAlreadyExist`): inline error under the Code input, drawer stays open, typing clears it.

### Product details page

- New route `/product-catalog/products/:productId/:tab` (Overview + Product items / Plans / Activity logs placeholder tabs; activity logs gated premium + `auditLogsView`).
- Breadcrumb with a link back to the catalog and a static grey "Product" crumb (new pathless `BreadcrumbItem` support).
- Actions dropdown: Edit (`productsUpdate`) reopens the drawer, Delete (`productsDelete`) opens a danger dialog that destroys, evicts from the Apollo cache (no post-delete 404 flash) and navigates back to the list.
- Overview tab: product information grid with copyable code and an inline Edit entry point.
- Products list rows now link to the details overview, and each row exposes Edit (opens the drawer prefilled) and Delete (confirm dialog) actions, permission-gated with a composed tooltip.
- Products list layout is inset to the page gutter (padded wrapper + `containerSize` 4 + non-sticky pager, same as the customer subscriptions tab) so row dividers stop at the gutter per the design.

### Shared groundwork

- `src/components/drawers/createMore/`: reusable create-more system for the upcoming product item / item filter drawers.
- Duplicate-code helpers hoisted to `~/core/form/existingCodeError.ts` (charge drawers keep their historical import via re-export).
- `ToastContainer` moved inside `BrowserRouter`: toast messages can now embed router links. Toasts fired from the devtools `MemoryRouter` still render (global `toastsVar`) and navigate the main app router, which is the intent.
- New `fade-in-right` tailwind animation; `BASE_DRAWER_CONTENT_TEST_ID` moved to `drawers/const.ts` so drawer bodies can target the scroll container without pulling BaseDrawer's `import.meta` chain into Jest.

## Notes

- `MainHeader.Configure` gets a `snapshotKey` covering `description`/`invoiceDisplayName`/`attachedToPlanOrSubscription`: the config snapshot strips the action closures, so without it a description-only edit left the header Edit action holding the stale product and re-saving reverted the edit (same pitfall documented in `SubscriptionDetails`).
- Known edge case left open on purpose (product decision): clicking the create-more toast link while the next product's form is dirty navigates away and discards the typed data without the unsaved-changes prompt.
- 9 test suites (66 tests) cover the drawer flows, create-more system, details page, delete dialog, breadcrumb and toast container.


src/core/form/existingCodeError.ts
```diff
@@ -0,0 +1,15 @@
+import { AnyFormApi } from '@tanstack/react-form'
+
+// Backend "code already exists" message, surfaced under the Code input on save
+// and cleared when the user edits the code (so submit re-enables). Shared by
+// every entity form with a unique code so the string can't drift between them.
+export const EXISTING_CODE_ERROR_MESSAGE = 'text_632a2d437e341dcc76817556'
+
+// Surfaces the backend duplicate-code rejection under the Code input (keeps the
+// drawer open). Same pattern as plan-settings code.
+export const applyExistingCodeError = (formApi: AnyFormApi): void => {
```
domenicofalco, line 10: should this pattern be shortly documented on the CLAUDE.md?

src/pages/catalog/drawers/product/ProductDrawerContent.tsx
```diff
@@ -0,0 +1,190 @@
...
+  props: productDrawerContentDefaultProps,
+  render: function ProductDrawerContentRender({ form, isEdit, disableCodeInput, resetSignal }) {
+    const rootRef = useRef<HTMLDivElement>(null)
+    const resetIteration = useCreateMoreResetIteration(resetSignal)
+
+    useEffect(() => {
+      if (resetIteration === 0) return
+
+      rootRef.current
+        ?.closest<HTMLElement>(`[data-test="${BASE_DRAWER_CONTENT_TEST_ID}"]`)
```
domenicofalco: I think we should be using an alternative data selector as `data-test` is for int tests only

src/pages/catalog/drawers/productItem/useProductItemDrawer.tsx
```diff
@@ -0,0 +1,66 @@
...
+    resetForm()
+
+    drawer.open({
+      title: translate('text_1783622030703m9jlurg4jsn'),
+      form: { id: PRODUCT_ITEM_FORM_ID, submit: form.handleSubmit },
+      closeOnSubmitSuccess: false,
+      onEntered: focusFirstInput,
+      mainAction: (
+        <form.AppForm>
+          <form.SubmitButton dataTest="product-item-drawer-submit">
```
domenicofalco: following current pattern `product-item-drawer-submit` should be an exportable constante for the tests

src/pages/catalog/drawers/productItemFilter/useProductItemFilterDrawer.tsx
```diff
@@ -0,0 +1,66 @@
...
+    resetForm()
+
+    drawer.open({
+      title: translate('text_17836220307039rf790f045t'),
+      form: { id: PRODUCT_ITEM_FILTER_FORM_ID, submit: form.handleSubmit },
+      closeOnSubmitSuccess: false,
+      onEntered: focusFirstInput,
+      mainAction: (
+        <form.AppForm>
+          <form.SubmitButton dataTest="product-item-filter-drawer-submit">
```
domenicofalco: same here

src/pages/catalog/ProductCatalog.tsx
```diff
@@ -84,6 +90,44 @@ const ProductCatalog = () => {
...
+          items: [
+            {
+              type: 'dropdown',
+              label: translate('text_1742230191029lznwj3y41nb'),
+              dataTest: 'product-catalog-create',
+              items: [
+                {
+                  label: translate('text_1783622030703h5vhmp73muk'),
+                  hidden: !hasPermissions(['productsCreate']),
+                  dataTest: 'create-product',
```
domenicofalco: are those dataTest really needed?

src/pages/catalog/details/ProductDetailsOverview.tsx
```diff
@@ -0,0 +1,117 @@
...
+    id
+    name
+    code
+    description
+    invoiceDisplayName
+    ...ProductForProductDrawer
+  }
+
+  query getProductForDetailsOverview($id: ID!) {
+    product(id: $id) {
```
domenicofalco, line 31: it seems that `product` query is also being performed by the parent component `ProductDetails`. Apollo handles it nicely by looking at its cache first but on a cold start it fires both queries in parallel I think. I am wondering if there a better way to avoid this

src/pages/catalog/drawers/product/ProductDrawerContent.tsx
ansmonjol: Done — added a dedicated `data-drawer-content` attribute (`BASE_DRAWER_CONTENT_ATTR` in `drawers/const.ts`) on the scroll container, and the reset-scroll now selects on that. `data-test` stays reserved for integration tests. (639667bcd)

src/pages/catalog/drawers/productItem/useProductItemDrawer.tsx
ansmonjol: Done — extracted `PRODUCT_ITEM_DRAWER_SUBMIT_TEST_ID` and applied it to the SubmitButton. (421738645)

src/pages/catalog/drawers/productItemFilter/useProductItemFilterDrawer.tsx
ansmonjol: Done — extracted `PRODUCT_ITEM_FILTER_DRAWER_SUBMIT_TEST_ID` and applied it to the SubmitButton. (421738645)

src/pages/catalog/ProductCatalog.tsx
ansmonjol: These are used by the jest suites: `ProductCatalog.test.tsx` asserts the create dropdown opens (`product-catalog-create`) and each item renders + routes to the right drawer (`create-product` / `create-product-item` / `create-product-item-filter`), and `ProductsList.test.tsx` uses `create-product` for the empty-state CTA — so they are needed. I can extract them into exported constants too (like the submit ids above) if you prefer that over inline strings.

ansmonjol: Done — extracted `PRODUCT_CATALOG_CREATE_TEST_ID` / `CREATE_PRODUCT_TEST_ID` / `CREATE_PRODUCT_ITEM_TEST_ID` / `CREATE_PRODUCT_ITEM_FILTER_TEST_ID` and switched the test to import them. (23e35caa9)

src/pages/catalog/details/ProductDetailsOverview.tsx
ansmonjol: Checked how the app handles this elsewhere: it is the established detail-page pattern. `SubscriptionDetails` passes only the id to each tab and every tab self-queries — e.g. `SubscriptionDetailsV2Overview({ subscriptionId })` runs `getSubscriptionForDetailsV2Overview` while the parent runs `getSubscriptionForDetails` — so the same parent+child fetch exists there.

The child self-queries on purpose: `MainHeader` snapshots the tab `content` and only re-pushes it when the serializable config or `snapshotKey` changes. Passing the loaded `product` down as a prop would freeze in that snapshot and go stale after an edit unless `snapshotKey` enumerated every displayed field (name/code/description/invoiceDisplayName).

On cold start the two queries fire in parallel (not a waterfall, so no added latency), overlap in fields, and both normalize into the same `Product:{id}`; warm navigation reads from cache. So I have kept it as-is to stay consistent with the pattern — happy to revisit with a shared lift-and-pass (or Apollo `useFragment`) approach if we want to standardise it across all detail pages.

src/core/form/existingCodeError.ts
ansmonjol: Documented it in the Formik->TanStack migration skill (`.agents/skills/migrate-formik-to-tanstack/SKILL.md`, Step 2.6 next to `NameAndCodeGroup`) rather than CLAUDE.md, since it is a form-migration pattern. (b162671c9)

PR #3899, opened 2026-07-13
feat(catalog): Product items list, drawer and details

ansmonjol:
Fixes BIL-257

Implements the **Product Item** area of the Catalog & pricing section, mirroring the shipped Product flow.

## Included
- Product items **list** (Product items tab): search + two filters (Product, Type), a filters-count ("# items filters") column, numbered pagination, row edit/delete, deep-linkable page/filters.
- Create/Edit **drawer** reusing the shared "Create more" system. Item type = Usage reveals the Billable metric selector. Item type, attached product and billable metric are create-only (disabled on edit); the code locks when the item is attached to a plan/subscription.
- Product item **details page** (new route `/product-catalog/product-items/:productItemId/:tab`, tabs: Overview / Rate cards / Item filters / Plans / Activity logs). Overview shows the supported fields; the other tabs are placeholder stubs.
- Product **details -> Product items tab**: capped preview list + "View all N product items" deep-linking to the list pre-filtered by that product.

## Backend notes
- The schema evolved mid-PR: `Query.productItems` now takes `productIds: [ID!]` (was a single `productId`) plus `withoutProduct: Boolean`, and `ProductItem` gained `filtersCount`. Queries were migrated to `productIds`, and the **"# items filters" column is now included** (binds `filtersCount`).
- Follow-up (now backend-supported, not yet built): the Product and Type filters are still **single-select** and the **"Not defined"** (no-product) option is not wired, though `productIds` (multi) + `withoutProduct` now allow both.
- Grouping settings (pricing / presentation group keys) still have no schema field -> omitted from the drawer and overview.
- `UpdateProductItemInput` only accepts name/code/description/invoiceDisplayName -> item type, product and billable metric are immutable on edit.


src/components/designSystem/Filters/filtersElements/FiltersItemProductItemProduct.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export const FiltersItemProductItemProduct = ({
+  value,
+  setFilterValue,
+}: FiltersItemProductItemProductProps) => {
+  const { translate } = useInternationalization()
+  const { displayInDialog } = useFilters()
+
+  const { data } = useGetProductsForFilterItemProductQuery({
+    variables: { page: 1, limit: 500 },
```
domenicofalco, line 40: should this use pagination instead?

src/pages/catalog/ProductItemsList.tsx
```diff
@@ -1,9 +1,189 @@
...
+  )
+
+  // network-only: tabs are route-based so this component remounts on tab switch
+  // and `?page` is dropped; a cache-first read would flash the previously viewed
+  // page before the page-1 refetch.
+  const [getProductItems, { data, error, loading, variables }] = useProductItemsLazyQuery({
+    variables: { limit: DEFAULT_PAGE_SIZE, page, ...filtersForProductItemsQuery },
+    notifyOnNetworkStatusChange: true,
+    fetchPolicy: 'network-only',
+    nextFetchPolicy: 'network-only',
```
domenicofalco, line 99: is `nextFetchPolicy` needed here?

domenicofalco (approved):

AllanMichay (approved):

src/components/designSystem/Filters/filtersElements/FiltersItemProductItemProduct.tsx
ansmonjol: `MultipleComboBox` has no `searchQuery` prop, so multi-select filters cannot paginate/search server-side. The convention is to bulk-load `limit: 500` and filter client-side, same as `FiltersItemMultipleCustomers`. Server-side pagination is only available on the single-select `ComboBox` (e.g. `FiltersItemPlanCode` via `searchQuery`). Keeping as-is for consistency.

src/pages/catalog/ProductItemsList.tsx
ansmonjol: Yes. `fetchPolicy` only governs the first execution; re-executions (pagination, debounced search) otherwise fall back to the global default `nextFetchPolicy: cache-first` (`init.ts`). Because `createSinglePageFieldPolicy` excludes `page` from the cache key (one page per filter combo, replace merge), a cache-first re-read on page change would paint the previously-viewed page for a frame before the network resolves. `network-only` keeps pagination on a clean skeleton then the correct page. Matches `ProductsList` and the other numbered lists, and is pinned by `ProductItemsList.test.tsx`.

PR #3906, opened 2026-07-14
refactor(wallet form): migrate wallet form to TanStack

domenicofalco:
## Context

The wallet create/edit form was one of the last customer forms still running on the
legacy form stack (Formik + Yup). The subscription and customer forms already moved to
the new architecture (`useAppForm` + Zod + `withForm` sections), which is a prerequisite
for the upcoming drawer UX and the multi-connection customer UI.

This PR is a **pure Formik→TanStack migration: the UI and behaviour are byte-for-byte
identical** — only the plumbing changes. It covers both **create and edit** (same
component) at `/customer/:customerId/wallet/create` and `/customer/:customerId/wallet/:walletId`.

Dive-in: [FE] Wallet, Top-up & One-off form migration to Tanstack Form (Notion)

## Description

Built up through 7 individually-reviewed PRs merged into this feature branch:

- **ING-379** — Zod schema (`formInitialization/validationSchema.ts`, 1:1 parity with the
  old Yup schema, verified via a differential audit on 29 scenarios), API↔form mappers
  (`mappers/`), and the missing `CurrencyPickerFieldForTanstack` wrapper
- **ING-380** — core swap: `useFormik` → `useAppForm` (+ `<form>` wrapper,
  `form.SubmitButton`, `FormLoadingSkeleton`) and `SettingsSection` → `withForm`
- **ING-381** — `ScopeSection` (appliesTo fee types + billable metrics, controlled
  ComboBox + Chips)
- **ING-382 / ING-426** — recurring top-up rule split by axis: method (Fixed/Target) and
  trigger (Interval/Threshold), with the Formik cascading resets preserved
- **ING-383** — recurring `transactionMetadata[]` rows + `PaymentMethodsInvoiceSettings`
  (wallet + rule level, via a structural `{ values, setFieldValue }` adapter)
- **ING-384** — Yup schema removal (`topUpAmountError` kept as a shared util for the
  top-up form), ~50 new Jest tests (schema, mappers, all three sections, CreateWallet
  integration, field wrappers)

Fixes ING-377

PR #3917, opened 2026-07-16
refactor(one-off invoice form): migrate one-off invoice form to TanStack

domenicofalco:
## Context

Part of the TanStack form migration (dive-in linked in ING-378). The one-off invoice
create page was one of the last customer forms still on Formik + Yup. Following the
wallet migration (#3906), this moves it to `useAppForm` + Zod with the UI unchanged.

## Description

- **Zod parity schema** (`createInvoice/formInitialization/validationSchema.ts`),
  differential-audited against the old Yup schema through formik's own
  `validateYupSchema` path: 0 divergences across 32 scenarios (incl. `''`-as-absent
  semantics). One deliberate UX deviation: submitting with no items now shows
  "You must add at least one item" under the *Add an item* button (submit button
  enabled upfront, per the TanStack convention) instead of a silently disabled button.
- **API↔form mappers** extracted (`mapFromApiToForm` / `mapFormToCreateInput`).
  Defaults are recomputed inline on every render — TanStack re-seeds untouched forms
  on deep-change, replacing `enableReinitialize`. The regenerate-from-voided route now
  waits for the preview before showing the form (TanStack won't re-seed after touch).
- **Core swap** in `CreateInvoice.tsx` (`useAppForm` + `revalidateLogic`, native
  `<form>` wrapper, void-before-create flow preserved) and **FeesSection extracted**
  to a `withForm` section (`fees` via `mode="array"`); the four per-fee dialogs are
  untouched, only their callbacks are rewired to `form.setFieldValue`.
- **Wallet fix ridden along**: validator errors live directly on `errorMap.onDynamic`
  (the `.fields` sub-shape is only for manual `setErrorMap`) — fixes the recurring
  top-up accordion validity icon never showing invalid, + migration skill updated.
- **Tests**: 97 across 5 suites (schema, mappers, FeesSection, page, TopUpSection).

Behavior deltas shared with the wallet migration (not regressions): validation
surfaces at the first submit attempt then live; Enter submits; the dirty gate on
submit is dropped.

Fixes ING-378


AllanMichay (approved):
lgtm

PR #3924, opened 2026-07-16
refactor(wallet top-up form): migrate wallet top-up form to TanStack

domenicofalco:
## Context

Final form of the TanStack migration (dive-in linked in ING-392). The wallet top-up
page was the last customer form still on Formik + Yup. Follows the wallet (#3906) and
one-off invoice migrations, with the UI unchanged.

## Description

- **Zod parity schema** (`wallet/topUp/formInitialization/validationSchema.ts`) —
  differential-audited against the old Yup schema through formik's own
  `validateYupSchema` path: 0 divergences across 32 scenarios. The schema is a
  **factory** taking the target wallet's `{rateAmount, min, max, currency}`: the
  bounds live on the wallet, not in the form values (the old Yup read them by
  closure). Reuses the shared `topUpAmountError` and `zodMetadataSchema`.
- **API↔form mappers** extracted; defaults recomputed inline — this also fixes the
  `invoiceRequiresSuccessfulPayment` seed on deep-links (the old form captured it
  once at mount, so it only worked when the wallet was already in the Apollo cache).
- **Core swap** in `CreateWalletTopUp.tsx`: `useAppForm` + `revalidateLogic`, native
  `<form>` wrapper, void-before-top-up flow preserved (incl. refetchQueries);
  metadata rows use the wallet TopUpSection `AppField` pattern; payment settings via
  the `PaymentMethodsForm` adapter.
- **Visible feedback added** (same convention as the one-off form): submitting with
  no credit amount now shows "Field is required" under the visible credits field
  (the old form silently disabled the button). The live paid top-up bounds label is
  unchanged. `TopUpAmountError` enum is now exported from `wallet/form.ts`.
- **Tests**: CreateWalletTopUp 20 (4 new + 2 translation-key selectors cleaned),
  schema 20, mappers 12 — 156 green across the wallet suites.

Behavior deltas shared with the previous migrations (not regressions): submit button
enabled upfront (gates after the first attempt); Enter submits.

Fixes ING-392


src/pages/wallet/CreateWalletTopUp.tsx
```diff
@@ -161,59 +164,28 @@ const CreateWalletTopUp = () => {
...
-            return false
-          }
-
-          return !isNaN(Number(paidCredits)) || !isNaN(Number(grantedCredits))
-        },
+  // The bounds live on the target wallet, not in the form values — rebuild
+  // the schema when the wallet resolves.
+  const validationSchema = useMemo(
+    () =>
+      topUpFormValidationSchema({
```
AllanMichay: Should we rename to `getTopUpFormValidationSchema`?

AllanMichay (approved):

AllanMichay (approved):

PR #3939, opened 2026-07-17
feat(auth): add Microsoft Entra ID (OIDC) SSO flows

mikeh-lago:
Mirrors the Okta SSO implementation: login page, OAuth callback, invitation flow, authentication settings (add, edit, delete dialogs and details page) with premium gating, routes, error mapping and translations. Entra specifics: tenant ID instead of organization name, optional host defaulting to login.microsoftonline.com.


src/pages/settings/teamAndSecurity/authentication/SSOAuthenticationDetails.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  const onDeleteCallback = () => {
+    navigate(AUTHENTICATION_ROUTE)
+  }
+
+  const onEditCallback = () => {
+    refetch()
+  }
+
+  if (!integration) {
```
ansmonjol: **Redirects while the query is still loading.**

`useGetEntraIdIntegrationQuery` / `useGetOktaIntegrationQuery` resolve `Query.integration(id:)`, which has no cache field policy in `src/core/apolloClient/cache.ts`. First render is therefore always `loading: true, data: undefined`, so `integration` is null and we navigate straight back to the list. Deep-linking `/settings/authentication/entra/:id` or hard-refreshing the page bounces the user out.

Same reason makes the `loading ?` skeleton branch on L131-L134 unreachable.

This is carried over from `OktaAuthenticationDetails`, but the shared component now spreads it to two pages, so worth fixing here. Also, navigating during render triggers React's "cannot update a component while rendering a different component" warning - better in an effect:

```tsx
useEffect(() => {
  if (!loading && !integration) navigate(AUTHENTICATION_ROUTE)
}, [loading, integration, navigate])

if (loading) return <skeletons />
if (!integration) return null
```

src/pages/settings/teamAndSecurity/authentication/__tests__/EntraIdAuthenticationDetails.test.tsx
```diff
@@ -0,0 +1,389 @@
...
+    ]
+
+    await prepare({ mocks: emptyMocks })
+
+    await waitFor(() => {
+      expect(mockNavigateFn).toHaveBeenCalled()
+    })
+  })
+
+  it('navigates away when no cached integration on initial render', async () => {
```
ansmonjol: This test and `renders integration details after loading` (L93) assert opposite behaviors, and both pass - `mockNavigateFn` is a `jest.fn()`, so nothing actually unmounts and the component keeps rendering after the "redirect".

That means the suite cannot catch the loading-redirect issue flagged in `SSOAuthenticationDetails.tsx`. Once that is fixed, this test should assert the opposite: no navigation while `loading` is true, navigation only once the query resolves empty (which the `integration: null` test on L169 already covers).

src/pages/auth/EntraIdAuthCallback.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+gql`
+  mutation entraIdLoginUser($input: EntraIdLoginInput!) {
+    entraIdLogin(input: $input) {
+      token
+    }
+  }
+`
+
+const EntraIdAuthCallback = () => {
```
ansmonjol, line 17: This file is a verbatim clone of `OktaAuthCallback.tsx` - same structure, same control flow, same comments, only the provider names differ. The PR extracted the login page, the details page, the add dialog and the delete dialog; the callback is the one place that got copy-pasted instead.

Suggest a `useSSOAuthCallback({ loginMutation, userinfoErrorCode, notAuthorizedCode, providerLoginRoute, codeParam, stateParam })` hook that both callbacks call, so a fix to the StrictMode guard or the redirect race only has to be made once.

src/pages/auth/EntraIdAuthCallback.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const client = useApolloClient()
+  const hasRun = useRef(false)
+  const [entraIdLoginUser] = useEntraIdLoginUserMutation({
+    context: { silentErrorCodes: [LagoApiError.UnprocessableEntity] },
+    fetchPolicy: 'network-only',
+  })
+
+  const [searchParams] = useSearchParams()
+  const code = searchParams.get('code') || ''
+  const state = JSON.parse(searchParams.get('state') || '{}')
```
ansmonjol: Nit (same as the Okta callback): unguarded `JSON.parse` on a URL param throws during render if the IdP ever echoes back a non-JSON `state`, which blanks the page instead of falling through to the login redirect. Also re-parses on every render. Cheap to wrap in a try/catch while extracting the shared hook.

src/pages/settings/teamAndSecurity/authentication/dialogs/AddSSOIntegrationDialog.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  const [deleteIntegration] = useDestroyIntegrationMutation()
+
+  const form = useAppForm({
+    defaultValues: defaultFormValues,
+    validationLogic: revalidateLogic(),
+    validators: {
+      // TanStack detects standard schemas at runtime; the generic form value
+      // type prevents statically relating the provider schema, so it is cast.
+      onDynamic: validationSchema as unknown as never,
```
ansmonjol, line 155: **Type-safety escapes in the shared hook.**

Between `validationSchema: ZodTypeAny` (L51), this `as unknown as never`, `as unknown as TUpdateVars` (L170), `as unknown as TCreateVars` (L173) and `fieldValue as never` (L201), the generic signature no longer relates the form values to the mutation variables. A provider wired to the wrong schema or the wrong mutation compiles fine.

Typing the config as `validationSchema: ZodType<TFormValues>` and inferring `TCreateVars` from `{ input: TFormValues }` would remove most of these. If a couple have to stay for TanStack's runtime schema detection, that is fine - the comment already explains that one.

src/pages/settings/teamAndSecurity/authentication/Authentication.tsx
```diff
@@ -74,12 +115,54 @@ const Authentication = () => {
...
+      method: AuthenticationMethodsEnum.Okta,
+      titleKey: 'text_664c732c264d7eed1c74fda2',
+      subtitleKey: 'text_664c732c264d7eed1c74fda8',
+      editLabelKey: 'text_664c8fa719b5e7ad81c86018',
+      deleteLabelKey: 'text_17522481192202remk2eytrr',
+      route: OKTA_AUTHENTICATION_ROUTE,
+      icon: <Okta />,
+      shouldSee: shouldSeeOktaIntegration,
+      integration: oktaIntegration,
+      openAddDialog: openAddOktaDialog as SSOProviderConfig['openAddDialog'],
```
ansmonjol: These four casts (L145-L146, L158-L159) plus the widened `integration: OktaIntegration | EntraIdIntegration | undefined` on L83 mean nothing stops an Entra integration being handed to `openAddOktaDialog`, or vice versa. It compiles either way, and the runtime failure would be a mangled dialog rather than a type error.

Making `SSOProviderConfig` generic over the integration type (or a discriminated union keyed on `method`) would let the array stay as-is while keeping each entry internally consistent. These are the casts I would most want gone, since this is the one place the two providers get paired up by hand.

src/pages/settings/teamAndSecurity/authentication/EntraIdAuthenticationDetails.tsx
```diff
@@ -0,0 +1,109 @@
...
+      getDetailRows={(currentIntegration) => [
+        {
+          icon: 'globe',
+          labelKey: 'text_1784307344255m1d8phj5f9r',
+          value: currentIntegration.domain,
+        },
+        {
+          icon: 'globe',
+          labelKey: 'text_1784307344255fan2blwpos6',
+          value: currentIntegration.host || 'N/A',
```
ansmonjol: The PR description says host is optional and defaults to `login.microsoftonline.com`. If the backend applies that default, showing `N/A` here tells the user something false about their live config.

Either render the default when the field is empty, or have the API return the resolved host.

src/pages/settings/teamAndSecurity/authentication/__tests__/EntraIdAuthenticationDetails.test.tsx
```diff
@@ -0,0 +1,389 @@
...
+      expect(buttonsAfterClick.length).toBeGreaterThan(allButtons.length)
+    })
+
+    // Click the first button in the popper (edit)
+    const buttonsAfterClick = screen.getAllByRole('button')
+    const popperButtons = buttonsAfterClick.filter(
+      (btn) => !allButtons.includes(btn) && btn.textContent,
+    )
+
+    if (popperButtons.length > 0) {
```
ansmonjol: Assertions wrapped in `if (popperButtons.length > 0)` - and the same pattern on L351 and L378 - pass vacuously when the list is empty. If the popper stops rendering entirely, these three tests still go green.

Drop the guard so the test fails when there is nothing to click, and select the target explicitly rather than by position.

src/pages/settings/teamAndSecurity/authentication/__tests__/Authentication.test.tsx
```diff
@@ -594,6 +622,147 @@ describe('Authentication', () => {
...
+    // (EmailPassword + GoogleOauth get theirs immediately, Entra ID only after query resolves;
+    // Okta shows the sparkles icon since it is not part of the premium integrations)
+    await waitFor(() => {
+      expect(screen.getAllByTestId(/dots-horizontal/).length).toBe(3)
+    })
+
+    const selectorButtons = screen.getAllByRole('button').filter((el) => el.tagName === 'DIV')
+
+    // Click the Entra ID selector (4th one)
+    await user.click(selectorButtons[3])
```
ansmonjol: Index-based element picking (`selectorButtons[3]` here and on L699/L724, `dotsIcons[2]` on L755) plus `getAllByRole('button').filter(el => el.tagName === 'DIV')` makes these tests silently retarget a different provider as soon as the selector order changes - which is now a one-line edit to the `ssoProviders` array.

Per `.agents/docs/testing-practices.md`, add exported `*_TEST_ID` constants on the Selector / dots button and query those instead.

Related: `expect(selectors.length).toBeGreaterThanOrEqual(4)` on L631 (and L693/L718) is the entire body of `renders a selector for Entra ID` - it passes even if Entra never renders. Assert on the Entra selector specifically.

src/pages/settings/teamAndSecurity/authentication/dialogs/__tests__/DeleteEntraIdIntegrationDialog.test.tsx
```diff
@@ -0,0 +1,231 @@
...
+            },
+          },
+        },
+      },
+    ]
+
+    await act(() =>
+      render(
+        <NiceModalWrapper>
+          <TestComponent integration={undefined as any} />
```
ansmonjol: `as any` is banned in this repo, and it is not needed here - the `integration` prop on `TestComponent` can just be typed `integration?: typeof defaultIntegration | undefined` and passed `undefined` directly.

src/pages/auth/__tests__/LoginEntraId.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+        await act(async () => {
+          render(
+            <MemoryRouter>
+              <LoginEntraId />
+            </MemoryRouter>,
+          )
+        })
+
+        const emailInput = screen.getByTestId('submit').closest('form')
```
ansmonjol: This ternary has no effect - the form always exists at this point, and both branches resolve to the same single input. It can collapse to the same lookup used on L147 and L217.

More generally, all three cases reach for `document.querySelector('input')` instead of RTL queries. `screen.getByLabelText(...)` (as used in `AddEntraIdDialog.test.tsx`) would be both shorter and resilient to a second input being added to the card.

src/core/router/AuthRoutes.tsx
```diff
@@ -16,10 +16,13 @@ const InvitationInit = lazyLoad(() => import('~/pages/InvitationInit'))
 const GoogleAuthCallback = lazyLoad(() => import('~/pages/auth/GoogleAuthCallback'))
 const LoginOkta = lazyLoad(() => import('~/pages/auth/LoginOkta'))
 const OktaAuthCallback = lazyLoad(() => import('~/pages/auth/OktaAuthCallback'))
+const LoginEntraId = lazyLoad(() => import('~/pages/auth/LoginEntraId'))
+const EntraIdAuthCallback = lazyLoad(() => import('~/pages/auth/EntraIdAuthCallback'))
 
 // ----------- Routes -----------
 export const LOGIN_ROUTE = '/login'
 export const LOGIN_OKTA = `${LOGIN_ROUTE}/okta`
+export const LOGIN_ENTRA_ID_ROUTE = '/auth/entra'
```
ansmonjol: Minor: `LOGIN_OKTA` is `/login/okta` but this is `/auth/entra`, so the two SSO login pages live in different namespaces. `/login/entra` would keep login pages under `/login` and callbacks under `/auth/*/callback`.

src/pages/settings/teamAndSecurity/authentication/dialogs/DeleteEntraIdIntegrationDialog.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const useDeleteEntraIdIntegrationDialog = () => {
+  const { openDeleteSSOIntegrationDialog } = useDeleteSSOIntegrationDialog({
+    integrationTypename: 'EntraIdIntegration',
+    titleKey: 'text_1784307344255lgty3uwoghl',
+    descriptionKey: 'text_17843073442556cjrcl7drw6',
+    successToastKey: 'text_17843073442557u380a217wd',
+    integrationNameKey: 'text_17843073442548zt904xoinv',
+  })
+
+  const openDeleteEntraIdIntegrationDialog = (data: DeleteEntraIdIntegrationDialogData) => {
```
ansmonjol: Minor: this is a one-line pass-through (same in `DeleteOktaIntegrationDialog`). It can be:

```ts
return { openDeleteEntraIdIntegrationDialog: openDeleteSSOIntegrationDialog }
```

The named type alias is still worth keeping for the fragment-typed `integration`.

translations/base.json
```diff
@@ -4482,5 +4482,39 @@
...
+  "text_1784307344255kkmg7664unz": "Type your Entra ID client ID",
+  "text_17843073442551xjnrw1h4bc": "Entra ID client secret",
+  "text_1784307344255ofy9u1w0hqh": "Type your Entra ID client secret",
+  "text_1784307344255tyzraziy4d1": "Entra ID tenant ID",
+  "text_1784307344255xv4zgs56gin": "Type your Entra ID tenant ID",
+  "text_17843073442559h8ul6r7wf1": "Connect to Microsoft Entra ID",
+  "text_1784307344255fc26gfvrmb5": "Edit Microsoft Entra ID connection",
+  "text_17843073442551nurtvrqz3y": "To edit Microsoft Entra ID connection, please edit the following information",
+  "text_1784307344255lgty3uwoghl": "Delete connection to Microsoft Entra ID",
+  "text_1784307344255dltentracon": "Delete Microsoft Entra ID connection",
```
ansmonjol: `text_1784307344255dltentracon` looks hand-written ("dlt entra con") rather than generated. Please regenerate it with `pnpm translations:add` - the repo rule is that keys are never authored by hand.

While in here, a few of the new keys duplicate strings that already exist and are already provider-agnostic:
- L4514 `"Delete connection"` = existing `text_664c732c264d7eed1c74fdb0`
- L4518 `"Identity provider"` = existing `text_664c732c264d7eed1c74fdbd`
- L4515-L4517 the three `{{integration}} integration successfully ...` strings = existing Okta keys `text_664c732c264d7eed1c74fde6` / `fde8` / `fdb4`

Reusing those saves six keys of translation work, and drops the need for the `deleteMenuLabelKey` prop on `SSOAuthenticationDetails` entirely.

ansmonjol (commented):
Nice work on this. Extracting `LoginSSO` / `SSOAuthenticationDetails` / `AddSSOIntegrationDialog` / `DeleteSSOIntegrationDialog` instead of copy-pasting Okta is the right call, and the premium gating, error mapping and route wiring all look consistent with the existing Okta flows.

Comments inline. Summary of what I'd want addressed before merge:

**High**
1. `SSOAuthenticationDetails` redirects to the list while the query is still loading - this affects the Okta page too, and the new test pins the behavior.
2. The new details test asserts both "navigates away on initial render" and "renders details after loading". Both pass because `useNavigate` is mocked, so the suite cannot catch (1).
3. `EntraIdAuthCallback.tsx` is a verbatim clone of `OktaAuthCallback.tsx`. Everything else in this PR got extracted; this one should too.

**Medium**
4. A batch of `as unknown as` / `as never` casts in `AddSSOIntegrationDialog` and `Authentication.tsx` means nothing stops an Entra integration reaching the Okta dialog.
5. Details page shows `N/A` for an empty host, but the PR description says it defaults to `login.microsoftonline.com`.
6. CI here only shows CodeQL. The Okta suites (`LoginOkta`, `OktaAuthCallback`, `OktaAuthenticationDetails`, `AddOktaDialog`, `DeleteOktaIntegrationDialog`, `Authentication`) are the regression net for the refactor - worth confirming they are green.

**Minor**: route naming, one handcrafted translation key, six translation keys duplicating existing generic strings, two pass-through hook wrappers, and the SVG being the generic Microsoft logo rather than the Entra ID mark.

No security or performance concerns: `state` handling is unchanged from the Okta path, the authorize URL is server-supplied, both new pages are lazy-loaded, no barrel imports, no new deps.

src/pages/settings/teamAndSecurity/authentication/SSOAuthenticationDetails.tsx
ansmonjol: Yet, not 100% sure the `if (!integration) return null` is the behaviour expected.
Better double check during QA and maybe a simple `if (!loading && !integration) {` to replace your initial condition would do the job

src/pages/settings/teamAndSecurity/authentication/Authentication.tsx
```diff
@@ -74,12 +121,57 @@ const Authentication = () => {
...
+      method: AuthenticationMethodsEnum.EntraId,
+      titleKey: 'text_17843073442548zt904xoinv',
+      subtitleKey: 'text_1784307344254qdygzl3hxxa',
+      editLabelKey: 'text_1784307344255fc26gfvrmb5',
+      deleteLabelKey: 'text_1785339647668vsn4wyo0j7n',
+      route: ENTRA_ID_AUTHENTICATION_ROUTE,
+      icon: <MicrosoftEntraId />,
+      shouldSee: shouldSeeEntraIdIntegration,
+      integrationId: entraIdIntegration?.id,
+      openAddDialog: (callback) =>
```
ansmonjol, line 167: **The mispairing this fix targets is still not caught - not by types, not by tests.**

I swapped the Entra entry's opener for the Okta one:

```ts
openAddDialog: (callback) => openAddOktaDialog({ integration: oktaIntegration, callback }),
```

All 26 `Authentication.test.tsx` tests pass and `tsc --noEmit` is clean.

The casts are gone, which is the improvement I asked for, but the type relationship went with them. `openAddDialog: (callback?: (id: string) => void) => void` erases the integration type, so the compiler can no longer see that the Entra entry is supposed to close over `entraIdIntegration`. Previously a wrong pairing at least needed a cast to compile; now it needs nothing.

The two new tests that walk this path (`opens Entra ID add dialog when clicking Entra ID without integration`, `shows edit and delete in Entra ID popper when integration exists`) only assert that *a* `[class*="MuiDialog"]` exists, or that pen + trash icons rendered, so neither can tell which provider's dialog opened.

Either close the loop with types (make `SSOProviderConfig` generic over `TIntegration`, or a union keyed on `method`), or pin it with a test that asserts the Entra dialog specifically. `ENTRA_ID_INTEGRATION_SUBMIT_BTN` is already exported, and asserting the domain / tenant prefill would catch both a wrong dialog and a wrong integration in one go.

src/pages/auth/useSSOAuthCallback.ts
```diff
@@ -0,0 +1,128 @@
...
+  userinfoRedirectCode: LagoApiError
+  /** LagoApiError value surfaced in the URL when the login method is not authorized. */
+  notAuthorizedRedirectCode: LagoApiError
+  /** Invitation-form query param names carrying the code/state back to the invite flow. */
+  codeParam: string
+  stateParam: string
+}
+
+const parseState = (raw: string | null): { state?: string; invitationToken?: string } => {
+  try {
```
ansmonjol, line 29: **The try/catch has no test.**

Collapsing it back to the throwing version:

```ts
const parseState = (raw: string | null) => JSON.parse(raw || '{}')
```

passes 20/20 across `EntraIdAuthCallback.test.tsx` and `OktaAuthCallback.test.tsx`. Every case in both suites sets `state: JSON.stringify({ ... })`, so the throwing branch is never reached.

Since the guard now lives in the shared hook, one case covers both providers: a valid `code` plus `state: 'not-json'`, asserting the callback still calls the login mutation with `state: ''` instead of blanking the page. That is exactly the behaviour the guard buys, and it is the sort of thing a later refactor of `parseState` would drop without anyone noticing.

ansmonjol (commented):
Second pass: went back through my comments from yesterday and they are all genuinely handled, not just papered over. Verified on `dc1f7ca1e` - `tsc --noEmit` clean, `eslint` clean, 99/99 across the 9 affected suites. I also mutation-tested the loading-redirect fix (reverted the `!loading &&` guard) and both the Entra and Okta details suites fail on it, so that one is properly pinned.

Two of the fixes are not covered by anything, though - types or tests. Details inline.

mikeh-lago:
2nd pass comments shall now be fixed @ansmonjol

ansmonjol (approved):

PR #3959, opened 2026-07-21
feat(catalog): product item filters

ansmonjol:
Partially fix BIL-257

Adds a full CRUD surface for `ProductItemFilter` in the product catalog, mirroring the existing Product and Product Item flows.

## What's included
- Paginated **Product item filters** list tab (search, empty/error states, single-page cache policy).
- **Filters panel** (Product + Product item). Backend supports a single `productItemId` + `searchTerm`, so the product-item filter maps to a single id (multi-select is lossy) and the product filter is UI-only pending backend.
- **Create/edit drawer** with a billable-metric-filter values editor (`{ billableMetricFilterId, value }`, chips rendered `key: value` joined by AND), a restricted "attached product item" combobox, and the "define at least one filter" alert.
- **Details page**: overview + real activity logs (gated on premium + `auditLogsView`); rate cards + plans are stubs.
- **Nested preview** inside Product Item details, Item filters tab (first 7 + search + view-all deep-link + create-with-prefill, gated on the product item having billable-metric filters).
- **Delete dialog** (CentralizedDialog, cache eviction across both the list and preview watchers).
- Detail route + tabs enum + translations. Permissions and activity-log types already existed.

## Notes
- One isolated `chore(codegen)` commit captures unrelated upstream schema drift so each feature commit's generated-types diff stays clean.


src/components/designSystem/Filters/filtersElements/FiltersItemProductItemFilterProduct.tsx
```diff
@@ -0,0 +1,81 @@
...
+    // every product, matching the Figma reference. It has no query effect on this list.
+    return [
+      { label: translate('text_1784214117868fh6rndi4m75'), value: filterWithoutProductValue },
+      ...productOptions,
+    ]
+  }, [data?.products?.collection, translate])
+
+  const selectedProductsValue = useMemo(
+    () =>
+      (value ?? '')
```
domenicofalco: consider to have such articulate function in a separate utils for better isolation and dedicated unit tests perhaps

src/components/designSystem/Filters/filtersElements/FiltersItemProductItemFilterProductItem.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import {
+  filterDataInlineSeparator,
+  FiltersFormValues,
+  filterWithoutProductItemValue,
+} from '../types'
+import { escapeFilterLabel, unescapeFilterLabel } from '../utils'
+
+gql`
+  query getProductItemsForFilterItemProductItem($page: Int, $limit: Int) {
```
domenicofalco, line 19: Ideally the design system should never be coupled with the data source. Even tho this pattern is repeated for other DS components as well, the design system should be kept agnostic and decoupled from domain business logic. I feel this element should live in the component folder where it's being used if possible, but I understand if that is difficult due to current patterns

src/components/designSystem/Filters/filtersElements/FiltersItemProductItemFilterProductItem.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    // "Not defined" is injected client-side (not returned by the API) and pinned on top,
+    // matching the Figma reference. It has no query effect on this list.
+    return [
+      { label: translate('text_1784214117868fh6rndi4m75'), value: filterWithoutProductItemValue },
+      ...productItemOptions,
+    ]
+  }, [data?.productItems?.collection, translate])
+
+  const selectedProductItemsValue = useMemo(
```
domenicofalco, line 77: this function is duplicated from src/components/designSystem/Filters/filtersElements/FiltersItemProductItemFilterProduct.tsx 

This block also appears in FiltersItemCurrencies, FiltersItemMultipleCustomers, FiltersItemUserIds, FiltersItemBillingEntity and more. Worth extracting a shared helper

src/components/designSystem/Filters/filtersElements/FiltersItemProductItemFilterProductItem.tsx
ansmonjol: yeah sounds like it should just live our of the design system folder yet still in components!

ansmonjol: Agreed in principle - a data-coupled filter element living under `designSystem/` is the wrong layer. Holding off in this PR though, because doing it partially makes things worse: `FiltersPanelItemTypeSwitch` (also in `designSystem/Filters/`) statically imports all ~70 filter elements, and ~40 of them already query GraphQL directly. Moving only these 2 out would invert the dependency (designSystem importing feature code) and split them from their catalog siblings (`FiltersItemProductItemProduct`, `FiltersItemProductItemType`) that stay put.

The clean version is relocating the whole Filters framework (elements + the switch) out of `designSystem/`, or inverting it with a self-registration registry. That is a dedicated refactor, and it is worth prioritizing - I will open a follow-up ticket for it.

src/components/designSystem/Filters/filtersElements/FiltersItemProductItemFilterProduct.tsx
ansmonjol: Extracted into `parseLabeledMultiFilterValue` in `filtersElements/utils.ts`, with dedicated unit tests. Done in the latest commit.

src/components/designSystem/Filters/filtersElements/FiltersItemProductItemFilterProductItem.tsx
ansmonjol: Deduped: the label-parsing block and the on-change join now both go through shared helpers in `filtersElements/utils.ts` (`parseLabeledMultiFilterValue` plus the existing `formatMultiFilterValue`), applied to the 3 catalog product / product-item filters.

The Variant-A filters you listed (Currencies, MultipleCustomers, UserIds, BillingEntity) can adopt the same helper in a follow-up. Left them out here because `FiltersItemCurrencies` skips `unescapeFilterLabel`, so it needs a careful look rather than a blind swap.

PR #3964, opened 2026-07-21
feat(plans): add plan metadata editing to the plan create/edit UI

domenicofalco:
## Context

Plans support custom `metadata` (key–value pairs) through the API, but there was no way to view or edit it in the UI. A prospect (Bosch, via Pixelpublic) uses plan metadata as the source of truth for the display content of their customer-portal pricing pages, and their non-technical PMs need to edit it without a developer.

## Description

**Plan create/edit form** (`CreatePlan`)
- New `PlanMetadataSection` after Entitlements: section header with `+ Add metadata` (inline below the description when empty, top-right when metadata exists), and a `SectionAccordion` card displaying the key-value pairs read-only, with an `…` menu (Edit / Delete).
- Editing happens in a drawer with one key/value input row per pair, per-row delete, `+ Add metadata` to add rows, and Zod validation (required, key uniqueness, max lengths). The drawer always opens prefilled with the existing pairs, since it edits the whole metadata map.

**Plan details view** (`PlanDetailsV2`)
- New `MetadataAccordion` in the Advanced section (same read view + drawer, saved through `useUpdatePlanWithCascade`), and a `Metadata` leaf in the left sidebar.

**Shared `src/components/metadata/`** — metadata is generic (the API `Metadata::ItemMetadata` model has a polymorphic owner), so the reusable pieces live outside `plans/`: `ItemMetadataDrawer` + `ItemMetadataDrawerContent` (owner-specific copy injected via a `description` prop), `MetadataInfo` (read-only rows) and the API limits as constants (max 50 pairs, key ≤ 100 chars, value ≤ 255 chars). Only the plan wiring stays under `plans/`.

**Supporting changes**
- `EditPlan` / `PlanForUpdateWithCascade` fragments now fetch `metadata { key value }`; form defaults initialize it (serializer already passed it through).
- `zodMetadataSchema` accepts an optional `keyMaxLength` (plan limits differ from the customer ones); existing callers unchanged.
- Reused existing translation keys (Metadata, Add metadata, Key/Value, Edit, Delete, errors); one new key for the section description.
- Tests: new `MetadataAccordion` suite (including a regression test asserting the drawer opens prefilled so existing pairs are never overwritten), plus updated `CreatePlan`, `PlanDetailsV2` and `PlanDetailsV2AdvancedSection` suites.

Fixes LAGO-1696


src/components/metadata/ItemMetadataDrawer.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+interface ItemMetadataDrawerProps {
+  // Owner-specific copy shown under the drawer title, e.g.
+  // "Store custom key-value pairs on this plan."
+  description: string
+  onSave: (values: ItemMetadataFormValues) => void | boolean | Promise<void | boolean>
+  onDelete?: () => void
+}
+
+export const ItemMetadataDrawer = forwardRef<ItemMetadataDrawerRef, ItemMetadataDrawerProps>(
```
ansmonjol, line 41: This drawer uses the old drawer approach. Prefer using the new hook instead!

ansmonjol (approved):

src/components/metadata/ItemMetadataDrawer.tsx
domenicofalco: great catch!

domenicofalco: done

PR #3991, opened 2026-07-23
refactor(billing-entity): migrate EditFinalizeZeroAmountInvoiceDialog to FormDialog and TanStack Form

ansmonjol:
## Context

`EditFinalizeZeroAmountInvoiceDialog` was one of the legacy `forwardRef` + `Dialog` + Formik consumers flagged by the `no-restricted-imports` ESLint rule. It's a small combobox-picker dialog shared by two consumers (billing-entity invoice settings and customer settings), so the migration also updates both call sites in the same PR.

## Description

- **`src/components/settings/invoices/EditFinalizeZeroAmountInvoiceDialog.tsx`**
  - Replaced the `forwardRef` + `useImperativeHandle` + `Dialog` + Formik/yup stack with a `useEditFinalizeZeroAmountInvoiceDialog` hook backed by `useFormDialog` + `useAppForm` + Zod (`z.string().min(1)` on `finalizeZeroAmountInvoice`).
  - Data (`entity`, current `finalizeZeroAmountInvoice`) comes in through the open function and is stashed in a `dataRef`; `form.reset()` + `form.setFieldValue()` seeds the form synchronously before the dialog opens.
  - Preserved the Customer vs BillingEntity branching in `onSubmit`: the customer path uses the `FinalizeZeroAmountInvoiceEnum` union (`'finalize'` / `'skip'`), the billing-entity path serializes to a boolean via the `'true'` / `'false'` string values.
  - `handleSubmit` returns `Promise<DialogResult>` using a `successRef` set inside each mutation's `onCompleted` — throws on failure (with `closeOnError: false`) so the dialog stays open; returns `{ reason: 'success' }` on success so it auto-closes.
  - Wrapped `children` in `<div className="p-8">` to match the header/footer gutter (BaseDialog does not auto-pad JSX children).
  - First field is a `ComboBoxField`, so `onEntered` clicks its `MuiInputBase-root` to open the dropdown on enter. Added a dedicated `FINALIZE_ZERO_AMOUNT_INVOICE_INPUT_CLASSNAME` constant in `src/core/constants/form.ts` under the Billing entity group.
  - Exports a stable `EDIT_FINALIZE_ZERO_AMOUNT_INVOICE_FORM_ID` constant used as the dialog's form id.
  - Dropped `EditFinalizeZeroAmountInvoiceDialog` / `EditFinalizeZeroAmountInvoiceDialogRef` exports.

- **`src/pages/settings/BillingEntity/sections/BillingEntityInvoiceSettings.tsx`** and **`src/components/customers/CustomerSettings.tsx`**
  - Dropped the `useRef<EditFinalizeZeroAmountInvoiceDialogRef>` + `<EditFinalizeZeroAmountInvoiceDialog ref={…} />` JSX renders on each consumer.
  - "Edit" buttons now call `openEditFinalizeZeroAmountInvoiceDialog({ entity, finalizeZeroAmountInvoice })` directly.
  - In `CustomerSettings`, the trailing "dialogs" render simplifies to a single element — dropped the now-useless fragment wrapper.

- **`src/components/customers/__tests__/CustomerSettings.test.tsx`**
  - Replaced the `React.forwardRef` stub for `EditFinalizeZeroAmountInvoiceDialog` with the hook shape (`{ openEditFinalizeZeroAmountInvoiceDialog }`) that `CustomerSettings` now consumes. All 33 tests still pass.

## Scope

Strictly scoped to the `no-restricted-imports` warnings for `EditFinalizeZeroAmountInvoiceDialog`. The other three dialogs in the same billing-entity/customer-settings cluster (`EditBillingEntityGracePeriodDialog`, `EditBillingEntityInvoiceIssuingDatePolicyDialog`, `EditCustomerIssuingDatePolicyDialog`) each ship in their own dedicated PRs.

## Test plan

- [ ] Billing entity settings → Invoice settings → "Finalize empty invoices" row → "Edit" opens the dialog with the combobox dropdown auto-opened.
- [ ] Selecting Finalize / Skip and submitting fires `updateBillingEntityFinalizeZeroAmountInvoice` with the correct boolean and displays the success toast. Dialog closes.
- [ ] Customer settings → same row, both the direct "Edit" (Inherit state) and the popper's "Edit" (non-Inherit) open the dialog with the current value preselected. Submitting calls `updateCustomerFinalizeZeroAmountInvoice` with the correct enum value.
- [ ] Submit button stays disabled until a value is picked (validation).
- [ ] Cancel/close does not mutate; the customer popper's "Delete" still opens the delete dialog.
- [ ] `pnpm exec tsc --noEmit`, `pnpm exec eslint` on the touched files — clean.
- [ ] `pnpm test src/components/customers/__tests__/CustomerSettings.test.tsx` — 33/33 passing.

---
_Generated by [Claude Code](https://claude.ai/code/session_0111EcTfY7fC4j1t9oLrSpYW)_

### DIFF
--- a/src/components/customers/CustomerSettings.tsx
+++ b/src/components/customers/CustomerSettings.tsx
@@ -500,9 +500,10 @@ export const CustomerSettings = ({ customerId }: CustomerSettingsProps) => {
                         disabled={loading}
                         variant="inline"
                         onClick={() =>
+                          customer &&
                           openEditFinalizeZeroAmountInvoiceDialog({
                             entity: customer,
-                            finalizeZeroAmountInvoice: customer?.finalizeZeroAmountInvoice,
+                            finalizeZeroAmountInvoice: customer.finalizeZeroAmountInvoice,
                           })
                         }
                       >
@@ -523,10 +524,12 @@ export const CustomerSettings = ({ customerId }: CustomerSettingsProps) => {
                               variant="quaternary"
                               align="left"
                               onClick={() => {
-                                openEditFinalizeZeroAmountInvoiceDialog({
-                                  entity: customer,
-                                  finalizeZeroAmountInvoice: customer?.finalizeZeroAmountInvoice,
-                                })
+                                if (customer) {
+                                  openEditFinalizeZeroAmountInvoiceDialog({
+                                    entity: customer,
+                                    finalizeZeroAmountInvoice: customer.finalizeZeroAmountInvoice,
+                                  })
+                                }
                                 closePopper()
                               }}
                             >
--- a/src/components/customers/CustomerSettings.tsx
+++ b/src/components/customers/CustomerSettings.tsx
@@ -32,10 +32,7 @@ import {
   SettingsListWrapper,
   SettingsPaddedContainer,
 } from '~/components/layouts/Settings'
-import {
-  EditFinalizeZeroAmountInvoiceDialog,
-  EditFinalizeZeroAmountInvoiceDialogRef,
-} from '~/components/settings/invoices/EditFinalizeZeroAmountInvoiceDialog'
+import { useEditFinalizeZeroAmountInvoiceDialog } from '~/components/settings/invoices/EditFinalizeZeroAmountInvoiceDialog'
 import { useEditNetPaymentTermDialog } from '~/components/settings/invoices/EditNetPaymentTermDialog'
 import {
   INVOICE_ISSUING_DATE_ADJUSTMENT_SETTING_KEYS,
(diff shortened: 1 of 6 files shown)


src/components/customers/CustomerSettings.tsx
```diff
@@ -524,7 +523,10 @@ export const CustomerSettings = ({ customerId }: CustomerSettingsProps) => {
                               variant="quaternary"
                               align="left"
                               onClick={() => {
-                                editFinalizeZeroAmountInvoiceDialogRef?.current?.openDialog()
+                                openEditFinalizeZeroAmountInvoiceDialog({
```
AllanMichay: Should we check if customer exist like we did before?

AllanMichay (approved):

PR #3992, opened 2026-07-23
refactor(billing-entity): migrate EditBillingEntity/EditCustomer invoice issuing date policy dialogs to FormDialog

ansmonjol:
## Context

`EditBillingEntityInvoiceIssuingDatePolicyDialog` and `EditCustomerIssuingDatePolicyDialog` were both flagged by the `no-restricted-imports` ESLint rule. They share the same body (two comboboxes + a live alert) via a `forwardRef`-based `EditInvoiceIssuingDatePolicyDialogContentBase`, which was the only reason either dialog was still on the legacy `Dialog` system. Migrating one and leaving the other on the legacy base would leave the shared file in an inconsistent state, so both hooks migrate together in this PR alongside the shared body's rewrite.

## Description

- **`src/components/invoiceIssuingDatePolicy/EditInvoiceIssuingDatePolicyFormContent.tsx`** (new)
  - Introduces a `withForm({...})` renderer that owns the two-combobox + alert body. Extra prop: `gracePeriod` (used by `useIssuingDatePolicy.getIssuingDateInfoForAlert` to compute the alert copy).
  - `useStore` subscribes to the two field values so the info alert stays live as the user picks options — same behavior as the Formik-driven base.
  - Preserves the two test className constants (`EDIT_INVOICE_ISSUING_DATE_POLICY_DIALOG_ANCHOR_COMBOBOX_TEST_CLASSNAME` / `..._ADJUSTMENT_COMBOBOX_TEST_CLASSNAME`).

- **`src/components/invoiceIssuingDatePolicy/EditInvoiceIssuingDatePolicyDialogContentBase.tsx`** and its test — deleted. The tested behavior (form validation, submit wiring, dialog close) now lives inside the two hook-owned dialogs and their integration flows.

- **`src/components/settings/invoices/EditBillingEntityInvoiceIssuingDatePolicyDialog.tsx`**
  - Rewritten as `useEditBillingEntityInvoiceIssuingDatePolicyDialog`, backed by `useFormDialog` + `useAppForm` + Zod (both fields are optional strings). `dataRef` stashes `{ billingEntity }`; `form.reset()` + `form.setFieldValue()` seeds the two fields synchronously before the dialog opens.
  - `handleSubmit` returns `Promise<DialogResult>` with `closeOnError: false` and a `successRef` set inside the mutation's `onCompleted`.
  - Preserves the legacy fallback: empty → `AlignWithFinalizationDate` / `NextPeriodStart` when submitting for a billing entity (matches the Formik path).

- **`src/components/customers/settings/EditCustomerIssuingDatePolicyDialog.tsx`**
  - Same shape, rewritten as `useEditCustomerIssuingDatePolicyDialog`. Customer's mutation input accepts `null` (unlike the billing-entity variant), so empty values submit as `null`.
  - Preserved the customer path's "clear both fields → delete toast" branch, but evaluated on the mutation response fragment (which returns both fields under `billingConfiguration`) instead of Formik values.

- **First-field UX:** both dialogs open with the anchor combobox's dropdown auto-opened via `onEntered` clicking `.<anchor-className> .MuiInputBase-root` (per the migration skill's branch 2b for combobox-first dialogs).

- **`src/pages/settings/BillingEntity/sections/BillingEntityInvoiceSettings.tsx`** and **`src/components/customers/CustomerSettings.tsx`**
  - Dropped the `useRef<*DialogRef>` + `<*Dialog ref={…} />` JSX renders on each consumer. Buttons now call `openEditBillingEntityInvoiceIssuingDatePolicyDialog({ billingEntity })` / `openEditCustomerIssuingDatePolicyDialog({ customer })` directly.
  - `CustomerSettings`'s trailing render collapses to a single element, so the fragment wrapper is removed.

- **`src/components/customers/__tests__/CustomerSettings.test.tsx`**
  - `EditCustomerIssuingDatePolicyDialog` mock swaps the `React.forwardRef` stub for the hook shape (`{ openEditCustomerIssuingDatePolicyDialog }`). All 33 tests still pass.

## Scope

Strictly scoped to the `no-restricted-imports` warnings for `EditBillingEntityInvoiceIssuingDatePolicyDialog`, `EditCustomerIssuingDatePolicyDialog`, and the shared `EditInvoiceIssuingDatePolicyDialogContentBase`. The other two dialogs in the same billing-entity/customer-settings cluster (`EditBillingEntityGracePeriodDialog`, `EditFinalizeZeroAmountInvoiceDialog`) each ship in their own dedicated PRs.

## Test plan

- [ ] Billing entity settings → Invoice settings → "Issuing date policy" row → "Edit" opens the dialog with the anchor combobox dropdown auto-opened and both fields pre-filled from the entity.
- [ ] Customer settings → "Issuing date policy" row → "Add" opens the same dialog with the customer's current values pre-filled (or blank if inheriting).
- [ ] Picking anchor / adjustment options updates the info alert copy live.
- [ ] Billing entity submit: `updateBillingEntityInvoiceIssuingDatePolicy` fires with the two enums (empty → defaults). Success toast displays. Dialog closes.
- [ ] Customer submit: `updateCustomerIssuingDatePolicy` fires with the two enums (empty → null). Success toast keyed correctly for delete-both vs update.
- [ ] Cancel/close does not mutate. Reopening the dialog re-seeds current values.
- [ ] `pnpm exec tsc --noEmit`, `pnpm exec eslint` on the touched files — clean.
- [ ] `pnpm test src/components/customers/__tests__/CustomerSettings.test.tsx` — 33/33 passing.

---
_Generated by [Claude Code](https://claude.ai/code/session_0111EcTfY7fC4j1t9oLrSpYW)_


AllanMichay (approved):
Maybe we can use `/make-tests`. Right now coverage is at 11%

PR #3993, opened 2026-07-23
feat: add action to delete draft invoice

endenis:
## Context

_Part of the "Delete draft invoice" feature._

Draft invoices can now be deleted via the API using the `deleteInvoice` GraphQL mutation (introduced in this [backend PR](https://github.com/getlago/lago-api/pull/5968)).

This PR adds the front-end UI to trigger that deletion.

## Description

Adds a "delete invoice" action, available only on **draft** invoices to users with the `invoicesDelete` permission.
This action is added to 3 places:

- Customer invoices tab => row action menu
- Global invoices list => row action menu
- Invoice detail page => actions dropdown

Selecting the "delete invoice" opens a confirmation dialog.

On confirm, it runs the `deleteInvoice` mutation and shows a success toast.
From the invoice detail page the user is navigated back to the customer's invoices tab.

**Changes:**
- `DeleteInvoiceDialog` - new confirmation dialog + `useDeleteInvoiceDialog` hook (`useDeleteInvoiceMutation`).
- `usePermissionsInvoiceActions` - new `canDelete` (draft + `invoicesDelete`).
- `useInvoiceAuthorizations` - exposes `canDelete` for the detail page.
- Wired the action into the two lists and the detail dropdown.
- Added translation keys for the dialog, toast, menu label.
- ⚠️ Tests were generated using an LLM


src/components/invoices/DeleteInvoiceDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+      actionText: translate('text_17848001862070vhaaoyut3y'),
+      colorVariant: 'danger',
+      cancelOrCloseText: 'cancel',
+      onAction: async () => {
+        const result = await deleteInvoice({
+          variables: { input: { id: invoice?.id || '' } },
+        })
+
+        if (result.data?.deleteInvoice?.id) {
+          client.refetchQueries({
```
ansmonjol: Prefer `evictFromCache` (`src/core/apolloClient/evictFromCache.ts`) over `refetchQueries` here. It's the util built for exactly this case: nice-modal deletes that remove an item from paginated `{ collection, metadata }` lists. Newer delete dialogs already use it (`DeleteFeatureDialog`, `DeleteCustomSectionDialog`).

Why it's better than the refetch:
- No network roundtrip refetching both full lists.
- Actually evicts the deleted `Invoice` entity from the cache. `refetchQueries` leaves it stale.
- Handles the detail-page delete: the invoice detail query is `cache-and-network` and stays retained/cache-watched after the page unmounts. `evictFromCache` nulls the root ref and suppresses that watcher so it doesn't refire into a 404. The current refetch only avoids the 404 by not evicting at all.

Invoice lives in two lists (`invoices` + `customerInvoices`):
```ts
evictFromCache(client, {
  id: result.data.deleteInvoice.id,
  __typename: 'Invoice',
  listFieldName: ['invoices', 'customerInvoices'],
  listQueryDocument: [GetInvoicesListDocument, GetCustomerInvoicesDocument],
})
```

src/components/invoices/DeleteInvoiceDialog.tsx
```diff
@@ -0,0 +1,59 @@
+import { gql, useApolloClient } from '@apollo/client'
+
+import { useCentralizedDialog } from '~/components/dialogs/CentralizedDialog'
+import { addToast } from '~/core/apolloClient'
+import { InvoiceForDeleteInvoiceFragment, useDeleteInvoiceMutation } from '~/generated/graphql'
+import { useInternationalization } from '~/hooks/core/useInternationalization'
+
+gql`
+  fragment InvoiceForDeleteInvoice on Invoice {
```
ansmonjol: This fragment is defined but never spread by the consumers. The lists and the detail page pass their own invoice objects, relying on structural typing. It works today only because the single field needed is `id`, which every invoice always has.

The trap: if someone later adds a field here (e.g. `number` for the toast), they'll naturally update this fragment expecting it to be enforced, but consumers won't be forced to select it. No compile error, the field is just `undefined` at runtime.

Either spread `InvoiceForDeleteInvoiceFragmentDoc` into the list/detail fragments so codegen enforces the selection, or drop the fragment since it's `id`-only.

ansmonjol (commented):
The PR description says it handles the new `invoice_deleted` activity-log type (label + message), but I don't see that in the diff. Was it dropped, or is the description stale?

src/components/invoices/DeleteInvoiceDialog.tsx
endenis: Done, I dropped this fragment 👍

src/components/invoices/DeleteInvoiceDialog.tsx
endenis: Switched to `evictFromCache` 👍 
Thank you

endenis:
> The PR description says it handles the new `invoice_deleted` activity-log type (label + message), but I don't see that in the diff. Was it dropped, or is the description stale?

It was done in a [separate PR](https://github.com/getlago/lago-front/pull/3929) and is already merged. Sorry for the stale entry in PR description

src/components/invoices/DeleteInvoiceDialog.tsx
ansmonjol: well, not entirely sure it's a good idea to drop it. I think it should be kept but spread at the consumer place, the parent that invokes the modal.
Doing so we ensure the passed data correspond exactly to what we use inside the modal + you can change the type in the modal from `invoice: { id: string } | null | undefined` to `invoice: YourFramentName`

endenis: The modal param is now `invoice: InvoiceForDeleteInvoiceFragment`

I used it in the `InvoiceListItem` and `AllInvoiceDetailsForCustomerInvoiceDetails`

alex-additional (approved):

ansmonjol (approved):

src/components/invoices/DeleteInvoiceDialog.tsx
ansmonjol: Awesome!

PR #3999, opened 2026-07-24
feat(quotes): display wallet preview

AllanMichay:
## Context

The `/credits` command shipped the wallet block but left its preview as a placeholder showing only the wallet name; wallet blocks also rendered nothing at all in saved-quote / PDF previews because `buildPreviewEntities` dropped the `walletCredits` slice entirely.

## Description

Adds `buildWalletPreviewData` to turn a wallet payload into a structured multi-row preview and exposes it on `EntityData.wallet`, populated in both the edit-drawer (`fromWallets`) and saved/PDF (`buildPreviewEntities`) paths. The credits block now renders in preview mode as a Name/Billed/Units/Price table (`WalletPreviewTable`) consistent with the plan and coupon tables — a primary row with the wallet name plus an expiration + scope caption, and separate rows for free credits and each recurring top-up, in the customer's locale/currency. Introduces dedicated wallet-owned translation keys across all locales (the customer-contextual-locale translator has no English fallback, so every key must exist in every locale file).

Fixes LAGO-1611
Fixes LAGO-1414


src/components/designSystem/RichTextEditor/CreditsBlock/CreditsBlockView.tsx
```diff
@@ -9,30 +15,43 @@ export const CREDITS_BLOCK_VIEW_EMPTY_TEST_ID = 'credits-block-view-empty'
...
   const blockLabel = translate('text_1783352692386xocpgvrz3na')
   const displayName = entity?.name || blockLabel
 
-  // Preview mode: minimal placeholder, non-interactive (preview table is out of scope).
+  // Preview renders in the customer's locale/currency (matches the discount block).
+  const currency =
+    entity?.wallet?.currency ??
+    customerCurrency ??
+    organization?.defaultCurrency ??
+    CurrencyEnum.Usd
```
domenicofalco: is it correct to have USD as fallback? I am sure this would not happen but still I am not sure if it is correct to have a non-deterministic currency fallback

domenicofalco (approved):

src/components/designSystem/RichTextEditor/CreditsBlock/CreditsBlockView.tsx
AllanMichay: Good catch. You're right that a non-deterministic currency fallback is undesirable — so I've dropped the `customerCurrency ?? organization?.defaultCurrency ?? CurrencyEnum.Usd` chain here entirely.

The credits preview now reads `entity.wallet.currency` directly. The wallet carries its own resolved currency (set from the quote currency when the payload is serialized), and it's the currency the credits are actually denominated in — so it's the correct, deterministic source. In the preview render path `entity.wallet` is always present, so there's no undefined case to guard.

Note the Discount/Pricing blocks keep the customer/org fallback because they have no per-entity currency; the wallet is the one block that does.

PR #4016, opened 2026-07-27
refactor(multi-connection): move the recurring top-up rule form into a drawer

domenicofalco:
## Context

Per the latest designs, the recurring top-up rule no longer renders as an inline Accordion inside the wallet form: the section shows a create CTA (no rule) or a Selector card, and the whole rule form lives in a drawer. The payment & invoicing settings inside the rule become stacked drawers (drawer-in-drawer, natively supported by the drawer stack).

## Description

- New `RecurringRuleDrawer`: hosts the full rule form (method with its reset cascades, credits, trigger conditionals, recap alert, expiration, metadata, invoicing/payment selectors) with a drawer-owned form. Seed on open, commit on save back to `recurringTransactionRules[0]` — cancelling never touches the wallet form, so an aborted create leaves no invalid draft behind.
- Rule validation extracted from the wallet form's monolithic `superRefine` into a shared helper: the parent schema and the drawer schema (`recurringRuleValidationSchema`, wallet bounds passed as context) run the same checks and cannot drift.
- `TopUpSection`: Accordion replaced by the section header + create CTA (premium-gated) or a Selector card (hover actions: edit/delete). The rule section now sits before the invoicing/payment sections.
- Since rule fields are no longer visible at wallet submit, schema errors keyed under `recurringTransactionRules*` (e.g. a committed rule invalidated by a later bounds change) surface as an error caption under the card; wallet-level fields keep their own inline errors.
- New `robot` icon added to the design system for the card.
- Tests: `RecurringRuleDrawer` suite added; `TopUpSection` suite rewritten around the card/CTA/drawer wiring.

No payload change: same `recurringTransactionRules` input, same create/update semantics (`lagoId` passthrough preserved).

Fixes [ING-499](https://linear.app/getlago/issue/ING-499)


src/pages/wallet/components/TopUpSection.tsx
```diff
@@ -158,29 +77,103 @@ export const TopUpSection = withForm({
...
+            <div className="flex flex-col gap-1">
+              <Selector
+                icon="robot"
+                title={translate('text_6657c29c84ad4500ad764ed6')}
+                endContent={
+                  <Button icon="chevron-right-filled" variant="quaternary" tabIndex={-1} />
+                }
+                hoverActions={
+                  <SelectorActions
+                    actions={[
```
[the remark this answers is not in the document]
domenicofalco: Thanks — the "no delete without a pointer" gap is real, but it isn't specific to this card and the fix doesn't belong in this PR.

Two clarifications:

- Editing **is** reachable by keyboard. `Selector` is `role="button"` with `tabIndex={0}` and maps Space/Enter onto a click (`src/components/designSystem/Selector.tsx`), and this card's `onClick` opens the rule drawer. So a keyboard user can focus the card and open the rule to edit it — only the destructive delete is stranded.
- The cause is the shared `hoverActions` slot in the design-system `Selector`, which renders as `hidden … group-hover/selector:flex`. `display:none` descendants aren't focusable, so the same gap already applies to every other consumer: usage charges, commitments, progressive billing, fixed charges, subscription fee, feature entitlements, integrations, and the rich-text credits/pricing blocks — 10 production call sites besides this one. It also affects touch devices, which never hover.

Fixing it properly means adding a `group-focus-within/selector` variant to `Selector` and mirroring it on `endContent`, which changes focus behaviour for all of those call sites. That's a design-system change deserving its own PR and design sign-off rather than riding along in a wallet refactor, so I'm tracking it separately.

— Domenico

src/pages/wallet/components/RecurringRuleDrawer.tsx
```diff
@@ -0,0 +1,612 @@
...
+      openDrawer: (values) => {
+        // Spread over the defaults so a partially-shaped rule (edit mode)
+        // still seeds every field — lagoId rides along untouched.
+        form.reset({ ...DEFAULT_RULES, ...values }, { keepDefaultValues: true })
+        openRecurringRuleDrawer()
+      },
+      closeDrawer: () => {
+        drawer.close()
+      },
+    }))
```
ansmonjol: Can you use the new drawer system?

src/pages/wallet/components/RecurringRuleDrawer.tsx
```diff
@@ -0,0 +1,612 @@
...
+            {rule.method === RecurringTransactionMethodEnum.Fixed && (
+              <>
+                <form.AppField name="paidCredits">
+                  {(field) => (
+                    <field.AmountInputField
+                      currency={currency}
+                      beforeChangeFormatter={['positiveNumber']}
+                      label={translate('text_62e79671d23ae6ff149de944')}
+                      silentError={true}
+                      errorOverride={recurringPaidCreditsError?.label}
```
ansmonjol: If the user submit without selecting a currency, the error shown under the input will be hidden right?
If i'm correct, this is not expected

src/pages/wallet/components/RecurringRuleDrawer.tsx
```diff
@@ -0,0 +1,612 @@
...
+import { useInternationalization } from '~/hooks/core/useInternationalization'
+import { useAppForm, withForm } from '~/hooks/forms/useAppform'
+import { TransactionMetadataGroup } from '~/pages/wallet/components/TransactionMetadataGroup'
+import { topUpAmountError, walletFormErrorCodes } from '~/pages/wallet/form'
+import { recurringRuleValidationSchema } from '~/pages/wallet/formInitialization/validationSchema'
+import { TWalletDataForm, TWalletRecurringRule } from '~/pages/wallet/types'
+
+const RECURRING_RULE_FORM_ID = 'recurring-rule-drawer-form'
+
+export const RECURRING_RULE_DRAWER_SAVE_TEST_ID = 'recurring-rule-drawer-save'
```
ansmonjol: is it used in a test?

src/pages/wallet/components/RecurringRuleDrawer.tsx
```diff
@@ -0,0 +1,612 @@
...
+                description={translate('text_17848881050570gm2uu5e7sz', {
+                  object: translate(VIEW_TYPE_TRANSLATION_KEYS[ViewTypeEnum.WalletRecurringTopUp]),
+                })}
+              />
+              <InvoicingSettingsSelector
+                viewType={ViewTypeEnum.WalletRecurringTopUp}
+                customerId={customerData.customer.id}
+                value={rule.invoiceCustomSection}
+                onChange={(value) => form.setFieldValue('invoiceCustomSection', value)}
+                data-test="rule-invoicing-settings-selector"
```
ansmonjol: I know it was like this before but worth moving it in a const of used in tests. Same for `'rule-payment-settings-selector'`

src/pages/wallet/components/RecurringRuleDrawer.tsx
domenicofalco: Done in bd0219207 — it now uses the hook system.

It was already going through `useFormDrawer`, but wrapped in the older shape: a `forwardRef` component with `useImperativeHandle` that rendered `null` and existed only to hand `openDrawer`/`closeDrawer` back to the parent. That's now `useRecurringRuleDrawer({ customerData, walletCreatedAt, walletValues, onSave }) => { openDrawer }`, modelled on `usePlanSettingsDrawer`. `TopUpSection` drops the `useRef` and the phantom element and just calls the hook, with `onSave` moving from a JSX prop into the hook argument.

`openDrawer` keeps the same semantics — no argument seeds `DEFAULT_RULES` (create), an argument seeds a copy (edit).

While moving it I also dropped `closeDrawer`: it was on the ref interface but no consumer ever called it.

I kept the filename as `RecurringRuleDrawer.tsx` rather than renaming to `useRecurringRuleDrawer.tsx` — the convention elsewhere is a `use*.tsx` file with the content component in its own module, but renaming drags both test files' imports along and this PR is already big. Happy to rename if you'd prefer the full convention.

— Domenico

src/pages/wallet/components/RecurringRuleDrawer.tsx
domenicofalco: Good catch — no, it wasn't. It was referenced only inside its own module (on the `SubmitButton`), and nothing in `src` or `cypress` imported it.

Unexported it in bd0219207, so it's now a plain private const. The save path itself is covered by the drawer suite through the drawer's `form.submit`, which is why the id never needed to be public.

— Domenico

src/pages/wallet/components/RecurringRuleDrawer.tsx
domenicofalco: I have also added CLADE.md instruction on which drawer system to use when migrating or creating new ones

src/pages/wallet/components/RecurringRuleDrawer.tsx
domenicofalco: done

src/pages/wallet/formInitialization/validationSchema.ts
```diff
@@ -61,6 +61,189 @@ const addExpirationIssue = (
...
+
+/**
+ * Standalone schema for the recurring-rule drawer form (flat rule paths).
+ * Wallet-level bounds are frozen at drawer-open time — the drawer is modal,
+ * so they cannot change during the editing session.
+ */
+export const recurringRuleValidationSchema = (bounds: RecurringRuleBoundsCtx) =>
+  z
+    .custom<TWalletRecurringRule>()
+    .superRefine((rule, ctx) => addRecurringRuleIssues(rule, bounds, ctx, (field) => [field]))
```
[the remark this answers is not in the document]
domenicofalco: Fixed in 89451e9a2 — you're right that the destructure is unguarded, though I put the check somewhere else than suggested.

Guarding inside the refinement would only close the drawer path. `addRecurringRuleIssues` has a second caller: `walletFormValidationSchema` does `rules.forEach((rule) => addRecurringRuleIssues(rule, ...))`, where `Array.isArray` validates the container but says nothing about the elements — so a `null` element still reaches the same destructure. The guard now lives inside `addRecurringRuleIssues`, which covers both callers with one check.

It also adds a `required` issue on the rule rather than returning silently, so a malformed rule fails validation instead of passing unnoticed.

For the record it wasn't reachable through the UI today: the drawer's only caller seeds its values from `DEFAULT_RULES` via `form.reset`, so the value is always an object. But the file already states the invariant on the `Array.isArray` guard — "superRefine must NEVER throw (a throwing validator leaves the form stuck in isSubmitting)" — and the sibling entry point wasn't holding up its end of it.

Covered in `validationSchema.test.ts`: `undefined`/`null`/string against the drawer schema (no throw, reports invalid), plus the array-element case, plus a positive case so the guard can't mask real validation.

— Domenico

src/pages/wallet/components/RecurringRuleDrawer.tsx
```diff
@@ -0,0 +1,605 @@
...
+
+    const rule = useStore(form.store, (state) => state.values)
+
+    const { currency, rateAmount, paidTopUpMinAmountCents, paidTopUpMaxAmountCents } = walletValues
+
+    const recurringPaidCreditsError = topUpAmountError({
+      rateAmount,
+      paidCredits: rule.paidCredits || '',
+      paidTopUpMinAmountCents: (paidTopUpMinAmountCents ?? undefined) as string | undefined,
+      paidTopUpMaxAmountCents: (paidTopUpMaxAmountCents ?? undefined) as string | undefined,
```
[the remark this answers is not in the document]
domenicofalco: Fixed in 735857ea4 — good catch, this one was a live bug, not just an inconsistency. I put the normalization one level down from where you suggested, though.

Normalizing at this call site would have fixed the drawer and left three identical holes: `CreateWalletTopUp.tsx:293`, `WalletFreeAndPaidCreditsFields.tsx:178` and `customerPortal/wallet/WalletPage.tsx:106` all pass raw form values too. Only the three *schema* call sites wrap with `prepared`. So the check now lives inside `topUpAmountError`: `''` becomes `undefined` right after the early returns, and all eight callers agree with the schema by construction. The schemas' `prepared` wrapping is redundant now, but harmless, so I left it.

Confirming the user-visible symptom you predicted: clear a wallet's min or max field and the inline error appeared on paid credits while submit went through, because `hasMax` was true and `Number('') === 0`.

One correction for the record — it isn't new here. Pre-refactor `TopUpSection` passed the same `?? undefined` (lines 164-165), so this PR carried the behaviour over unchanged rather than introducing it.

Covered in `form.test.ts`: emptied max, emptied min, both emptied, plus a case asserting an emptied min does not disarm a max that *is* set — the change should only ever remove spurious errors, never suppress real ones. 448 tests across the wallet, wallet-components and customer-portal suites pass.

— Domenico

src/pages/wallet/components/RecurringRuleDrawer.tsx
```diff
@@ -0,0 +1,605 @@
...
+    onSubmit: async ({ value }) => {
+      await onSave(value)
+      drawer.close()
+    },
+  })
+
+  const openDrawer = (values?: TWalletRecurringRule): void => {
+    // Spread over the defaults so a partially-shaped rule (edit mode)
+    // still seeds every field — lagoId rides along untouched.
+    form.reset({ ...DEFAULT_RULES, ...values }, { keepDefaultValues: true })
```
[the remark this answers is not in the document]
domenicofalco: Real observation, but I'm not taking the fix — it would push us the wrong way.

`mapFromFormToApi` deliberately wants `startedAt` sent as `null` when the user didn't pick a date, so the backend anchors the recurrence to the wallet's `createdAt`. Its comment spells it out: "Defaulting to today here would silently move the recurrence anchor on every save." Stamping `now` at open time means the client always sends a date, which is exactly that.

Impact is also smaller than it looks: the page is lazy-loaded, so `DEFAULT_RULES` is evaluated on first navigation to the wallet form, not at app boot.

If we change anything here it's the other direction — `startedAt: null` in `DEFAULT_RULES` — but that leaves the date picker empty, so it's a product call rather than a drive-by fix. Not in this PR.

— Domenico

src/pages/wallet/components/RecurringRuleDrawer.tsx
domenicofalco: that's right and it was inherited by the previous behavior, good catch! I am fixing in a way that if a user tries to submit, there will be and error under the 2 inputs stating that at least one has to be filled.

AllanMichay (approved):

PR #4028, opened 2026-07-29
chore(ci): trigger build workflow smoke test

IxDay:
## Context

Dummy no-op PR to verify the `Lago Internal` build workflow works end-to-end. It dispatches a `front-push-main` event to `getlago/lago-deploy` on every push to `main`, so this PR confirms that path still fires and the downstream build in `lago-deploy` is green.

## Description

- Adds a single trailing newline in `README.md`.
- No functional changes.
- Once merged, the push to `main` fires `.github/workflows/lago-internal.yml`, which sends the repository_dispatch to `lago-deploy` — that's the path we want to confirm.

## Test plan

- [ ] Verify PR CI (linter, tests, cypress, codegen) stays green
- [ ] After merge, check the `Lago Internal` workflow run succeeds
- [ ] Verify the corresponding build in `lago-deploy` is triggered

Safe to revert immediately after the smoke test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,4 +131,3 @@ Machine-specific tweaks go in `.conductor/settings.local.toml` (gitignored). It 
 ## License
 
 Lago is open-source under the GNU Affero General Public License Version 3 (AGPLv3) or any later version.
-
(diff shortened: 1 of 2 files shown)


domenicofalco:
No longer needed

PR #4033, opened 2026-07-29
feat(multi-connections): customer information — connections view

domenicofalco:
## Context

Replaces the scattered connection rows (`CustomerIntegrationRows`) and the flat `CustomerPaymentMethods` section on the customer information tab with a single master-detail "Connections to external apps" section (one connection per type). No backend dependency: writes persist immediately through the existing `updateCustomer` mutation, behind an injected persistence hook that the dedicated connection mutations (ING-530/431/438) will later replace without touching the view (ING-517).

## Description

- **`CustomerConnectionsSection`** (master-detail): category-grouped connection list (Connection + Status columns — Status stays empty until ING-441) · settings panel with Edit → shared connection drawer · payment-methods block scoped to the payment connection (add-method dialog with the provider preselected and locked). First connection auto-selected, `GenericPlaceholder` empty state, connection-scoped toasts.
- **`useCustomerConnectionsPersistence`**: minimal `updateCustomer` input — `{ id, externalId }` + only the payment keys, or the `integrationCustomers` list rebuilt verbatim from the customer fragment. Deliberate: the backend destroys any integration customer missing from a submitted list, and the form mappers resolve slots against org queries (could drop sibling links mid-flight). Payment delete sends explicit nulls (also destroys saved payment methods — the dialog warns).
- **UI updates without reloads**: refresh via standalone `client.query` (silent cache broadcast) instead of refetching the active `getCustomer` observer (which would unmount the section and lose the selection); bounded poll for async-created integration links; `snapshotKey` on the information tab so the MainHeader snapshot re-pushes the fresh content.
- **Shared**: `CustomerConnectionsList` extended (grouping, Status column, selection, denser 16px-logo rows); provider hooks / drawer content / avatar mappings lifted out of `pages/createCustomers` into shared modules + new `useConnectionOptions`; `ExternalAppsAccordion` refactored on top, no behavior change.
- **Removed**: `CustomerIntegrationRows`, `CustomerPaymentMethods`. The hardcoded "Manual payment" row moved to ING-441 — meaningless without the Default concept, and the backend defers the `manual` connection (ING-437).

Out of scope: default flow (ING-441), connection `code` (M2), pagination (M2), dedicated mutations (ING-517).

Fixes ING-434


src/components/customers/connectionsSection/useCustomerConnectionsPersistence.ts
```diff
@@ -0,0 +1,279 @@
...
+          // Only required UpdateCustomerInput field besides id; the backend
+          // applies it solely on editable customers, echoing it back is a no-op
+          externalId: customer.externalId,
+          ...input,
+        },
+      },
+    })
+
+    return !errors?.length
+  }
```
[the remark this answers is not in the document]
domenicofalco: Thanks for the suggestion! This is already covered on the current head — `runUpdate` is gone since 47f682c5, which swapped this hook onto the dedicated per-connection mutations, and its replacement `runMutation` does exactly what you describe:

```ts
try {
  const { data, errors } = await mutate()

  return !errors?.length && !!data && !!getPayload(data)
} catch {
  return false
}
```

So both halves hold: a throw resolves to `false` instead of bubbling out of the delete confirmation, and a resolve without its payload (`getPayload(data)` empty) is treated as a failed write rather than success.

On the optional toast — deliberately not added. Neither mutation declares `silentErrorCodes`, so any non-silenced error already goes through the global `errorLink` (Sentry + generic danger toast); a local toast would double it up. Same reasoning as the equivalent thread on #4061.

Both paths are pinned by tests in `__tests__/useCustomerConnectionsPersistence.test.ts`: *"GIVEN a mutation that throws (network error) → should resolve false instead of bubbling an unhandled rejection"* and *"GIVEN a mutation that resolves without its payload → should treat the write as failed"*.

---

AllanMichay (approved):

src/components/customers/connectionsSection/useCustomerConnectionsPersistence.ts
```diff
@@ -0,0 +1,387 @@
...
+
+  const [createPaymentConnection] = useCreateCustomerPaymentConnectionMutation()
+  const [updatePaymentConnection] = useUpdateCustomerPaymentConnectionMutation()
+  const [destroyPaymentConnection] = useDestroyCustomerPaymentConnectionMutation()
+  const [createIntegrationConnection] = useCreateCustomerIntegrationConnectionMutation()
+  const [updateIntegrationConnection] = useUpdateCustomerIntegrationConnectionMutation()
+  const [destroyIntegrationConnection] = useDestroyCustomerIntegrationConnectionMutation()
+  const [clearPaymentProvider] = useClearCustomerPaymentProviderMutation()
+
+  const refreshCustomer = async (): Promise<void> => {
```
AllanMichay: Since it seems to be linked to integration refetch, should we poll it? A bit like we do in CustomerDetails or is it okay this way?

AllanMichay (approved):

src/components/customers/connectionsSection/ConnectionDetailsPanel.tsx
```diff
@@ -0,0 +1,282 @@
...
+              value: (
+                <ProviderCustomerIdValue
+                  externalId={providerCustomer?.providerCustomerId}
+                  url={
+                    isStripe && providerCustomer?.providerCustomerId
+                      ? buildStripeCustomerUrl(providerCustomer.providerCustomerId)
+                      : undefined
+                  }
+                  missingIdReason={getMissingProviderIdReason({
+                    syncsOnFirstInvoice: false,
```
ansmonjol, line 168: For Cashfree and Flutterwave payment connections this always resolves to the `unavailable` reason and renders "Unable to link this customer to your provider. Review the integration settings and try again". Those two providers are in `PROVIDERS_WITHOUT_MAPPING` (ConnectionDrawerProviderContent) / `PROVIDERS_WITHOUT_EXTERNAL_ID` (CustomerConnectionDrawer), so the drawer never collects an external customer id and the schema never requires one: `providerCustomerId` is always empty and `syncWithProvider` always false. A perfectly healthy connection is therefore reported as broken. Worth adding a fourth reason (or hiding the Provider customer ID row) for providers that have no mapping.

src/components/customers/connectionsSection/CustomerConnectionsSection.tsx
```diff
@@ -0,0 +1,349 @@
...
+    category: ConnectionCategory,
+    values: ConnectionFormValues,
+    { isEdition }: { isEdition: boolean },
+  ) => {
+    const succeeded = await saveConnection(category, values, { isEdition })
+
+    // The drawer only closes once onSave resolves — reject to keep it open
+    // (and its values intact) when the mutation fails
+    if (!succeeded) {
+      throw new Error('Connection save failed')
```
ansmonjol: This rejection is never awaited, so every failed connection save raises an unhandled promise rejection. `BaseDrawer` submits via `onSubmit={(e) => { e.preventDefault(); form.submit() }}` with no `.catch`, and since `closeOnSubmitSuccess: false` that `submit` is TanStack's `handleSubmit`, which re-throws whatever `onSubmit` throws. Result: a console error plus a Sentry capture (default `onunhandledrejection` handler) on each failure, even though the UX is correct. `useCustomerConnectionsPersistence` already avoids this at its own boundary (`runMutation` resolves false instead of bubbling); catching in the drawer's `onSubmit` before `drawer.close()`, or having `onSave` return a status the drawer checks, keeps the drawer open without the rejection.

src/core/constants/integrationPolling.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * The backend creates integration customers asynchronously (perform_later), so
+ * a read right after the mutation may not include a freshly added link yet.
+ * Shared by the customer-details post-edit poll and the connections
```
ansmonjol, line 4: These constants are not used by the connections immediate-save refresh: `useCustomerConnectionsPersistence.refreshCustomer` fires a single `network-only` `getCustomer` with no retry and does not import this module. If integration customers really are created asynchronously as the comment above states, a connection added from the information view can be absent from that one response, and the new row will not appear until a manual reload. Either wire the retry loop into `refreshCustomer` or drop that half of the sentence.

src/components/customerConnections/CustomerConnectionsList.tsx
```diff
@@ -58,9 +81,147 @@ export const CustomerConnectionsList = ({
...
+  }
+
+  const renderBody = () => {
+    if (!grouped) return rows.map(renderRow)
+
+    return GROUP_ORDER.filter((category) => rows.some((row) => row.category === category)).map(
+      (category) => (
+        <div key={category} className="flex flex-col">
+          <div
+            className="border-b border-grey-300 bg-grey-000 p-2"
```
ansmonjol: `grey-000` is defined as `#FFFFFF` in the shared tailwind config, so `bg-grey-000` renders pure white here, identical to the row background, while the `grouped` prop doc describes "grey category header rows" - the group headers end up separated only by their bottom border. If a tint was intended, `bg-grey-100` matches the existing table headers; if white was intended, `bg-white` avoids adding a palette alias to the shared design-system config (this is currently its only usage).

src/components/customers/connectionsSection/useCustomerConnectionsPersistence.ts
domenicofalco: Good question — you were right, it needed it. Applied in a68e9c2f: the refresh now retries with the same interval/attempt budget as the CustomerDetails poll, and stops as soon as the saved link shows up. Payment connections stay a single read since those writes are synchronous.

src/components/customers/connectionsSection/ConnectionDetailsPanel.tsx
domenicofalco: Good catch, thanks! Applied in a68e9c2f — the Provider customer ID row is now dropped entirely for providers without mapping, so a healthy Cashfree/Flutterwave connection no longer reads as broken. Went with hiding the row rather than a fourth reason since the id carries no information for them.

src/components/customers/connectionsSection/CustomerConnectionsSection.tsx
domenicofalco: Confirmed and fixed in a68e9c2f. onSave now returns a boolean and the drawer closes only on true, so a failed save keeps it open without the rejection. Note the same pattern exists on every drawer using closeOnSubmitSuccess: false — catching at the BaseDrawer submit boundary would fix the whole class, but that is shared by all form drawers so it deserves its own PR.

src/core/constants/integrationPolling.ts
domenicofalco: Right, the doc was ahead of the code. Fixed in a68e9c2f by wiring the retry in rather than dropping the sentence — refreshCustomer now imports both constants and reads again until the saved link appears, so a newly added connection no longer waits for a manual reload.

src/components/customerConnections/CustomerConnectionsList.tsx
domenicofalco: Thanks, good spot. White was the intent, so a68e9c2f switches the class to bg-white and drops "grey" from the grouped prop doc.

src/components/customers/connectionsSection/useCustomerConnectionsPersistence.ts
```diff
@@ -0,0 +1,418 @@
...
+      const cleared = await runMutation(
+        () =>
+          clearPaymentProvider({
+            variables: {
+              input: {
+                id: customer.id,
+                externalId: customer.externalId,
+                paymentProvider: null,
+                providerCustomer: null,
+              },
```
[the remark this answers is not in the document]
domenicofalco: Good catch, thanks! Confirmed against the API: `Customers::UpdateService` only nils `payment_provider_code` inside the `old_provider_customer && payment_provider_removed?` branch, and this fallback runs precisely when there is no provider customer — so the code did survive the removal. Fixed in b60dfddc by passing `paymentProviderCode: null` in the same input.

domenicofalco:
**Run Codegen** is still red on `8ec1f85b` for the same reason as before: `paymentProviderCustomers` / `integrationCustomers` don't exist yet on `api`'s `main` (ING-517 hasn't merged there). No fix to port; this resolves once that API branch lands.

---

domenicofalco:
**Run Test E2E** is also red on `8ec1f85b`, same cascade as before: customer-touching specs fail because the API doesn't yet expose `paymentProviderCustomers`/`integrationCustomers` (ING-517 not on `api`'s `main`). No fix to port; resolves once that branch merges.

---

PR #4072, opened 2026-08-05
refactor(customers): Swap connection persistence to the dedicated per-connection mutations

domenicofalco:
## Context

The customer-information master-detail connections view shipped on a bridge persistence strategy: every add/edit/delete went through the generic `updateCustomer` mutation, echoing the full `integrationCustomers` list verbatim on each integration write because the backend destroys any entry missing from a submitted list. All six dedicated per-connection mutations are now merged on the API, so the injected strategy can be swapped — same hook interface, zero UI change.

## Description

- Rewrite `useCustomerConnectionsPersistence` on the six dedicated mutations (create/update/destroy × payment/integration), with the gql documents colocated in the hook. `updateCustomer` disappears from this surface entirely.
- Same-code edits update the link in place by its id; switching a category to another provider destroys the old link then creates the new one (the update mutations expose no provider change); deletes destroy by link id.
- Drop the verbatim `integrationCustomers` list rebuild and its guards — writing a single connection by id removes the sibling-wipe hazard by construction — and the bounded refresh poll, since the dedicated create services run synchronously (the shared polling constants stay for their other consumer).
- `createIntegrationCustomer` targets the org integration by id: the hook now receives the org lists from `useConnectionOptions` and resolves the drawer's connection code to the integration id, aborting with an error toast when the code cannot be resolved.
- The mutations return the connection object rather than the customer, so the silent standalone customer query after each write is now required (it was belt-and-braces before); it also runs after a failed provider switch so a landed destroy-half is reflected. The page's active getCustomer observer is still never refetched.
- Persistence tests rewritten against the six mutations: exact variables per path, destroy-before-create ordering, switch-abort on destroy failure, unresolved-code and missing-link guards.

> Stacked on #4033 (ING-434) — the base flips to main automatically when it merges.

Fixes ING-517

### DIFF
--- a/src/components/customers/connectionsSection/useCustomerConnectionsPersistence.ts
+++ b/src/components/customers/connectionsSection/useCustomerConnectionsPersistence.ts
@@ -11,6 +11,7 @@ import {
   GetCustomerQuery,
   ProviderPaymentMethodsEnum,
   ProviderTypeEnum,
+  useClearCustomerPaymentProviderMutation,
   useCreateCustomerIntegrationConnectionMutation,
   useCreateCustomerPaymentConnectionMutation,
   useDestroyCustomerIntegrationConnectionMutation,
@@ -52,6 +53,15 @@ gql`
 
   mutation destroyCustomerIntegrationConnection($input: DestroyIntegrationCustomerInput!) {
     destroyIntegrationCustomer(input: $input) {
+      id
+    }
+  }
+
+  # Fallback only: removing a payment connection that has no provider-customer
+  # link (nothing for the dedicated destroy to target) is only expressible
+  # through the customer's explicit-null removal semantics
+  mutation clearCustomerPaymentProvider($input: UpdateCustomerInput!) {
+    updateCustomer(input: $input) {
       id
     }
   }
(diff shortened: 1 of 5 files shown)


src/components/customers/connectionsSection/useCustomerConnectionsPersistence.ts
```diff
@@ -258,21 +278,30 @@ export const useCustomerConnectionsPersistence = ({
...
-    const succeeded =
+    const existingId =
       category === ConnectionCategory.Payment
-        ? // Removal semantics: the key must be PRESENT and null. The backend
-          // also clears the code and destroys the customer's payment methods.
-          await runUpdate({ paymentProvider: null, providerCustomer: null })
-        : await runUpdate({
-            integrationCustomers: buildIntegrationCustomersInput(customer, category, null),
-          })
+        ? customer.providerCustomer?.id
```
ansmonjol, line 327: Delete now keys off `customer.providerCustomer?.id`, but a payment connection can exist on `paymentProvider`/`paymentProviderCode` with no `providerCustomer` record: a provider that has no provider-side customer (Cashfree/Flutterwave), or any provider saved with sync off and no external customer id (the create/edit form's `mapFromFormToApi` sends `providerCustomer: null` in that case). For such a connection `existingId` is undefined, so delete aborts with a danger toast and leaves it in place, even though the row is still shown and offered for deletion (`useCustomerConnectionRows` renders it from `paymentProviderCode` alone). The previous `updateCustomer({ paymentProvider: null })` path removed it in every case. Consider falling back to clearing the payment provider when there is no `providerCustomer.id`.

ansmonjol (approved):
Just one comment otherwise we can merge

src/components/customers/connectionsSection/useCustomerConnectionsPersistence.ts
domenicofalco: Good catch, thanks! You are right — the row derives from `paymentProvider`/`paymentProviderCode` alone, so a link-less connection was left undeletable. Applied in 00e2f7624: that case now falls back to clearing the payment provider with explicit nulls via a minimal colocated `updateCustomer` document (the dedicated destroy needs a link id, so there is nothing for it to target there).

PR #4079, opened 2026-08-06
refactor(subscriptions): send purchaseOrderNumber unconditionally again

domenicofalco:
## Context

`Subscriptions::UpdateService` used to reject the whole `updateSubscription`
call as soon as the `purchase_order_number` key was present on a subscription
that is neither pending nor active — it checked key presence, not an actual
value change. Editing the name or the dates of a terminated subscription
therefore failed with `purchase_order_number_not_editable` (405).

The front-end worked around it by omitting the key for those statuses, which
put a per-field authorization rule in the client. The API guard is now
value-aware (getlago/lago-api#6085, merged to main): it rejects only when the
caller actually attempts to change the value. The workaround can go, and the
client goes back to the usual convention — send the full payload, let the API
decide.

## Description

- `useUpdateSubscriptionInformation`: `buildInput` sends
  `purchaseOrderNumber: normalizePurchaseOrderNumber(...)` unconditionally
  again; the status-based conditional spread and the now-unused
  `isSubscriptionPurchaseOrderNumberEditable` import are gone.
- `useAddSubscription`: same in the update branch of the mutation — plain
  `purchaseOrderNumber` instead of the conditional spread. The create branch was
  never guarded and is untouched.
- `isSubscriptionPurchaseOrderNumberEditable` stays in
  `components/purchaseOrder/utils.ts` and keeps being used by
  `SubscriptionInformationFormSection` to disable the field in the UI: the API
  still refuses an actual change on a non-editable subscription, so the form
  must not offer it. Its doc comment is updated to describe the value-aware rule
  rather than the old key-presence one.
- Tests are inverted, not deleted: the terminated/canceled cases in
  `useUpdateSubscriptionInformation.test.tsx` and `useAddSubscription.test.tsx`
  asserted the key was omitted, they now assert it is sent. Since all four
  statuses expect the same payload, each pair of `describe.each` blocks is
  collapsed into a single one covering pending, active, terminated and canceled.
  The explicit `null` case is preserved — that is what clears the column.

The Sentry fingerprinting helpers added by the same earlier PR
(`getGraphQLErrorCode` / `buildGraphQLErrorFingerprint`) are unrelated to this
rule and are left untouched.

Fixes BIL-418

### DIFF
--- a/src/hooks/customer/useAddSubscription.tsx
+++ b/src/hooks/customer/useAddSubscription.tsx
@@ -4,7 +4,6 @@ import { useMemo } from 'react'
 import { generatePath, useSearchParams } from 'react-router-dom'
 
 import { PlanFormInput } from '~/components/plans/types'
-import { isSubscriptionPurchaseOrderNumberEditable } from '~/components/purchaseOrder/PO'
 import { REDIRECTION_ORIGIN_SUBSCRIPTION_USAGE } from '~/components/subscriptions/SubscriptionUsageLifetimeGraph'
 import { addToast, hasDefinedGQLError } from '~/core/apolloClient'
 import { FORM_TYPE_ENUM } from '~/core/constants/form'
@@ -461,13 +460,10 @@ export const useAddSubscription: UseAddSubscription = ({
                   // subscription column, meaning "inherit from customer".
                   billingEntityId: billingEntityId || null,
                   paymentMethod: parsedPaymentMethod,
-                  // Omitted entirely when not editable, otherwise the BE rejects
-                  // the whole update and the rest of the form can no longer be
-                  // edited. When editable the key must stay even as `null` —
-                  // that is the clear mechanism.
-                  ...(isSubscriptionPurchaseOrderNumberEditable(existingSubscription?.status)
-                    ? { purchaseOrderNumber }
-                    : {}),
+                  // Always sent, whatever the status: the BE only rejects an
+                  // actual value change on a non-editable subscription. The key
+                  // must stay even as `null` — that is the clear mechanism.
+                  purchaseOrderNumber,
                   planOverrides,
                 },
               },
--- a/src/hooks/customer/useAddSubscription.tsx
+++ b/src/hooks/customer/useAddSubscription.tsx
@@ -460,9 +460,6 @@ export const useAddSubscription: UseAddSubscription = ({
                   // subscription column, meaning "inherit from customer".
                   billingEntityId: billingEntityId || null,
                   paymentMethod: parsedPaymentMethod,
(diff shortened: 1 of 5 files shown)


src/hooks/customer/useAddSubscription.tsx
```diff
@@ -461,13 +460,10 @@ export const useAddSubscription: UseAddSubscription = ({
...
-                  // Omitted entirely when not editable, otherwise the BE rejects
-                  // the whole update and the rest of the form can no longer be
-                  // edited. When editable the key must stay even as `null` —
-                  // that is the clear mechanism.
-                  ...(isSubscriptionPurchaseOrderNumberEditable(existingSubscription?.status)
-                    ? { purchaseOrderNumber }
-                    : {}),
+                  // Always sent, whatever the status: the BE only rejects an
+                  // actual value change on a non-editable subscription. The key
+                  // must stay even as `null` — that is the clear mechanism.
```
ansmonjol: maybe those can be removed too!

src/hooks/customer/useUpdateSubscriptionInformation.ts
```diff
@@ -24,11 +21,9 @@ export const useUpdateSubscriptionInformation = ({
...
       endingAt: value.endingAt ? DateTime.fromISO(value.endingAt).toUTC().toISO() : null,
-      // Omitted entirely when not editable, otherwise the BE rejects the whole
-      // update and name/dates can no longer be edited. When editable the key
-      // must stay even as `null` — that is the clear mechanism.
-      ...(isSubscriptionPurchaseOrderNumberEditable(subscription?.status)
-        ? { purchaseOrderNumber: normalizePurchaseOrderNumber(value.purchaseOrderNumber) }
-        : {}),
+      // Always sent, whatever the status: the BE only rejects an actual value
+      // change on a non-editable subscription. The key must stay even as `null`
+      // — that is the clear mechanism, `undefined` gets stripped.
```
ansmonjol: same

ansmonjol (approved):

src/hooks/customer/useAddSubscription.tsx
domenicofalco: Good catch, thanks! Removed in 5a16c574f — the null-vs-undefined rule was already documented a few lines above on `billingEntityId`.

src/hooks/customer/useUpdateSubscriptionInformation.ts
domenicofalco: Removed too in 5a16c574f — same reasoning, the `name` field right above already explains the null-on-clear rule.

PR #4081, opened 2026-08-06
feat(agents): add the loop pipeline skills and scripts

domenicofalco:
## Context

Ticket-to-PR work has been driven by a personal set of loop skills kept outside the repo. The team approved adopting it, so the pipeline moves into lago-front where everyone can run it and improve it through review.

## Description

Adds six skills under `.agents/skills` — the `loop-run` orchestrator plus `loop-spec`, `loop-build`, `loop-review`, `loop-revise` and `loop-clean` — and the two scripts they rely on:

- `scripts/loop-iter.sh` keeps the retry budgets on disk and refuses the fourth attempt, so the cap does not depend on the agent remembering its own count.
- `scripts/loop-notify.sh` sends the "stuck, need a human" DM to whoever runs the loop, resolving the recipient from `SLACK_LOOP_USER_ID` or the git email.

Nothing is tied to a specific developer: identity comes from `gh` auth, `git config` and per-developer Slack env vars, and run state lives outside the repo under `$LOOP_STATE_DIR` (default `~/.claude/loop-state`).

Beyond the port, the pipeline gains an escalating retry (each attempt must differ from the last, with an oscillation check against the iteration history), a structured `impediment.md` on every stop, a per-run journal, and a flywheel file where the loop proposes improvements to its own instructions without ever editing them.

`.agents/skills/loop-run/README.md` documents setup, the run-state layout and the team policies the pipeline assumes: it commits and pushes on its own branch, humans merge, `#frontend` is announced only on green CI, and the full jest suite is never run.


AllanMichay (approved):

ansmonjol:
@copilot rebase this PR on origin main

PR #4089, opened 2026-08-07
refactor(feature-flags): remove multi-currency and multi-entity billing gates

domenicofalco:
## Context

Multi-currency and multi-entity billing both reach general availability in the
same release, so both frontend feature flags are removed together. The two
removals are not separable: the legacy fallback components in the customer
overview and credit-notes surfaces only render when *both* flags are off.

`FeatureFlagEnum` and `useOrganizationInfos().hasFeatureFlag` stay in place —
eleven other flags still rely on them. No GraphQL document changed, so no
codegen run is required; the enum members disappear when the backend removal
lands. Frontend and backend must be merged together.

## Description

- Removed every `FeatureFlagEnum.MultiEntityBilling` gate and collapsed the
  always-on branch: `BillingEntityFormPicker` (no longer self-gates or skips
  its query), `SubscriptionInformationFields`, `WalletInformations`,
  `SubscriptionsPage` (billing-entity column and `billingEntityIds` filter
  always available), `CreateInvoice`, `CustomerInformation` (field always
  editable, EU-tax alert always rendered), `CreateSubscription`,
  `useCustomerFilterDefaults` and `CustomerRequestOverduePayment`.
- Removed every `FeatureFlagEnum.MultiCurrency` gate: the currency never locks
  in `BillingAccordion`, the wallet `SettingsSection` always shows the currency
  dropdown, and the `CreateInvoice` currency field is never disabled.
- Deleted the two `@deprecated` legacy fallbacks —
  `CustomerCreditNotesLegacyCard` and `CustomerInvoiceBalancesLegacyCards` —
  together with their specs. `CustomerCreditNotesList` and `CustomerOverview`
  now always render the breakdown, and the `CUSTOMER_OVERVIEW_LEGACY_CARDS`
  test id is gone.
- Dropped the resulting dead `userCurrency` prop chain (`CustomerOverview` →
  `CustomerInvoicesTab` → `useCustomerDetailsHeaderTabs`) and the unused
  `customer` prop on `BillingAccordion`.
- Removed seven translation keys that became unreachable, from `base.json` and
  every locale carrying them.
- Rewrote the affected specs so the enabled path is the default: all
  "flag disabled" blocks are gone and no test mocks either flag.

Note: in `CustomerRequestOverduePayment` the three-way "missing currency /
missing entity / missing both" toast selection is kept. Both scopes remain
independently optional in the URL, so all three messages stay reachable —
behaviour is identical to today's both-flags-on state.

Fixes ING-391


src/pages/CustomerRequestOverduePayment/index.tsx
```diff
@@ -116,18 +114,13 @@ const CustomerRequestOverduePayment: FC = () => {
...
-  // operator back to the invoices tab so they pick a row from the breakdown.
-  const needsCurrencyScope = hasMultiCurrency && !currencyParam
-  const needsEntityScope = hasMultiEntityBilling && !billingEntityIdParam
+  // Guard: both axes must be scoped explicitly. Without scoping, the total
+  // amount would sum across mismatched buckets and the BE would later reject
+  // the mutation (invoices_have_different_currencies / _billing_entities).
+  // Redirect the operator back to the invoices tab so they pick a row from
+  // the breakdown.
+  const needsCurrencyScope = !currencyParam
+  const needsEntityScope = !billingEntityIdParam
```
AllanMichay, line 123: According to Claude, src/hooks/customer/useCustomerDetailsHeaderActions.tsx:74 navigates here without any `billingEntityId` nor `currency`

Check if this does not create any edge case / regression

domenicofalco: Good catch, thanks — you were right, it was a real dead end. That menu item navigated without either scope param, so with the flags gone it could only ever bounce back to the invoices tab with the info toast.

Removed it in 0f802f85d. The Request payment button on the breakdown row stays the single entry point and it passes both `currency` and `billingEntityId`. It also sits right next to the row the operator is acting on, so nothing is lost in discoverability.

ansmonjol (approved):

PR #4090, opened 2026-08-07
refactor(add-ons): migrate the add-on form from Formik to TanStack Form

AllanMichay:
## Context

Part of the Formik → TanStack Form migration project. The add-on creation/edition page was one of the last resource forms still using Formik + Yup, which is deprecated in the codebase.

## Description

Migrates `src/pages/CreateAddOn.tsx` from `useFormik` to `useAppForm`, following the established CreateTax/CreateCoupon conventions:

- Adds a zod schema `src/pages/createAddOn/validationSchema.ts` porting the Yup rules 1:1 (required name/code, required currency, and the `amountCents` display-string refine with the 0.01 minimum).
- Renders name+code via the shared `NameAndCodeGroup` field group (replacing the Formik-coupled `updateNameAndMaybeCode`), the amount + currency block via `field.AmountInputField` / `field.ComboBoxField`, and keeps the description toggle and `TaxesSelectorSection` wiring.
- Wraps the page in a real `<form>` element, uses `form.SubmitButton` (validity via `canSubmit`, no dirty gate in create mode), awaits `onSave`, and migrates the existing-code server error to the two-effect `setFieldMeta` pattern.
- Leaves `useCreateEditAddOn`, `AddOnCodeSnippet`, `TaxesSelectorSection` and the GraphQL layer untouched.
- Adds tests: `createAddOn/__tests__/validationSchema.test.ts` and `__tests__/CreateAddOn.test.tsx` (29 tests).

Fixes LAGO-1154

### DIFF
--- a/src/pages/CreateAddOn.tsx
+++ b/src/pages/CreateAddOn.tsx
@@ -14,7 +14,7 @@ import { FORM_ERRORS_ENUM, SEARCH_TAX_INPUT_FOR_ADD_ON_CLASSNAME } from '~/core/
 import { ADD_ON_DETAILS_ROUTE, ADD_ONS_ROUTE, useNavigate } from '~/core/router'
 import { deserializeAmount } from '~/core/serializers/serializeAmount'
 import { scrollToTop } from '~/core/utils/domUtils'
-import { CreateAddOnInput, CurrencyEnum } from '~/generated/graphql'
+import { CurrencyEnum } from '~/generated/graphql'
 import { useInternationalization } from '~/hooks/core/useInternationalization'
 import { useAppForm } from '~/hooks/forms/useAppform'
 import { useCreateEditAddOn } from '~/hooks/useCreateEditAddOn'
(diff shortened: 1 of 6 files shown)


src/pages/CreateAddOn.tsx
```diff
@@ -216,145 +249,64 @@ const CreateAddOn = () => {
...
+                    </form.SubmitButton>
+                  </form.AppForm>
                 </div>
               </>
             )}
           </div>
         </Main>
         <Side>
-          <AddOnCodeSnippet loading={loading} addOn={formikProps.values} />
+          <AddOnCodeSnippet loading={loading} addOn={formValues as unknown as CreateAddOnInput} />
```
domenicofalco: any chance to fix the double assertion?

domenicofalco (approved):

src/pages/CreateAddOn.tsx
AllanMichay: Good catch, thanks! Fixed in 94284736f — dropped the double cast and loosened the snippet's `amountCents` to `string | number` (it's a display value and serializeAmount already accepts both), so the form values pass through with no cast.

PR #4116, opened 2026-08-12
fix(deps): bump react-router-dom to v7.18.2

ansmonjol:
Fixes Dependabot alerts #250, #251, #252, #253 (CVE-2026-53668, CVE-2026-53666, CVE-2026-53669).

No patched 6.x release exists (`6.30.4` is latest on the v6 line) — fixes only landed in `react-router@7.13.0`/`7.18.0`. Bumped `react-router-dom` to `7.18.2` (root + design-system peer dep), fixed resulting v7 type breaks:

- `test-utils.tsx`: default import of `react-router-dom` no longer exists in v7 (pure ESM) — switched to namespace import.
- Several `generatePath` call sites cast a required path param (`customerId`) to `string | null` — v7 tightened required-param types to disallow `null`. Changed to `as string` (matches the existing `walletId as string` sibling pattern; `?? null`/cast never changed runtime behavior since `customerId` is always defined at these call sites).

No data router (`createBrowserRouter`) or v6 future-flags in use, so migration surface was limited to the above.


AllanMichay (commented):
Bump itself looks good — lockfile diff is minimal and there's no `createBrowserRouter` / `json()` / `defer()` anywhere, so the migration surface really is that small.

Three things before merge:

1. **The `SettingsHomePage` guard doesn't actually fire** — `t30-multi-org-redirect` is red on exactly that case (`expected '/settings/billing-entity/...' to include '/settings/team-and-security'`). Inline comment below.
2. **Two undisclosed behavior changes** — `BillingEntity` and `QuoteDetails` switch their redirect condition from "no data" to "NotFound error". Worth listing in the PR body so it isn't read as a pure dep bump.
3. **Sonar gate** — 57.1% coverage on new code, and the uncovered part is the guard from (1).

The `Run Codegen` failure is unrelated (`ActivityTypeEnum` missing `rate_card_*`).

src/pages/settings/SettingsHomePage.tsx
```diff
@@ -20,6 +21,13 @@ const SettingsHomePage = () => {
       return
     }
 
+    // react-router v7 wraps navigations in startTransition, so this page can stay mounted
+    // (and this query can still resolve) after the user has already navigated elsewhere from
+    // the settings nav. Bail out instead of clobbering that navigation with a stale redirect.
+    if (strippedPathname !== SETTINGS_ROUTE) {
+      return
+    }
```
AllanMichay, line 29: This guard doesn't fire. During v7's `startTransition` the outgoing tree still reads the **old** `useLocation()`, so `strippedPathname` is still `/settings` when the query resolves — the guard passes and the stale redirect wins anyway. `t30-multi-org-redirect.cy.ts` fails on exactly this.

`navigate()` pushes history synchronously and only defers the React state update, so the DOM URL is already correct:

```ts
if (!window.location.pathname.endsWith(SETTINGS_ROUTE)) return
```

Also, the exact `!==` means `/settings/` with a trailing slash slips through.

src/pages/settings/BillingEntity/BillingEntity.tsx
```diff
@@ -1,34 +1,39 @@
...
-    if (!billingEntityLoading && !billingEntity) {
-      navigate(SETTINGS_ROUTE, { replace: true })
-    }
-  }, [billingEntity, billingEntityLoading, navigate])
+  useNotFoundRedirect({
+    error: billingEntityError,
+    loading: billingEntityLoading,
+    redirectTo: SETTINGS_ROUTE,
+    translateKey: 'text_17865284954420o93wb1ssqz',
+  })
```
AllanMichay, line 36: Direction is right (matches the existing `useNotFoundRedirect` adopters), but this changes the redirect condition from "loaded with no data" to "`NotFound` error".

Can you confirm the API actually returns `not_found` for an unknown code? If it returns `{ billingEntity: null }` with no error, the page now renders header + empty body forever — and the new test locks that behavior in. Same question for `QuoteDetails`.

src/test-utils.tsx
```diff
@@ -3,7 +3,8 @@ import { MockedProvider, MockedResponse } from '@apollo/client/testing'
 import { ThemeProvider } from '@mui/material/styles'
 import { configure, render, RenderOptions } from '@testing-library/react'
 import React, { ReactElement, useEffect } from 'react'
-import Router, { BrowserRouter } from 'react-router-dom'
+import * as Router from 'react-router-dom'
+import { BrowserRouter } from 'react-router-dom'
```
AllanMichay, line 7: `jest.spyOn` on a namespace import only works here because `jest-setup.ts` replaces the module with a plain object literal. Drop that mock and the exports become getter-backed and this throws.

The new `QuoteDetails.test.tsx` already uses the durable idiom — `jest.requireMock('react-router-dom').useParams` — worth using here too.

PR #4120, opened 2026-08-12
feat(quotes): drop startDate on subscription amendment quotes

AllanMichay:
## Context

A `subscription_amendment` quote amends an existing subscription, whose start date is already fixed by that subscription — so showing a start-date field is noise, and shipping `startDate` in the payload is a lie.

## Description

For amendment quotes only, the start date is no longer displayed (aside row and pricing-block subscription-settings picker) and the `startDate` key is omitted entirely — not sent as `null` — from both the `updateQuoteVersion` mutation and the plan billing-items payload (`toPlanBillingItems` gained an `omitStartDate` option). This also fixes a related bug: the pricing block detected amendments via `!!subscriptionId`, which went falsy once `billingItems` held a plan and silently re-enabled the start-date picker after the first save; it is replaced by an explicit `isAmendment` flag threaded down from `EditQuote`, and the settings drawer's zod schema is now built per-mode so the hidden, empty field can't block Save. `endDate` and all other order types are untouched.

Fixes LAGO-1814

### DIFF
--- a/src/pages/quotes/editQuote/EditQuoteAside.tsx
+++ b/src/pages/quotes/editQuote/EditQuoteAside.tsx
@@ -109,6 +109,8 @@ const EditQuoteAsideForm = ({
 
   const hasSubscription = !!quote.subscription
   const isOneOff = quote.orderType === OrderTypeEnum.OneOff
+  // The amended subscription owns the start date: it is neither displayed nor sent (LAGO-1814).
+  const isAmendment = quote.orderType === OrderTypeEnum.SubscriptionAmendment
   const versionId = quote.currentVersion.id
 
   const getDefaultValues = (): EditQuoteAsideFormValues => {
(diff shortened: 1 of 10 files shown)


src/pages/quotes/editQuote/EditQuoteAside.tsx
```diff
@@ -120,7 +122,9 @@ const EditQuoteAsideForm = ({
       subscriptionLabel: quote.subscription
         ? `${quote.subscription.plan?.name ?? ''} - ${quote.subscription.externalId}`
         : undefined,
-      startDate: quote.subscription?.subscriptionAt ?? quote.currentVersion.startDate ?? undefined,
+      startDate: isAmendment
+        ? undefined
+        : (quote.subscription?.subscriptionAt ?? quote.currentVersion.startDate ?? undefined),
```
ansmonjol:
```suggestion
      startDate: isAmendment || !quote.subscription?.subscriptionAt ? 
        ? undefined
        :  quote.currentVersion.startDate,
```

Even tho it makes a not condition on `subscriptionAt`, maybe moving to a method makes it more readable?

ansmonjol: Same for another ternary in src/pages/quotes/EditQuote.tsx

ansmonjol (approved):

src/pages/quotes/editQuote/EditQuoteAside.tsx
AllanMichay: Extracted into a `getStartDate()` helper with an early return, in both `EditQuoteAside.tsx` and `EditQuote.tsx` (ffd26ae0b).

One note on the literal suggestion: besides the stray double `?`, `isAmendment || !subscription?.subscriptionAt ? undefined : currentVersion.startDate` inverts the precedence — when `subscriptionAt` *is* set it would return `currentVersion.startDate` instead of `subscriptionAt`, and it drops the start date entirely for a quote that has a version start date but no subscription. So I kept the existing `subscriptionAt ?? currentVersion.startDate ?? undefined` order and only changed the shape:

```ts
const getStartDate = (): string | undefined => {
  if (isAmendment) return undefined

  return quote.subscription?.subscriptionAt ?? quote.currentVersion.startDate ?? undefined
}
```

Say the word if you did want the precedence flipped.

PR #4146, opened 2026-08-17
feat(subscriptions): Offer cancel on incomplete subscriptions

ancorcruz:
## Context

A payment-gated subscription stays `incomplete` while its activation payment is pending. The dashboard offered no way to act on one: every action but copying the id was hidden, so a merchant looking at a subscription stuck waiting on a payment had nothing to click.

## Description

Offer cancellation for incomplete subscriptions, in the subscriptions list and on the subscription details page. Nothing else opens up: there is no settled plan to edit, upgrade or downgrade, and no usage to alert on, so cancelling is the only action added.

The confirmation says what cancelling actually does. Unlike a pending subscription, an incomplete one has a payment in flight, so the dialog warns that Lago will attempt to cancel it and that the payment provider may not allow it, rather than promising the customer will not be charged.

Three places had each decided independently that a pending or incomplete subscription ends in a cancellation rather than a termination — and the details page had decided nothing at all, labelling its action "Terminate subscription" while opening a dialog titled "Cancel subscription?". That rule now has one home, which also settles the pre-existing wording mismatch for pending subscriptions.

## Depends on

Cancelling an incomplete subscription is rejected by the API until getlago/lago-api#6164 ships. This should merge after it.

Part of INT-214

### DIFF
--- a/src/pages/subscriptions/__tests__/SubscriptionDetails.test.tsx
+++ b/src/pages/subscriptions/__tests__/SubscriptionDetails.test.tsx
@@ -12,12 +12,10 @@ import { StatusTypeEnum } from '~/generated/graphql'
 import { render, testMockNavigateFn } from '~/test-utils'
 
 import SubscriptionDetails, {
+  SUBSCRIPTION_DETAILS_CANCEL_TEST_ID,
   SUBSCRIPTION_DETAILS_TERMINATE_TEST_ID,
   SUBSCRIPTION_DETAILS_UPGRADE_DOWNGRADE_TEST_ID,
 } from '../SubscriptionDetails'
-
-const CANCEL_SUBSCRIPTION_KEY = 'text_64a6d736c23125004817627f'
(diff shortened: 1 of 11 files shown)


src/pages/subscriptions/__tests__/SubscriptionDetails.test.tsx
```diff
@@ -16,6 +16,9 @@ import SubscriptionDetails, {
   SUBSCRIPTION_DETAILS_UPGRADE_DOWNGRADE_TEST_ID,
 } from '../SubscriptionDetails'
 
+const CANCEL_SUBSCRIPTION_KEY = 'text_64a6d736c23125004817627f'
+const TERMINATE_SUBSCRIPTION_KEY = 'text_62d904b97e690a881f2b867c'
```
domenicofalco: you shouldn't be using translations keys for testing as they are too fragile. try data-test-id instead

domenicofalco (approved):

PR #4172, opened 2026-08-23
fix(plans): keep the plan form fields required locally

rsempe:
## Context

getlago/lago-api#5779 makes `interval`, `amountCents` and `payInAdvance` optional on `CreatePlanInput`/`UpdatePlanInput` — product-catalog plans carry no plan-level pricing, and a catalog organization manages plans through the existing mutations. `PlanFormInput` derives from that input type, so the fields would go optional across the entire legacy plan form flow (13 type errors), even though that form always sets them.

## Description

Re-require the three fields on `PlanFormInput`, at the type boundary rather than at 13 call sites. The legacy plan form always initializes them, and the product-catalog flow does not use this form.

**This can merge before the API change**: the `Omit`-then-re-add shape compiles against both the current schema (fields required) and the upcoming one (fields optional) — verified in both directions, with the plans/serializers/hooks suites green (1197 tests).

Follow-up to #4152, which handled the output-side nullability of the same fields.

### DIFF
--- a/src/components/plans/types.ts
+++ b/src/components/plans/types.ts
@@ -9,6 +9,7 @@ import {
   EntitlementInput,
   EntitlementPrivilegeInput,
   FixedChargeInput,
+  PlanInterval,
   PrivilegeValueTypeEnum,
   PropertiesInput,
   TaxForPlanAndChargesInPlanFormFragment,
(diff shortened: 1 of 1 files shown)


rsempe:
Superseded by a design change on getlago/lago-api#5779: instead of loosening CreatePlanInput/UpdatePlanInput, the API keeps them strict for legacy and adds dedicated createCatalogPlan/updateCatalogPlan mutations with catalog-only inputs. Front main compiles against the new schema unchanged, so no front change is needed. #4152 (output nullability) remains valid and merged.

PR #4173, opened 2026-08-24
fix(filters): prevent invalid date ranges in date filters

domenicofalco:
## Context

In Invoices and Developer tools > Activity logs, users could select an end date earlier than the start date (for example `08/20/2026` to `08/04/2026`). Nothing constrained the end date against the selected start date, and the panel validation only checked that both bounds were non-empty, so the inverted range was applied, sent to the API, and answered with no results.

The scope was wider than those two surfaces: seven date range filters are rendered by four components, three of which wrote each side of the range blindly. Only `FiltersItemWebhookDate` already clamped the opposite bound.

## Description

Aligns every date range filter on the `webhookDate` behaviour, extracted once instead of copy-pasted a fourth time.

- New `useDateRangeFilterValue` hook, lifted from `FiltersItemWebhookDate`: owns the `from,to` parsing, clamps the opposite bound when a pick would invert the range, exposes the calendar bounds, and always serialises with `.toISO()` (the three buggy components interpolated the `DateTime` object directly). It also writes an empty bound when a date is cleared, instead of persisting an unparsable `Invalid DateTime`.
- `FiltersItemDate`, `FiltersItemIssuingDate`, `FiltersItemLoggedDate` and `FiltersItemWebhookDate` all consume the hook, so there is a single implementation. Per-surface props are untouched (UTC `defaultZone` for `date`/`loggedDate`, org timezone for `webhookDate`, none for `issuingDate`, `disableFuture` where it already was).
- `DatePicker` gains a `maxDate` passthrough mirroring the existing `minDate` one. The start picker is now capped at the selected end date and the end picker floored at the selected start date, so the invalid selection is prevented rather than silently corrected.
- `isValidDateRangeValue` replaces the `/\w+,\w+/` match in the panel schema: it requires both bounds to parse and to be ordered. This is the actual guarantee — **Apply** stays disabled if an inverted range reaches the form from a typed date, a cleared bound, a hand-edited URL, or a future date filter.

`FiltersItemAmount` and the other numeric from/to filters have the same class of bug with numbers; left out as the ticket specifies.

This ships as one coherent PR closing the ticket rather than the split the description sketches.

### Tests

Hook contract (both clamp directions, single bound, cleared bound, unparsable bound), the validation helper, clamp cases on all four components, the calendar bounds wiring for the four of them at once, and the panel keeping **Apply** disabled on a range that cannot be queried.

Two sets of pre-existing tests were rewritten against real behaviour: `FiltersItemWebhookDate.test.tsx` re-implemented the clamp logic inside the test rather than exercising the component, and the `FiltersItemDate`/`IssuingDate`/`LoggedDate` "keeps the other bound" cases picked a `from` after the `to`, encoding the bug.

Fixes ING-592

---
_Generated by [Claude Code](https://claude.ai/code/session_015oSEE2MhXAuuCLx7p34x1Z)_


domenicofalco:
Confirmed and fixed in `ce28a76`.

Reproduced it before changing anything, by capturing the props the pickers receive with the ambient zone set to `America/Los_Angeles` and `defaultZone="UTC"`:

```
fromPicker.defaultZone : UTC
minDate zone / day     : America/Los_Angeles 31
minDate ISO            : 2023-12-31T16:00:00.000-08:00
```

For the value `2024-01-01T00:00:00.000Z` the bound landed on December 31, so the calendars were capped a day off. `UserIdentifier` sets `Settings.defaultZone` app-wide to the organization's timezone, and `DatePicker` only switches it from an effect — and since that effect triggers no re-render of the parent, the wrong bound persisted for as long as the panel stayed open.

`parseBound` now takes the zone explicitly and every call site passes `defaultZone`, which drops the dependency on effect ordering. Behaviour is unchanged where the ambient zone already matched: `issuingDate` passes no zone and still resolves against the ambient one, and webhook logs already ran on the organization's.

Added three regression cases with the ambient zone behind and ahead of UTC, and mutation-checked them — reverting the fix fails all three with the expected wrong values (`31` instead of `1` in Los Angeles, `1` instead of `31` in Tokyo).

One correction to the report's scope: the calendar bounds are the part that was actually wrong in production. Threading the zone through the change handlers is defence in depth rather than an observable bug, since `DatePicker`'s effect always runs before a user can pick a date — the handler case only misbehaves with the picker mocked out.

---

ansmonjol (approved):

PR #4175, opened 2026-08-24
fix(conductor): robust archive teardown hook

ansmonjol:
The archive hook resolved its teardown script from `$CONDUCTOR_WORKSPACE_PATH`, which is asymmetric: Conductor reads `.conductor/settings.toml` from the remote default branch, but resolves the path against the workspace checkout. Any workspace whose branch predates `b148efab9` (the commit that added `scripts/conductor-front-container.sh`) has no copy of the file, so archiving it failed with:

```
zsh: no such file or directory: .../front/davao/scripts/conductor-front-container.sh
```

By archive time the workspace directory can also already be gone, so a re-run never succeeds either (the failed hook for a workspace archived on 2026-07-23 fired again a month later at app startup).

## Changes

`.conductor/settings.toml`
- resolve the script from `$CONDUCTOR_ROOT_PATH` (the main clone, same branch as the settings file), fall back to the workspace copy, then no-op instead of failing.
- `[scripts.run.*]` deliberately keep `$CONDUCTOR_WORKSPACE_PATH` so a branch-local edit to the script is what actually runs.

`scripts/conductor-front-container.sh`
- `down` no longer requires the workspace directory: it works off the generated compose file, which lives outside the workspace. `up`/`shell`/`host` assert the directory exists instead (`require_ws`).
- the compose project is named after `CONDUCTOR_WORKSPACE_NAME` at `up` time, so renaming a workspace made `down` miss the live project and orphan its ~600MB node_modules volume. `down` now also tears down any compose project whose `/app` mount is this workspace. This had leaked 5 orphaned projects / 10 volumes / ~2.5GB locally, since cleaned up.
- a stale or unparseable compose file falls back to removing the container and volumes directly, and docker being unreachable warns rather than failing the archive.

## Notes

- Verified in an isolated fake superproject: rename sweep removes container + volume + compose file (exit 0), unrelated projects untouched, archive with a missing workspace directory exits 0, `up` with a missing directory fails loudly.
- The settings change only takes effect once merged, since the Mac app reads the shared settings file from the remote default branch.

### DIFF
--- a/.conductor/settings.toml
+++ b/.conductor/settings.toml
@@ -14,7 +14,15 @@ delete_branch_on_archive = true
 
 [scripts]
 setup = "pnpm install"
-archive = "$CONDUCTOR_WORKSPACE_PATH/scripts/conductor-front-container.sh down"
+# The archive hook resolves the teardown script from the MAIN clone
+# ($CONDUCTOR_ROOT_PATH), not from the workspace being archived. A workspace
+# whose branch predates the commit that added scripts/conductor-front-container.sh
+# has no such file in its checkout, and by archive time its directory may already
+# be gone, so $CONDUCTOR_WORKSPACE_PATH/scripts/... fails with "no such file or
+# directory". Teardown only needs the generated compose file (kept outside the
+# workspace), so the main clone copy is always the right one. Falls back to the
+# workspace copy, then no-ops, so a missing script never fails the archive.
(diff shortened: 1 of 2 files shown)


ansmonjol:
Pre-review was against `bb0c955`. First point is fixed in `f79b82b`:

- `teardown_san` now removes the project-scoped volumes. It matches on `label=com.docker.compose.project=lago_front_ct_<san>` (which outlives the compose file) and keeps `lago_front_ct_<san>_front_nm_ct_<san>` / `_front_dist_ct_<san>` as literal fallbacks. Verified against a real project on both fallback routes, compose file deleted and compose file unparseable: container and both volumes removed, exit 0. Same commit also stops deleting the compose file before the container removal is confirmed, and drops the `require_ws` guard from `shell`, which never reads `$WS`.

Second point (regression coverage) is a real gap but not something this PR can carry: the repo has no shell test harness at all, no `bats`, no `shellcheck` dev dependency, and no workflow that executes a `.sh` file. Adding one is a tooling change of its own. Flagged for the author to decide separately.

domenicofalco (approved):

PR #4179, opened 2026-08-24
feat(payment-terms): structured payment terms on billing entity and customer

sarkissianraffi:
Replaces the single `net_payment_term` integer with the structured `payment_term` object from [spec #1](https://app.notion.com/p/getlago/Spec-More-Payment-Term-Types-392ef63110d28154b522ddb5518239ae), across the two levels the API supports today.

## Tickets

Covers every front-end ticket of the `[Core] Migration to new payment term` milestone and, because all six types ship at once rather than the parity subset, the `[Feature] Add more payment terms options` milestone too.

| Ticket | Delivered as |
|---|---|
| [BIL-552](https://linear.app/getlago/issue/BIL-552/f1-payment-term-graphql-types-term-type-constants-supported-types-gate) — F1: payment term GraphQL types + constants | `src/core/constants/paymentTerm.ts`. No `SUPPORTED_TERM_TYPES` gate: all six types ship together, so it has nothing to hide |
| [BIL-553](https://linear.app/getlago/issue/BIL-553/f2-computepaymentduedate-getpaymenttermlabel-helpers-unit-tests) — F2: due-date + label helpers | `src/core/utils/paymentTerm.ts` (`paymentTermDueDate`, `buildPaymentTermInput`, `resolvePaymentTerm`) and `src/hooks/usePaymentTerm.ts` for the label copy, which needs `translate` |
| [BIL-554](https://linear.app/getlago/issue/BIL-554/f3-payment-terms-translation-keys) — F3: translation keys | 39 new keys, 18 legacy ones removed, 9/9 locales consistent |
| [BIL-555](https://linear.app/getlago/issue/BIL-555/f4-rework-the-payment-terms-dialog-onto-payment-term-parity-due-on) — F4: rework the dialog | `EditPaymentTermDialog.tsx` + `paymentTerms/PaymentTermFormContent.tsx`. **Superset of its scope** — all six types, not the `due_on_receipt` + `net` subset |
| [BIL-556](https://linear.app/getlago/issue/BIL-556/f5-payment-terms-section-on-billing-entity-customer-settings-reset-to) — F5: both settings surfaces | `BillingEntityInvoiceSettings.tsx`, `CustomerSettings.tsx`, `DeleteCustomerPaymentTermDialog.tsx` |
| [BIL-557](https://linear.app/getlago/issue/BIL-557/f6-cypress-harness-data-driven-payment-terms-spec-both-surfaces) — F6: Cypress harness | `cypress/e2e/10-resources/t100-payment-terms.cy.ts`. Its `data-test` hooks were **not** already in place, so this adds them |
| [BIL-558](https://linear.app/getlago/issue/BIL-558/f7-quote-aside-payment-term-label-off-the-legacy-int) — F7: quote payment term label | `QuotePaymentTermLine` takes the structured term; resolution happens once in `EditQuote.tsx` |
| [BIL-560](https://linear.app/getlago/issue/BIL-560/f9-remove-the-last-netpaymentterm-reads-migration-gate) — F9: drop the last legacy reads | No production reader of `netPaymentTerm` left. `src/generated/graphql.tsx` keeps its own, by design — [BIL-526](https://linear.app/getlago/issue/BIL-526/t20-consolidate-net-payment-term-as-a-permanent-derived-mirror) keeps the alias as a permanent mirror |
| [BIL-561](https://linear.app/getlago/issue/BIL-561/f10-add-end-of-month-payment-term-option) — F10: End of month | `end_of_month` |
| [BIL-562](https://linear.app/getlago/issue/BIL-562/f11-add-net-days-after-end-of-month-us-payment-term-option) — F11: Net days after end of month | `net_end_of_month`. Its own flat enum value, **not** an `anchor` on a shared type as the ticket assumes |
| [BIL-563](https://linear.app/getlago/issue/BIL-563/f12-add-days-end-of-month-eu-payment-term-option) — F12: Days end of month | `days_end_of_month`, likewise its own value, so it is not blocked by F11 |
| [BIL-564](https://linear.app/getlago/issue/BIL-564/f13-add-month-following-invoice-mfi-payment-term-option) — F13: Month following invoice | `day_of_month` + `dayOfMonth` / `monthOffset`. Bounds mirror `PaymentTerms::ValidateService` exactly (1–31, 0–12) |

[BIL-559](https://linear.app/getlago/issue/BIL-559/f8-dunning-email-previews-payment-term-copy-off-the-legacy-int) (F8) is **deliberately left open** — the dunning sentence is deleted rather than reworded, which needs product sign-off. Details below.

Follow-ups this PR does not attempt, on the `[Feature] Add payment term options to more billing items` milestone: [BIL-673](https://linear.app/getlago/issue/BIL-673/f14-subscription-payment-term-override) (subscription, unblocked), [BIL-674](https://linear.app/getlago/issue/BIL-674/f15-one-off-invoice-payment-term-override) and [BIL-675](https://linear.app/getlago/issue/BIL-675/f16-wallet-payment-term-override) (both still blocked on the API).

## What is here

**Billing entity → Invoice settings → Payment terms.** The row renders the resolved term and the dialog edits it across all six types: due on receipt, net, end of month, net days after end of month, days end of month, and month following invoice.

**Customer settings.** Same dialog, plus inheritance. The row shows the effective term marked `(inherit from billing entity)` when the customer has none. The term type list leads with an inherit choice labelled with the value it would inherit, and picking it sends `paymentTerm: null`.

**Shared pieces.** `PaymentTermFormContent` is the one editor every level renders. `core/utils/paymentTerm.ts` mirrors the API's `PaymentTerm#due_date_for`, including the day-of-month clamp and roll-forward, which drives the live due-date preview in the dialog. The two end-of-month conventions take the same `days` and give different dates, so the preview is what tells them apart before saving.

**Legacy alias readers.** `net_payment_term` stays as a two-way alias, but the API returns it as `NULL` for the four types it cannot express. Every remaining reader of the integer would have rendered `-` or wrong copy the moment an account picked one of those types:

- Quote documents now resolve customer → billing entity → due on receipt and format with the shared helper. Quotes carry no term of their own, per the product decision in the BE dive-in, so this is a resolved read-only display.
- The dunning email drops its payment-terms sentence outright (spec #1 §7). With terms set per customer and billing entity, one generic sentence cannot state one truthfully. The API branch already removed it from the mailer. **Open question for product** — see [BIL-559](https://linear.app/getlago/issue/BIL-559/f8-dunning-email-previews-payment-term-copy-off-the-legacy-int); if the copy should come back instead, it is a small follow-up over `formatPaymentTerm` on the resolved term.

**End-to-end.** `cypress/e2e/10-resources/t100-payment-terms.cy.ts` drives every term type off a single `TERM_CASES` table: the billing entity sets each of the six and asserts its settings row, then the customer overrides its billing entity and deletes the override to inherit again. A new term option is a new row in that table, never a new spec file.

## Review fixes

Four defects reported on this PR, plus one found alongside them:

| Where | Defect |
|---|---|
| `usePaymentTerm.ts` | The due-date preview took the browser zone, where `endOf('month')` is `23:59:59.999`, so every End-of-month term previewed as the 1st of the next month anywhere west of Greenwich. It is now anchored to a fixed example issuing date and computed as a calendar date in UTC. Two tests pin it, one of them under a negative-offset zone. |
| `usePaymentTerm.ts` | The inherit option carried `value: ''`, which `ComboBoxField` normalises to `undefined` and `ComboBox` renders as `value={value \|\| null}` — it could never show as selected, and picking it visibly cleared the input. It now carries an `inherit` sentinel mapped back to `null` on submit. |
| `EditPaymentTermDialog.tsx` | An absent term type passed validation, so a billing entity that had never had a term spun its Save button forever with no field error anywhere. `termType` is now required. |
| `EditPaymentTermDialog.tsx` | Same root cause: the customer **Add** flow sent a no-op clear and toasted "successfully deleted". Inheriting a level that carries no term of its own now closes without a mutation. |
| `PaymentTermFormContent.tsx` | The schema's four `message: ''` placeholders are real keys now — Zod v4 substitutes its own untranslated "Invalid input" for an empty message. |

## Codegen

The types were previously generated from a locally composed schema, gitignored as `schema.local.graphql`. [lago-api#6118](https://github.com/getlago/lago-api/pull/6118) is green and mergeable, so `src/generated/graphql.tsx` is now a plain `pnpm codegen` against api `main` merged with that branch, and both the composed file and its `.gitignore` entry are gone.

That refresh also picks up api-`main` schema drift that front `main`'s generated file had not caught up with — hence the one-line `permissionsConst` snapshot update for the new `ContractsUpdate` permission.

## Blocked on the API, in two places

**This PR cannot leave draft until lago-api#6118 merges.** `.github/workflows/codegen.yml` checks lago-api out at its *default* branch for a `pull_request` event, regenerates, and runs `tsc`. Against api `main` that regeneration drops `PaymentTermTypeEnum` and the typecheck fails on everything here. The job is gated on `draft == false`, which is why every check currently reports SKIPPED. Same for the e2e spec: `cy.signup()` hits a real API.

**Two of the four billing-object surfaces still have no GraphQL.** Since this PR was opened the API branch gained `Subscription.paymentTerm` with `paymentTerm` on both subscription inputs, and `Invoice.paymentTerm` (read-only, off `snapshotted_payment_term`). Still absent: `Wallet.paymentTerm` with its wallet inputs ([BIL-522](https://linear.app/getlago/issue/BIL-522/t16-wallet-payment-term-paidcreditservice-resolution-api) / T16) and `CreateInvoiceInput.paymentTerm` ([BIL-521](https://linear.app/getlago/issue/BIL-521/t15-one-off-invoice-payment-term-param) / T15), both unstarted.

Correcting an earlier version of this description: the subscription override is **not** one central edit. `PaymentSettingsDrawer.tsx` is mounted only by `subscriptions/form/PaymentSettingsSection.tsx` and `subscriptions/details-v2/SubscriptionPaymentSection.tsx`. One-off invoice, the three wallet top-up surfaces and quote invoicing mount `PaymentSettingsSelector.tsx` instead. Those are two components, so the `[Feature] Add payment term options to more billing items` milestone is two changes, tracked on [BIL-673](https://linear.app/getlago/issue/BIL-673/f14-subscription-payment-term-override), [BIL-674](https://linear.app/getlago/issue/BIL-674/f15-one-off-invoice-payment-term-override) and [BIL-675](https://linear.app/getlago/issue/BIL-675/f16-wallet-payment-term-override) rather than here.

Quote-level overrides are out of scope by product decision (BE dive-in, "Quotes"). Quotes show the resolved term instead, which this PR delivers.

## Verification

| Command | Result |
|---|---|
| `pnpm codegen` | ✅ against the live local API, no composed schema |
| `pnpm types` | ✅ zero errors |
| `pnpm lint` | ✅ zero errors (54 pre-existing formik-deprecation warnings elsewhere) |
| `pnpm translations:inspect` | ✅ no missing keys, no unused keys |
| `pnpm translations:ensure-consistency` | ✅ 9/9 files |
| `pnpm test` | ✅ 1028 suites, 12946 tests, 136 snapshots |
| `TZ=America/New_York pnpm test` | ✅ pins the preview timezone regression |
| `pnpm test:e2e` | ⏳ not run — needs an API serving `payment_term`; blocked with the codegen job above |

`TZ=Europe/Paris` fails one unrelated test, `DatePicker › GIVEN a year with fewer than four digits`: for year 26 AD Paris carries a +00:09 LMT offset. It predates this branch (from #4220) and passes in CI's UTC.

Also noticed while typechecking the specs, unrelated and left alone: `cypress/e2e/t40-wallet-top-up.cy.ts` imports `DESKTOP_ACTIONS_BLOCK_TEST_ID`, which `mainHeaderTestIds.ts` no longer exports. Broken on `main` too.

Fixes BIL-552
Fixes BIL-553
Fixes BIL-554
Fixes BIL-555
Fixes BIL-556
Fixes BIL-557
Fixes BIL-558
Fixes BIL-560
Fixes BIL-561
Fixes BIL-562
Fixes BIL-563
Fixes BIL-564

Ref BIL-559
Ref BIL-673
Ref BIL-674
Ref BIL-675


AllanMichay (commented):
Reviewed the full diff. Types are clean, the deleted translation keys have no remaining production references, and the date-math unit tests are solid. Four comments inline, plus one process note: the generated types were produced against a locally-derived schema (`schema.local.graphql`, per the new `.gitignore` entry) that includes the unmerged payment-terms API branch. `UpdateCustomerInput.paymentTerm` / `UpdateBillingEntityInput.paymentTerm` aren't confirmed live on staging. If this merges first, the unknown-field validation error kills the whole operation rather than just the row, so `GetCustomerSettings`, `getBillingEntitySettings` and `getQuote` all blank out. Worth gating the merge on the API side landing.

src/hooks/usePaymentTerm.ts
```diff
@@ -0,0 +1,141 @@
...
+   * can be told apart before saving. The API has no preview endpoint — the date is
+   * computed with the same calendar rules the API applies at finalization.
+   */
+  const getDueDatePreviewCopy = (
+    term: ResolvablePaymentTerm,
+    issuingDate: DateTime = DateTime.now(),
+  ): string => {
+    const { data, plural } = termCopyData(term)
+    const formatDate = (date: DateTime): string =>
+      intlFormatDateTime(date.toISO() ?? '', { formatDate: DateFormat.DATE_MED }).date
```
AllanMichay, line 93: The due-date preview is off by one day in every negative-offset timezone.

`intlFormatDateTime` defaults to `TimezoneEnum.TzUtc` (`src/core/timezone/utils.ts:183`), but `paymentTermDueDate` returns `issuingDate.endOf('month')`, i.e. `23:59:59.999` in the *browser's* zone. With `TZ=America/New_York`, `2026-07-31T23:59:59.999-04:00` renders as **"Aug 1, 2026"**. So for any US/Americas user the End-of-month preview always names the 1st of the next month, and `NetEndOfMonth` is likewise always +1 day.

`DateTime.now()` on line 89 compounds it: after 20:00 EDT the *issuing* date shown is already tomorrow's.

Since the whole point of the preview is to let the user tell `netEndOfMonth` from `daysEndOfMonth`, a systematically wrong date defeats it. Formatting in the zone the DateTime already carries (`setZone: true`, or passing the org timezone) and building `issuingDate` in that same zone should fix both.

src/hooks/usePaymentTerm.ts
```diff
@@ -0,0 +1,141 @@
...
+      value: termType,
+      label: translate(PAYMENT_TERM_TYPE_LABEL_KEYS[termType]),
+      description: translate(PAYMENT_TERM_TYPE_DESCRIPTION_KEYS[termType]),
+    }))
+
+    if (!options?.inheritedFrom) return termTypeOptions
+
+    return [
+      {
+        value: '',
```
AllanMichay, line 126: An option with `value: ''` can never render as selected in our `ComboBox`.

`ComboBox.tsx:132` passes `value={value || null}` to MUI Autocomplete and computes `hasValueSelected={!!value}`, so an empty string is indistinguishable from "nothing selected".

Concretely: open Payment terms on a customer with no own term. The settings row says "Net 30 days (inherit from billing entity)", but the dialog's combo box shows the placeholder "Search or select a payment term" and the inherit row isn't highlighted in the list. If the user explicitly picks the inherit option, `onChange('')` fires and the input visibly clears again.

No other combo box in `src/` uses a selectable `value: ''`. A sentinel (e.g. `'inherit'`) mapped back to `null` at submit time would keep the option selectable.

src/components/settings/invoices/EditPaymentTermDialog.tsx
```diff
@@ -0,0 +1,267 @@
...
+            days: value.days === '' ? 0 : Number(value.days),
+            dayOfMonth: value.dayOfMonth === '' ? null : Number(value.dayOfMonth),
+            monthOffset: value.monthOffset === '' ? null : Number(value.monthOffset),
+          })
+        : null
+
+      isClearingRef.current = paymentTerm === null
+
+      // The billing entity has nothing to inherit from, so it offers no way to clear.
+      if (!paymentTerm && !isCustomer(model)) return
```
AllanMichay, line 181: Submitting a billing entity with no term type is a silent dead end.

`paymentTermFormSchema` returns early when `termType` is falsy, so validation passes, and `canSubmit` is true (the SubmitButton doesn't require dirty). For a billing entity that has never had a term, `getInitialFormValues` seeds `termType: ''`, this guard returns without mutating, `successRef` stays false, and `handleSubmit` throws `'Submit failed'` with `closeOnError: false`. The user sees the button spin and nothing happen, with no field error anywhere.

The row's action is always "Edit" (never "Add") for billing entities, so this is reachable on any fresh one. The `'THEN should submit nothing'` test pins the behaviour but doesn't assert what the user ends up seeing.

Making `termType` required whenever `inheritedFrom` is absent would turn this into a normal field error.

src/components/settings/invoices/EditPaymentTermDialog.tsx
```diff
@@ -0,0 +1,267 @@
...
+      const paymentTerm = value.termType
+        ? buildPaymentTermInput({
+            termType: value.termType,
+            days: value.days === '' ? 0 : Number(value.days),
+            dayOfMonth: value.dayOfMonth === '' ? null : Number(value.dayOfMonth),
+            monthOffset: value.monthOffset === '' ? null : Number(value.monthOffset),
+          })
+        : null
+
+      isClearingRef.current = paymentTerm === null
```
AllanMichay, line 178: The Add flow can emit a "successfully deleted" toast.

`CustomerSettings` renders an **Add** button when `!customer.paymentTerm`, so the dialog opens titled "Add payment terms" with `termType: ''`. Pressing Save without picking anything sends `updateCustomer({ paymentTerm: null })` — a no-op clear — and because `isClearingRef` is true here, `onCompletedToast` fires `text_1787603382163macepxq32tf` ("Payment terms successfully deleted").

Same root cause as the required-`termType` comment below; fixing validation for the create path resolves this one too.

src/hooks/usePaymentTerm.ts
AllanMichay: Fixed in 524e6b7ac. Both halves of this were real.

I did not take the `setZone: true` route, though. The preview is now anchored to a fixed example issuing date and computed as a calendar date in UTC:

- `getDueDatePreviewCopy`'s default `issuingDate` is `PAYMENT_TERM_PREVIEW_ISSUING_DATE` (2026-07-15) parsed in UTC, so the issuing date no longer rolls over after 20:00 EDT.
- `asCalendarDate` re-anchors whatever it is handed via `setZone('utc', { keepLocalTime: true })`, so every DateTime derived from it — `endOf('month')` included — is UTC, and `intlFormatDateTime`'s UTC default agrees with it instead of fighting it.

Why not the org timezone: a due date here is a calendar date rather than an instant, and this preview is an illustrative example, not a real invoice. Two people looking at the same term should read the same sentence, which a zone- or clock-dependent preview can't guarantee. It also restores what BIL-553 (F2) actually specified — "pure function, no today, no timezone", off a fixed example issue date — which the first pass had dropped.

Two regression tests in `usePaymentTerm.test.ts`: End of month previewed from an `America/New_York` DateTime must still read Jul 31, 2026, and the no-argument call must read Jul 15, 2026. I reverted the fix to check they actually catch it — without it they report Aug 1, 2026 and today's date respectively.

This thread didn't auto-outdate because line 103 is still the `intlFormatDateTime` call verbatim; what changed is what gets fed into it.

src/components/paymentTerms/PaymentTermFormContent.tsx
```diff
@@ -0,0 +1,249 @@
...
+ * can inherit is seeded with `PAYMENT_TERM_INHERIT` by its caller.
+ *
+ * The numeric fields are seeded rather than left blank so that switching term type
+ * reveals a usable value straight away. Values belonging to a type the user moved away
+ * from are simply not sent: `buildPaymentTermInput` emits only the chosen type's fields.
+ */
+export const PAYMENT_TERM_FORM_DEFAULT_VALUES = {
+  termType: undefined as PaymentTermType | undefined,
+  days: 0 as number | '',
+  dayOfMonth: PAYMENT_TERM_DAY_OF_MONTH_MIN as number | '',
```
domenicofalco: ideally we shouldn't be using assertion unless they are inevitable

src/components/paymentTerms/PaymentTermFormContent.tsx
```diff
@@ -0,0 +1,249 @@
+import InputAdornment from '@mui/material/InputAdornment'
```
domenicofalco, line 1: this component is made of several parts:
- component
- utils
- schema
and more

following our patterns, this should be split accordingly

src/components/settings/invoices/EditPaymentTermDialog/EditPaymentTermDialog.tsx
```diff
@@ -0,0 +1,283 @@
+import { gql } from '@apollo/client'
```
domenicofalco, line 1: same here, this file should get smaller and spit by file types (hooks, components, utils and so on)

cypress/e2e/10-resources/t100-payment-terms.cy.ts
```diff
@@ -0,0 +1,119 @@
+import { CENTRALIZED_DIALOG_CONFIRM_BUTTON_TEST_ID } from '~/components/dialogs/const'
+import {
+  PAYMENT_TERM_ADD_BUTTON_TEST_ID,
+  PAYMENT_TERM_DELETE_BUTTON_TEST_ID,
+  PAYMENT_TERM_EDIT_BUTTON_TEST_ID,
+  PAYMENT_TERM_SETTINGS_ROW_TEST_ID,
+} from '~/components/paymentTerms/dataTestConstants'
+import { EDIT_PAYMENT_TERM_SUBMIT_BUTTON_TEST_ID } from '~/components/settings/invoices/EditPaymentTermDialog'
```
domenicofalco: Could we move this constant to `paymentTerms/dataTestConstants.ts` and import it from there in both the spec and the dialog? Importing the dialog module pulls in app dependencies, including `main.css`, and this spec currently fails to compile in CI with “The `text-blue` class does not exist”, before any tests run. This is separate from the pending API changes.

src/components/paymentTerms/PaymentTermFormContent.tsx
AllanMichay: Fixed in f86538901. The four casts only existed so `typeof PAYMENT_TERM_FORM_DEFAULT_VALUES` would widen `0` to `number | ''`. Inverted it — `PaymentTermFormValues` is now declared and the constant annotated with it, so the type is the source of truth and there are no assertions left in the file.

src/components/paymentTerms/PaymentTermFormContent.tsx
AllanMichay: Split in f86538901, following the layout in `.agents/docs/folder-architecture.md`:

- `PaymentTermFormContent.tsx` — the component only (249 → 138 lines)
- `types.ts` — `PaymentTermType`, `PaymentTermFormValues`, the defaults
- `utils.ts` — `isConcreteTermType`, `paymentTermFromFormValues`
- `validationSchema.ts` — the zod schema and its range helper
- `dataTestConstants.ts` — the two test ids that were sitting in the component

Also added `__tests__/validationSchema.test.ts`: the schema has conditional per-term-type rules, so the forms guide asks for a direct test, and extracting it to its own file means it would otherwise count as new uncovered code.

src/components/settings/invoices/EditPaymentTermDialog/EditPaymentTermDialog.tsx
AllanMichay: Agreed in principle, but doing it here would make things less consistent rather than more, so we'd like to handle it in a follow-up PR stacked on this branch.

All six sibling dialogs in `src/components/settings/invoices/` keep their zod schema and component in one file — `EditBillingEntityDocumentLocaleDialog`, `EditBillingEntityGracePeriodDialog`, `EditBillingEntityInvoiceNumberingDialog`, `EditBillingEntityInvoiceTemplateDialog`, `EditDefaultCurrencyDialog`, `EditFinalizeZeroAmountInvoiceDialog`. Splitting only the newest one leaves that folder half-converted.

The shape is also a bit different from `PaymentTermFormContent`: of its 283 lines, 46 are the `gql` block (which the styleguide wants colocated with its consumer) and 154 are the `useEditPaymentTermDialog` hook itself, which is one unit. It has no schema of its own — it imports the one that just moved. What's genuinely extractable is ~30 lines: `isCustomer`, `getInheritedFrom`, `getInitialFormValues`.

So the follow-up will convert all seven dialogs in the folder to the same layout, rather than this one in isolation. Happy to be overruled if you'd rather see it in this PR.

domenicofalco (approved):

cypress/e2e/10-resources/t100-payment-terms.cy.ts
domenicofalco: Could we move this constant to paymentTerms/dataTestConstants.ts and import it from there in both the spec and the dialog? Importing the dialog module pulls in app dependencies, including main.css, and this spec currently fails to compile in CI with “The text-blue class does not exist”, before any tests run. This is separate from the pending API changes @AllanMichay

AllanMichay: Done in 3362e3926 — and the diagnosis is exactly right, thanks for spotting the compile failure rather than just the import.

`EDIT_PAYMENT_TERM_SUBMIT_BUTTON_TEST_ID` now lives in `paymentTerms/dataTestConstants.ts`, imported from there by the spec, the dialog and the dialog's jest test.

Confirmed the mechanism: importing the dialog pulled the component graph in, and through the design-system barrel (`import './style.css'`) that put the app's CSS through postcss during spec bundling. The spec's `~/` import graph went from 49 modules to 4, none carrying CSS. It now matches every other spec in `cypress/e2e`, which all import test ids from constants modules and never from a component.

One thing worth flagging separately: the `text-blue` failure itself is not ours. `@apply text-blue` at `src/main.css:10` is identical on `main` and `text-blue` isn't defined in the tailwind theme reachable from that context — it has simply never surfaced because no spec had dragged CSS into its bundle before. Worth a separate fix, otherwise the next spec that imports a component hits the same wall.

domenicofalco (approved):

PR #4183, opened 2026-08-25
fix(catalog): design-system blue links, unify visited state

ansmonjol:
## Context

Cross-object links in the catalog details pages (product, product filter, rate card) rendered black instead of the design-system blue used everywhere else in the app, for instance the customer link in invoice details.

Chasing that down surfaced a second, app-wide inconsistency: the global `a` rule turned every visited link purple, and seven call sites had already worked around it one by one with a local `visited:text-blue`. This PR unifies that instead of adding an eighth workaround.

## Description

One commit per concern:

- **`fix(catalog): blue design-system links in details`** - the six catalog detail links wrapped their label in `<Typography variant="body" color="grey700">`, and that explicit colour overrode the global `a { @apply text-blue ... }` rule. The wrapper was redundant anyway, since `DetailsPage.InfoGrid` already wraps every value in exactly that Typography, so it is dropped and the anchors inherit the design-system blue.
- **`fix(design-system): keep visited links blue`** - unifies visited-link styling app-wide: drops `visited:text-purple` from the global `a` rule in `src/main.css` and `packages/design-system/src/style.css`, and removes the seven scattered `className="visited:text-blue"` overrides that existed only to undo it.
- **`refactor(catalog): extract product overview grid values`** - pulls the inline multi-line ternaries out of the `InfoGrid` array into consts above the return, matching the two sibling pages and the "Prefer Logic Out of JSX" rule.
- **`fix(catalog): show product display name on rate card`** - the rate card overview rendered `product.name` while the product filter overview renders `invoiceDisplayName || name` for the same Product. Selects the field in the fragment and aligns the two.
- **`fix(catalog): correct product filter empty-state copy`** - the "Attached category" row rendered the copy "No product"; also finishes the half-done `Product` -> `ProductCategory` test-id rename, and replaces a dead `color="info600"` on the rate card table link (it maps to purple and never rendered, the anchor rule always won) with `color="inherit"`.

No Linear ticket for this one.

### DIFF
--- a/packages/design-system/src/style.css
+++ b/packages/design-system/src/style.css
@@ -14,7 +14,7 @@
   }
 
   a {
-    @apply text-blue no-underline visited:text-purple hover:underline focus:rounded focus:ring;
+    @apply text-blue no-underline hover:underline focus:rounded focus:ring;
   }
 
   b {
--- a/src/components/creditNote/CreditNoteDetailsExternalSync.tsx
+++ b/src/components/creditNote/CreditNoteDetailsExternalSync.tsx
@@ -117,7 +117,7 @@ const OverviewLine: FC<
       )}
       {link && (
         <Link
-          className="w-fit line-break-anywhere visited:text-blue hover:no-underline"
+          className="w-fit line-break-anywhere hover:no-underline"
           target="_blank"
           rel="noopener noreferrer"
           to={link}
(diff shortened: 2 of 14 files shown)


ansmonjol:
Not applying: regression coverage for the link styling was raised in review this round and explicitly declined by the author. The change is a CSS-inheritance fix with no branch to assert, and the only hook available in jsdom would be the absence of `Typography`'s `data-test="body"` inside the anchor, which pins a design-system implementation detail rather than the behaviour.

This verdict was also computed against c5bb3ee8, four commits ago; the PR has since grown the app-wide visited-link unification and three further fixes.

domenicofalco (approved):

PR #4184, opened 2026-08-25
feat(catalog): select typed rateProperties subfields

rsempe:
## Context

getlago/lago-api#6224 types `rateProperties` as `Properties` instead of a raw `JSON` scalar (requested in Slack). A typed object requires subselections, so the bare `rateProperties` selection in the rate-card list fragment breaks codegen against the new schema.

## Description

Select the subfields `formatActiveRate` actually consumes: `amount`, `packageSize`, `rate`, and the per-tier `perUnitAmount` / `rate` on the three range types. The formatter already reads camelCase keys, so no logic change — the typed schema just makes those reads reliable for REST-created rates too (the JSON scalar passed through whatever casing the writer used).


rsempe:
Closing — superseded by #4185, which landed the same subselection via a shared `PropertiesForActiveRate` fragment and additionally retyped `formatActiveRate` properly. Nothing left to salvage from this one.

PR #4187, opened 2026-08-26
fix(deps): bump react-router-dom to 6.30.6

ansmonjol:
Related: LAGO-1859 (React Router v6 -> v7 upgrade)

Patches **CVE-2026-53668** / GHSA-jjmj-jmhj-qwj2 (open redirect leading to XSS), Dependabot alerts #250 and #252. `6.30.6` is the last v6 release ever published, and this is the only one of the three open react-router advisories that has a v6 fix.

The real change is `@remix-run/router` `1.23.3` -> `1.23.4`, which drops the `isAbsoluteUrl` early-return in `resolvePath`. Verified against the installed tree:

| input | 6.30.4 | 6.30.6 |
| --- | --- | --- |
| `javascript:alert(1)//` | passed through verbatim | `/acme/customers/javascript:alert(1)/` |
| `//evil.com` | `//evil.com` | `/evil.com` |

`react-router` and `react-router-dom` themselves only get a version-banner bump between 6.30.4 and 6.30.6 - the `.d.ts` surface is byte-identical, so there is no API risk. Resolved graph stays on a single version: `react-router-dom@6.30.6` -> `react-router@6.30.6` -> `@remix-run/router@1.23.4`.

Both manifests are bumped because both pin the version exactly (root `package.json` and `packages/design-system/package.json`).

The 6.30.6 changelog flags one possible breaking bug fix, "if you are using `navigate` for external navigations". We have zero external `navigate()` calls, and the only external `<Link to="https://...">` is in `src/pages/__devOnly/DesignSystem.tsx`, which goes through the `ABSOLUTE_URL_REGEX` path in `react-router-dom` that did not change.

### What this does not fix

CVE-2026-53669 (alert #253) and CVE-2026-53666 (alert #251) have **no v6 patch and never will** - React Router v6 has been EOL since 2026-06-18 and upstream has stated it will not backport further CVE fixes to the 6.x line. Both advisories use a `< 7.18.0` range, so they stay open on any v6 version. Neither is reachable in this app today (no SSR, and `ensureSlugPrefix` neutralises the only untrusted-input-to-navigate path). They are intentionally left open rather than dismissed, and the v7 upgrade is tracked in LAGO-1859.

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,7 +171,7 @@
     "react-dom": "18.3.1",
     "react-markdown": "10.1.0",
     "react-resizable-panels": "3.0.6",
-    "react-router-dom": "6.30.4",
+    "react-router-dom": "6.30.6",
     "recharts": "2.15.4",
     "remark-gfm": "4.0.1",
     "sanitize-html": "2.17.5",
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -30,7 +30,7 @@
     "@mui/x-date-pickers": "6.20.2",
     "react": "18.3.1",
     "react-dom": "18.3.1",
-    "react-router-dom": "6.30.4"
+    "react-router-dom": "6.30.6"
   },
   "devDependencies": {
     "@svgr/core": "8.1.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -199,8 +199,8 @@ importers:
         specifier: 3.0.6
         version: 3.0.6(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       react-router-dom:
-        specifier: 6.30.4
-        version: 6.30.4(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
+        specifier: 6.30.6
+        version: 6.30.6(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       recharts:
         specifier: 2.15.4
         version: 2.15.4(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
@@ -494,8 +494,8 @@ importers:
         specifier: 18.3.1
         version: 18.3.1(react@18.3.1)
       react-router-dom:
-        specifier: 6.30.4
-        version: 6.30.4(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
+        specifier: 6.30.6
+        version: 6.30.6(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
     devDependencies:
       '@svgr/core':
         specifier: 8.1.0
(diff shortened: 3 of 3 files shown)


ansmonjol:
Closing in favour of doing this properly as part of the v7 upgrade in **LAGO-1859**.

6.30.6 would have fixed CVE-2026-53668 (alerts #250/#252), but it is the last v6 release that will ever exist and the other two advisories (#251 CVE-2026-53666, #253 CVE-2026-53669) have no v6 fix and never will. Shipping this in isolation buys one closed alert and leaves us on an EOL major, so the review findings around it are better addressed once rather than twice.

Folded into LAGO-1859 instead:

- `packages/design-system/package.json` carries a `react-router-dom` peer dep that the package never imports (zero hits in `packages/design-system/src/`) - it survives only in `package.json:33` and `vite.config.ts:42`/`:47`. Drop it so the v7 bump touches one manifest, not two.
- Exact pins duplicated across both manifests with `autoInstallPeers: true` risk a second react-router copy (two Router contexts) on any partial bump. A pnpm `catalog:` entry removes the class.
- No open-redirect test coverage anywhere: `prependOrgSlug.test.ts` covers only well-formed paths, and `useNavigate.test.ts` mocks `prependOrgSlug` out entirely. The v7 upgrade changes exactly this normalisation, so the cases need pinning first.
- `ensureSlugPrefix` (`src/core/router/utils/orgSlug.ts:66`) does not reject external targets - `//evil.com` becomes `/acme//evil.com`, safe only because `@remix-run/router`'s `removeDoubleSlashes` collapses it afterwards. That is the function upstream rewrote twice in three months. Validate at the sink instead.
- No `renovate.json` rule parks the react-router major, and `vulnerabilityAlerts` bypasses `minimumReleaseAge`, so Renovate will keep re-opening the 6->7 PR.
- Worth recording: GHSA-jjmj-jmhj-qwj2 carries `first_patched_version: null` and 6.30.6 escapes only via its hard `<= 6.30.4` upper bound, and 6.30.5 clears that same bound while still resolving the **unfixed** `@remix-run/router@1.23.3`. Any future "just bump past the range" remediation can close the alert without fixing the code.

One correction to this PR's own description: it claimed the only external `<Link to="https://...">` was in `__devOnly/DesignSystem.tsx`. There are ~15 production ones (5 `mailto:` template literals plus the `build*Url` helpers in `src/core/constants/externalUrls.ts`). They are all still safe - `react-router-dom` sets `absoluteHref = to` for anything matching `ABSOLUTE_URL_REGEX` and renders `href: absoluteHref || href`, so `resolvePath` output never reaches the DOM for external targets - but the audit behind that sentence was narrower than stated. It matters for the v7 upgrade, where that regex changes.

All three Dependabot alerts stay open and undismissed until v7 lands. Branch `react-router-cve-patch-check` left in place for reference.

PR #4192, opened 2026-08-26
chore(skills): gate triage review FAIL on blocking defects

domenicofalco:
## Context

The triage-frontend-ticket skill runs an independent review at `full` depth, capped
at 2 rounds, with a binary PASS/FAIL gate. A reviewer with codebase access always
finds something, so any run whose analysis is otherwise sound still exhausts the
budget on wording-level findings and ships with the "could not confirm" warning
header — even when the diagnosis and the fix were confirmed.

Observed on a real run: round 1 caught a genuine regression in the proposed fix,
round 2 caught one criterion gap plus five nits (a wrong git hash, an enum with an
identical string value, a false parenthetical, a loose criterion, a mis-cited doc).
None of the five changed a line of the implementation, and a third round would have
produced five more at the same rate.

## Description

- Add a LEARNINGS entry that reserves FAIL for defects which change the
  implementation — wrong diagnosis, a fix that does not work, a missed call site, an
  acceptance criterion that lets the bug ship. Everything else becomes
  `PASS with notes`: the notes get applied without spending a review round. The cap
  of 2 stays as is — it is what makes the loop converge, and LEARNINGS may retune
  judgement but never soften a guardrail.
- Drop the tracker ticket ids from the existing entries and from the entry template,
  which now asks for the surface instead. This is a public repository; an internal
  ticket id means nothing to a reader outside the team.

### DIFF
--- a/.agents/skills/triage-frontend-ticket/LEARNINGS.md
+++ b/.agents/skills/triage-frontend-ticket/LEARNINGS.md
@@ -20,7 +20,8 @@ An entry that cannot name what corrected it does not belong here.
 ```
 ## YYYY-MM-DD — <one-line rule>
 
-**What happened:** <the specific outcome, with the ticket and the correcting signal>
+**What happened:** <the specific outcome, named by its surface — never a ticket id — with the
+correcting signal>
 **Rule:** <what to do differently, stated as an instruction>
 ```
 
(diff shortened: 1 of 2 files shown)


domenicofalco:
Thanks for the review! Leaving as is: the file's rule asks that an entry be backed by something "identifiable and human" — i.e. that it *name the correcting signal* — not that the signal be publicly linkable. The entry names it: an operator push-back on the run's binary gate, which is the same class as "a review that overruled the analysis" in the requirement's own list.

Two reasons the suggested remedy would make things worse rather than better:

- None of the four pre-existing entries name a person or carry a link either ("the independent review found `RoleDetails.tsx:72`", "a criterion read ..."). They name the surface and the signal. The new entry follows exactly that shape, so singling it out would be inconsistent.
- Adding a maintainer name would contradict the other half of this same PR, which strips tracker ticket ids from every entry precisely because this is a public repository. Swapping an internal id for an internal person's name trades one leak for a worse one.

CI is green on `b512ced`.

---

ansmonjol (approved):

PR #4193, opened 2026-08-26
feat(catalog): create and edit rate card rates

ansmonjol:
Fixes BIL-501

Adds the Rates tab on a rate card, the create/edit drawer, the rate detail page and the delete flow.

## What ships

- **Rates tab** with status, effective date, code, price and creation date. Numbered pagination through `PaginatedContent` + `usePageSearchParam`, `rateCardRates` registered in `queryFieldPolicies`. No search and no filters, per the design. Maneki empty state carrying the create CTA.
- **Create / edit drawer** following the canonical `useRateCardRateDrawer()` hook pattern. Rate settings (effective date, code, billing interval), Pricing settings, Invoicing settings.
- **Row and page actions**: View, Edit, Delete. Delete is offered on pending rates only, whatever the parent card is attached to: a pending rate has never priced anything, and `RateCardRates::DestroyService` refuses exactly one thing, a non-pending rate (`only_pending_rates_can_be_deleted`). A terminated rate, or a non-pending rate on a subscribed card, gets View alone.
- **Rate detail page** at `/product-catalog/rate-cards/:rateCardId/rates/:rateId/:tab`, breadcrumb middle crumb pointing back at the card's rates tab.
- **Toasts** for create, create-more and edit, the create ones linking to the new rate.

## Backend rules encoded in the form

The drawer mirrors `RateCardRates::CreateService` / `UpdateService` rather than guessing:

- The effective date seeds the code as `rate_MM_DD_YYYY`. Editing the code by hand stops the seeding for good, the same relationship name and code have elsewhere. Two refs keep the programmatic write from being read as a hand edit.
- A new date must land strictly after the currently effective rate (`validate_effective_from_is_appended`). The boundary is dropped when the active rate is the one being edited, otherwise every save of an active rate is rejected against itself.
- An active rate only accepts pricing changes. `FROZEN_ON_ACTIVE` is rejected on presence even when the value is unchanged, so the update payload omits those fields entirely rather than sending them back.
- The code is locked when the card is attached to a plan or a subscription, on edit only.
- A spending minimum is arrears-only, and the conversion rate is required when the card carries a pricing unit.
- `effectiveFrom` is a calendar day, so the picker, the payload and both display sites are pinned to UTC. Reading it in the organization timezone showed the previous day for anything west of UTC.
- Duplicate-code and duplicate-date both come back as the same error code, so the mapping is scoped by the field the error names. Anything else surfaces as a danger toast instead of being swallowed.

## Shared components

Pricing settings reuse `ChargeModelSelector` and `ChargeWrapperSwitch`, so every rate model gets the charge form it already has in plans, alerts included. Both were adjusted:

- `ChargeWrapperSwitch`: `isFilterForm` became `showPresentationGroupKeys` (default `true`). A rate has no charge filters, so it can never carry presentation group keys. The single existing caller was updated. The dead `isEdition` prop was also dropped, and the removal was carried through the whole chain it fed: `ChargeWrapperSwitch`, `ChargeFilterDrawerContext`, `ChargeFilterDrawerProvider`, `UsageChargeDrawerContent` and `UsageChargeDrawer`. Its last reader disappeared in #3107, but because it stayed declared and required, every call site kept computing and forwarding a value nothing read. `isEdition` elsewhere (plan settings, fixed charges, subscription fee) is still live and untouched.
- `ChargeModelSelector`: `label` is now **required** rather than defaulted, and the inline props object became a named `ChargeModelSelectorProps`. The two charge drawers pass the "Charge model" key explicitly and the rate drawer passes "Rate model", so the component no longer hardcodes a charge-specific default. `alreadyUsedChargeAlertMessage` and `isInSubscriptionForm` became optional, so the rate drawer stops passing `undefined` for both.

The details page reuses `PlanDetailsChargeWrapperSwitch` unchanged.

## Catalog design consistency

Reviewing the new rate pages surfaced three places where the catalog had forked from the rest of the app. One commit each, all reviewable independently of the feature.

- **Blue links** (`fix(catalog): blue links on rate detail overview`). The four cross-object links on the rate overview wrapped their label in `<Typography variant="body" color="grey700">`, and that explicit colour overrides the global `a { @apply text-blue }` rule, so they rendered black. Exactly the defect #4183 fixed on the sibling catalog pages. The wrapper is dropped, since `DetailsPage.InfoGrid` already applies that Typography. The grid values also moved into `attached*` consts above the return, matching the three sibling overviews.
- **`ActivityLogsSection`** (`refactor(catalog): use ActivityLogsSection for logs tabs`). `RateCardActivityLogs` and `ProductFilterActivityLogs` hand-rolled the heading, `PaginatedContent`, `ActivityLogsTable` and the developer-tool row-click that `ActivityLogsSection` already owns and nine other resources use. Both now pass `subtitle` plus the query result.
- **`PageSectionTitle`** (`refactor(catalog): adopt PageSectionTitle in details`). The `flex h-18 items-center justify-between gap-4` title/description/action header was copy-pasted at nine more catalog sites, a fork of `PageSectionTitle` in `~/components/layouts/Section`. All nine migrated, and `grep 'h-18 items-center justify-between' src/pages/catalog` now returns nothing. **This one is visible**: subtitles go from `caption` to `body`, and the wrapper from `h-18 items-center gap-4` to `mb-4 items-baseline gap-2`. If design prefers the old catalog sizing, the fix belongs in `PageSectionTitle` so every page moves together rather than in a catalog-only fork.
- **Rate activity-logs header** (`fix(catalog): rate activity logs tab header`). The empty-tab placeholder rendered its own `flex flex-col gap-1` block with a `body` subtitle, so it did not match the overview tab sitting next to it. Now `PageSectionTitle`, like every other activity-logs tab in the app.
- **Tab top spacing** (`fix(catalog): top spacing on details tab content`). `DetailsPage.Container` has no top padding: the old `h-18` header was what created the gap under the tab bar, and `PageSectionTitle` does not. Added `pt-6` at the 13 catalog `DetailsPage.Container` sites, matching the `pt-6` / `py-6` the plan, feature, quote, subscription and billable-metric tabs already use. Deliberately not added to `DetailsPage.Container` or `PageSectionTitle` themselves, since both would double-pad their existing consumers.

## Review fixes

A review pass over the branch turned up eleven items, all fixed. The behavioural ones:

- **Model switch discarded input silently** (`fix(catalog): keep rate drawer dirty on model switch`). Switching the rate model called `form.reset`, which clears every field's `isDirty`. The drawer's `shouldPromptOnClose` reads `form.state.isDirty`, so closing right after a model switch skipped the unsaved-changes prompt and threw the date, code and pricing away. Now only the outgoing model's property field meta is reset, via `resetField` on the `properties.*` paths.
- **Clearing the date broke validation** (`fix(catalog): translate cleared effective date error`). The picker writes `undefined`, and a strict `z.object` aborts before `superRefine`, so the field showed zod's untranslated "Required" and every other issue vanished until the date was refilled. The schema is now `z.custom<RateCardRateFormValues>().superRefine(...)`, the same shape the plan, invoice, wallet and coupon schemas use.
- **`rateProperties` cache clobber** (`fix(catalog): align rateProperties cache selections`). The detail query wrote the full `PropertiesForRateCardRate` while the rate-card list wrote the narrower `PropertiesForActiveRate` for the same normalized rate, and Apollo replaces array fields wholesale, so each write stripped the other's range fields. The wide fragment now spreads the narrow one, so the superset relationship is enforced by GraphQL, and the list selects the wide one.
- **`$0.00` spending minimum** (`fix(catalog): hide zero spending minimum on rate`). A rate saved without one stores 0, which rendered as a configured zero floor rather than "none set" and contradicted the drawer's "Add" affordance.
- **Effective-date formatting** (`fix(catalog): consistent effective date formatting`). The boundary date quoted in the error copy was hardcoded `MM/dd/yyyy` while the table and the overview render `Jun 24, 2026` for the same value.

The rest were fragment duplication in `RateCardRateForList` (six fields already carried by the spread `RateCardRateForDrawer`), a timezone test whose mock returned the same string for both formatters so it could not fail, a `Settings.defaultZone` restored inline instead of in `afterEach`, the `escapeDoubleQuotes` placement above, and a comment that described the delete flow as evicting instead of refetching when it does both.

## Notes

- The Activity logs tab ships empty. `rate_card_rate` is not in `Clickhouse::ActivityLog::RESOURCE_TYPES` yet, tracked in BIL-594.
- `escapeDoubleQuotes` was duplicated verbatim in four catalog drawers, so it moved to `src/pages/catalog/utils/`, next to `formatActiveRate`. Feature-local rather than `~/core/utils` because all five consumers are catalog drawers, which is what `folder-architecture.md` prescribes.
- The same pricing-unit code-versus-short-name bug that is fixed in the new columns hook still ships in `useRateCardTableColumns.tsx`. Left alone here to keep the diff scoped.
- `RateCardRateDetailsOverview` renders `product.name` while the sibling rate card overview renders `invoiceDisplayName || name` for the same Product, since #4183 aligned it. Not changed here because it needs a fragment field and a codegen run; say the word and it goes in.


src/core/constants/statusRateCardRateMapping.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * A rate's status is derived server-side from the card's append-only timeline (future rate =
+ * pending, latest effective = active, superseded = terminated), so a terminated rate is a normal
+ * end of life rather than an incident: it reads as disabled, not danger (unlike subscriptions).
+ */
+export const rateCardRateStatusMapping = (status: RateCardRateStatusEnum): StatusProps => {
+  switch (status) {
+    case RateCardRateStatusEnum.Active:
+      return { type: StatusType.success, label: 'active' }
```
domenicofalco, line 9: should those be translated labes?

src/pages/catalog/details/RateCardRateDetails.tsx
```diff
@@ -0,0 +1,260 @@
...
+export const RATE_CARD_RATE_DETAILS_DELETE_TEST_ID = 'rate-card-rate-details-delete'
+
+// New translation keys are exported as named constants (feature convention).
+/** The query selects both card fragments, so the snapshot builder can read either half. */
+type RateCardForRateDetailsAndDrawerFragment = RateCardForRateDetailsFragment &
+  RateCardForRateDrawerFragment
+
+export const RATE_CARD_RATE_BREADCRUMB_KEY = 'text_1787737220228sofw78j0u83'
+export const RATE_CARD_RATE_NOT_FOUND_KEY = 'text_1787737220228b97cdfac4py'
+export const RATE_CARD_RATE_ACTIVITY_LOGS_EMPTY_KEY = 'text_1787819065018edyn6m968e4'
```
domenicofalco, line 81: don't really get whey labels are exported and not used directly and why they live here

src/pages/catalog/details/RateCardRateDetailsOverview.tsx
```diff
@@ -0,0 +1,264 @@
...
+export const RATE_CARD_RATE_DETAILS_OVERVIEW_EDIT_TEST_ID = 'rate-card-rate-details-overview-edit'
+export const RATE_CARD_RATE_DETAILS_OVERVIEW_STATUS_TEST_ID =
+  'rate-card-rate-details-overview-status'
+
+// New translation keys are exported as named constants (feature convention) so tests and
+// siblings reference them instead of duplicating the raw ids.
+export const RATE_CARD_RATE_DETAILS_PRODUCT_CATEGORY_LABEL_KEY = 'text_17877372202296ejgkqky70w'
+export const RATE_CARD_RATE_DETAILS_RATE_CARD_LABEL_KEY = 'text_1787737220228091rkbqj1vl'
+export const RATE_CARD_RATE_DETAILS_CODE_LABEL_KEY = 'text_1787737220228i16tnwmeue3'
+export const RATE_CARD_RATE_DETAILS_BILLING_INTERVAL_VALUE_KEY = 'text_17877372202287udsa3vj1ul'
```
domenicofalco, line 78: same here. they seems to be used for testing purpose which should be an anti-pattern of our testing structure

src/pages/catalog/details/RateCardRateDetailsOverview.tsx
```diff
@@ -0,0 +1,264 @@
...
+              <TypographyWithCopy variant="body" color="grey700">
+                {rate.code}
+              </TypographyWithCopy>
+            ),
+          },
+          {
+            label: translate(RATE_CARD_RATE_EFFECTIVE_DATE_LABEL_KEY),
+            // A calendar day stored as a UTC instant: reading it in the organization timezone
+            // would show the previous day for anything west of UTC.
+            value: intlFormatDateTime(rate.effectiveFrom, { timezone: TimezoneEnum.TzUtc }).date,
```
domenicofalco, line 209: just making sure is it fine we don't use org tz here?

src/pages/catalog/details/RateCardRateDetailsOverview.tsx
```diff
@@ -0,0 +1,264 @@
...
+          },
+        ]}
+      />
+
+      <PlanDetailsChargeWrapperSwitch
+        currency={rateCard.currency}
+        // The two enums are distinct GraphQL types with identical string members, and the
+        // switch keys off those strings.
+        chargeModel={rate.rateModel as unknown as ChargeModelEnum}
+        values={rate.rateProperties as unknown as Properties}
```
domenicofalco: any chance to get rid of these 2 assertions?

src/pages/catalog/details/RateCardRatesTab.tsx
```diff
@@ -0,0 +1,148 @@
...
+  ${RateCardRateForListFragmentDoc}
+`
+
+export const RATE_CARD_RATES_TAB_TEST_ID = 'rate-card-rates-tab'
+export const RATE_CARD_RATES_CREATE_TEST_ID = 'rate-card-rates-create'
+
+// New translation keys are exported as named constants (feature convention) so tests and
+// siblings reference them instead of duplicating the raw ids.
+export const RATE_CARD_RATES_EMPTY_TITLE_KEY = 'text_1787737220228wjwpgmwu8fv'
+export const RATE_CARD_RATES_EMPTY_SUBTITLE_KEY = 'text_1787737220228bjix2qjx4rz'
```
domenicofalco, line 48: same here with translations

src/pages/catalog/details/RateCardRatesTab.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+type RateCardRatesTabProps = {
+  rateCardId: string
+  /**
+   * Undefined until the parent card query resolves. The rates query only needs the id, so it
+   * runs in parallel with it rather than behind it; the card gates the create CTA and the row
+   * write actions, which cannot be decided without it.
+   */
```
domenicofalco: I feel this is an unnecessary comment

src/pages/catalog/dialogs/useDeleteRateCardRateDialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+        const destroyedId = data?.destroyRateCardRate?.id
+
+        // A backend rejection resolves without data (errorPolicy 'all'); the global error
+        // link surfaces it as an error toast.
+        if (destroyedId) {
+          // Alongside the refetch, not instead of it: eviction drops the deleted entity and
+          // its row from the cached list pages straight away, so a still-mounted rate
+          // details query re-reads nothing rather than a record the server already removed,
+          // and the rates tab does not paint a stale row while the refetch is in flight.
```
domenicofalco: the code well explains itself here, i feel these 2 comments are unneeded

src/pages/catalog/drawers/rateCardRate/constants.ts
```diff
@@ -0,0 +1,240 @@
...
+// Copy shared by more than one surface (drawer, rates tab, row actions, details page) lives
+// here so a read-only surface never has to import the drawer module to reach a label.
+export const RATE_CARD_RATE_DRAWER_TITLE_CREATE_KEY = 'text_1787737220227lhrw4x3r4h8'
+export const RATE_CARD_RATE_DRAWER_TITLE_EDIT_KEY = 'text_1787737220227dhuxfszp0v6'
+export const RATE_CARD_RATE_DRAWER_DESCRIPTION_KEY = 'text_17877372202276uc54jqy1np'
+export const RATE_CARD_RATES_SECTION_TITLE_KEY = 'text_1784930705742tg0kbcsak2v'
+export const RATE_CARD_RATE_VIEW_ACTION_KEY = 'text_1787737220228sypguqmiv1l'
+export const RATE_CARD_RATE_DUPLICATE_DATE_KEY = 'text_1787753924848luck8g8y1qd'
+export const RATE_CARD_RATE_SAVE_FAILED_KEY = 'text_1787753924848adhyrzqb0gz'
+export const RATE_CARD_RATE_DELETE_ACTION_KEY = 'text_1787737220228txu8nd2qayi'
```
domenicofalco, line 29: not sure if this is a pattern we want to enforce not to have in CLADE.md? IIRC we don't reuse translation keys that way within the app when they are in more then 1 place

src/pages/catalog/drawers/rateCardRate/constants.ts
```diff
@@ -0,0 +1,240 @@
+import { DateTime } from 'luxon'
+import { z } from 'zod'
+
```
domenicofalco: I feel that many of the functions in this file should be placed in a utils file rather then `constants.ts` like schema and some helpers. wdyt?

src/pages/catalog/drawers/rateCardRate/constants.ts
```diff
@@ -0,0 +1,240 @@
...
+  return date.isValid ? `rate_${date.toFormat('MM_dd_yyyy')}` : undefined
+}
+
+/**
+ * The same rendering the rates table and the rate overview use, so the boundary date quoted in
+ * the error copy reads identically to the dates shown beside it. UTC for the reason above.
+ */
+export const formatEffectiveDate = (isoDate: string): string =>
+  intlFormatDateTime(isoDate, { timezone: TimezoneEnum.TzUtc }).date
```
domenicofalco: is UTC always what we want here?

src/pages/catalog/drawers/rateCardRate/constants.ts
```diff
@@ -0,0 +1,240 @@
...
+        path: ['conversionRate'],
+        message: VALUE_REQUIRED_KEY,
+      })
+    }
+
+    validateChargeProperties(
+      // The two enums are distinct GraphQL types with identical string members, and the
+      // charge validators key off those strings.
+      values.rateModel as unknown as AnyChargeModel,
+      values.properties as PropertiesZodInput | undefined,
```
domenicofalco: any change these 2 assertions (`values.rateModel` and `values.properties`) can be avoided?

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
```diff
@@ -0,0 +1,491 @@
...
+   * attached to a plan or a subscription (`RateCardRates::UpdateService`).
+   */
+  isCodeLocked: boolean
+  /**
+   * `effectiveFrom` of the rate currently in effect - a new date must land strictly after it.
+   * Read through a getter because `children` is captured once at open() while the boundary
+   * moves whenever a "create more" save produces an already-effective rate.
+   */
+  getEffectiveFromBoundary: () => string | null
+  /** Captured at open time so clearing the input does not collapse the section mid-edit. */
```
domenicofalco: not sure we need a comment to describe each of the props

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
```diff
@@ -0,0 +1,491 @@
...
+  RATE_CARD_RATE_EFFECTIVE_DATE_DESCRIPTION_KEY,
+  RATE_CARD_RATE_EFFECTIVE_DATE_LABEL_KEY,
+  RATE_CARD_RATE_FORM_DEFAULTS,
+  RATE_CARD_RATE_MODEL_LABEL_KEY,
+} from './constants'
+
+// New translation keys are exported as named constants (feature convention) so tests and
+// siblings reference them instead of duplicating the raw ids.
+export const RATE_CARD_RATE_SETTINGS_SECTION_TITLE_KEY = 'text_1787737220227io0cqa5y5jy'
+export const RATE_CARD_RATE_SPENDING_MINIMUM_DESCRIPTION_KEY = 'text_1787737220228b7rp4u9zxp6'
```
domenicofalco, line 55: labels exported here as well

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
```diff
@@ -0,0 +1,491 @@
...
+
+      return translate(RATE_CARD_RATE_EFFECTIVE_DATE_AFTER_ACTIVE_KEY, {
+        date: formatEffectiveDate(effectiveFromBoundary),
+      })
+    }, [effectiveFrom, effectiveFromBoundary, translate])
+
+    // ChargeModelSelector / ChargeWrapperSwitch are bound to the charge-world shape; a rate
+    // only carries the model, so adapt it into a synthetic local charge for them.
+    const localCharge = {
+      chargeModel: rateModel as unknown as ChargeModelEnum,
```
domenicofalco: any change this assertion can be mitigated as well?

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
```diff
@@ -0,0 +1,491 @@
...
+    } as LocalUsageChargeInput
+    const spendingMinimumLocalCharge = { minAmountCents } as unknown as LocalUsageChargeInput
```
domenicofalco: assertion: same here and along the entire component perhaps?

src/pages/catalog/drawers/rateCardRate/useRateCardRateDrawer.tsx
```diff
@@ -0,0 +1,545 @@
+import { FetchResult, gql } from '@apollo/client'
+import { revalidateLogic } from '@tanstack/react-form'
```
domenicofalco: I see more hooks in this file. any change we can split it to reduce the file size?

src/pages/catalog/drawers/rateCardRate/useRateCardRateDrawer.tsx
```diff
@@ -0,0 +1,545 @@
...
+        fromValue,
+        toValue,
+        flatAmount,
+        perUnitAmount,
+      }),
+    ),
+  }
+}
+
+const mapRateToFormValues = (
```
domenicofalco: I see functions like this in this hook that can perhaps live standalone in a utils with its own unit test so to reduce the file size

src/core/utils/escapeDoubleQuotes.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * `data-text` is a double-quoted HTML attribute in the linked-toast templates, so an entity
+ * name reaching one must not be able to break out of it.
+ */
+export const escapeDoubleQuotes = (value: string): string => value.replaceAll('"', '&quot;')
```
domenicofalco, line 2: I see this more a generic utils rather then bounded to catalog feature. perhaps it can be moved above. wdyt?

src/pages/catalog/useRateCardRateTableActions.tsx
```diff
@@ -0,0 +1,132 @@
...
+    tab: RateCardRateDetailsTabsOptionsEnum.overview,
+  })
+
+type UseRateCardRateTableActionsProps = {
+  rateCardId: string
+  /**
+   * Undefined while the parent card query is still in flight - the rows can already be listed
+   * by then, and every write action needs the card to know what the backend would accept, so
+   * they only appear once it lands.
+   */
```
domenicofalco: I feel here there are some unneeded comments over props within this file

src/pages/catalog/useRateCardTableColumns.tsx
```diff
@@ -45,16 +45,20 @@ gql`
     activeRate {
       id
       rateModel
+      # The rate detail page and the rates tab write the full PropertiesForRateCardRate for
+      # this same normalized rate. Apollo replaces array fields wholesale, so writing the
+      # narrower PropertiesForActiveRate here would strip the range fields they cached and
+      # drive them back to the network. The price cell still only reads the narrow subset.
```
domenicofalco: unneeded comment

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
```diff
@@ -0,0 +1,491 @@
...
+    const isSeedingCodeRef = useRef(false)
+
+    const rateModelComboboxData = useMemo(() => {
+      if (rateCard.productType === ProductTypeEnum.Fixed) {
+        return getFixedChargeModelComboboxData()
+      }
+
+      if (!rateCard.aggregationType) return []
+
+      return getUsageChargeModelComboboxData({
```
domenicofalco, line 140: Tied to my comment further down: this pushes `ChargeModelEnum.Custom` for a `custom_agg` metric with no premium gate, so if the custom-JSON write path stays unwired, filtering it out here would probably be the smallest way to close the dead end.

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
```diff
@@ -0,0 +1,491 @@
...
+
+          <ChargeModelSelector
+            label={translate(RATE_CARD_RATE_MODEL_LABEL_KEY)}
+            disabled={isActiveRate}
+            localCharge={localCharge}
+            chargeModelComboboxData={rateModelComboboxData}
+            handleUpdate={handleRateModelUpdate}
+          />
+
+          <ChargeWrapperSwitch
```
domenicofalco, line 353: `ChargeWrapperSwitch` here doesn't get `onExpandCustomCharge`, and I think that might leave the Custom rate model without a write path. `CustomCharge` passes the `JsonEditor` no `onChange` either, so as far as I can tell the only way `properties.customProperties` ever gets written is the expand sub-drawer in `UsageChargeDrawerContent` — which we don't wire here.

The precondition is narrow (a rate card whose product uses a `custom_agg` metric, and those are API/Lago-provisioned), but for an org that does have one it looks fully reachable: nothing filters `custom_agg` out of the product picker, Custom isn't premium-gated, and the rate card drawer's "available models" alert even advertises it. If I'm reading it right, the editor would then render as a normal editable field, the typed JSON would never reach the form, and Save would just do nothing visible — `validateCustom` raises on `properties.customProperties`, but `CustomCharge` passes no `error` prop so there'd be no message anywhere. Editing an API-created Custom rate would be worse: it'd resubmit the original JSON with a success toast.

Worth noting this shape isn't new — `ChargeFilterDrawerContent` on main renders `ChargeWrapperSwitch` the same way, and the missing `error` prop on `CustomCharge` looks like its own follow-up. But since this drawer is new, maybe we could either thread `onExpandCustomCharge` through like `UsageChargeDrawerContent` does, or just filter Custom out of the model list until the flow is supported? wdyt?

src/pages/catalog/details/RateCardRateDetails.tsx
```diff
@@ -0,0 +1,260 @@
...
+          {
+            title: translate('text_1747314141347qq6rasuxisl'),
+            link: generatePath(RATE_CARD_RATE_DETAILS_ROUTE, {
+              rateCardId: rateCardId as string,
+              rateId: rateId as string,
+              tab: RateCardRateDetailsTabsOptionsEnum.activityLogs,
+            }),
+            // No table yet: activity logs are emitted against the parent rate card, and
+            // `ResourceTypeEnum` has no rate member yet, so there is nothing to scope to this
+            // rate. The tab says so rather than showing a blank panel. Wired up in BIL-594.
```
domenicofalco: nit: could we drop the `BIL-594` ref? ticket ids in code tend to go stale, and the PR body already carries it.

src/core/constants/statusRateCardRateMapping.ts
ansmonjol: Good question, but they already are: `Status` translates the label itself. `active`/`terminated`/`pending` are members of its `StatusLabel` union and it looks them up in `statusLabelMapping` (`Status.tsx`). All four sibling mappings (subscription, invoice, coupon, credit note) pass raw labels the same way.

src/pages/catalog/details/RateCardRateDetailsOverview.tsx
ansmonjol: Yes, intentional. `effectiveFrom` is a calendar day stored as a UTC instant and the backend floors it to UTC midnight, so reading it in an org timezone west of UTC would display the previous day. The picker is pinned to UTC for the same reason, so what you see here is the day that was picked.

src/pages/catalog/details/RateCardRateDetailsOverview.tsx
ansmonjol: Both gone in 0966b4d2c. The enum one is now a `toChargeModel()` lookup backed by an exhaustive `Record<RateCardRateModelEnum, ChargeModelEnum>`, so if the two enums ever diverge it stops compiling instead of casting silently. The `values` one turned out to be unnecessary altogether: the fragment is assignable to `Properties` as is.

src/pages/catalog/details/RateCardRatesTab.tsx
ansmonjol: Trimmed to one line in 0966b4d2c.

src/pages/catalog/dialogs/useDeleteRateCardRateDialog.tsx
ansmonjol: Cut down to a single line in 0966b4d2c. I kept a short note on why we evict *and* refetch, since doing both looks redundant at a glance (evict drops the row instantly, the refetch brings back the server-side counts).

src/pages/catalog/drawers/rateCardRate/constants.ts
ansmonjol: Same as above: it is a calendar day, not an instant. The backend floors `effectiveFrom` to UTC midnight, so reading it back in the org timezone would move the derived code and the error copy to the previous day for anything west of UTC.

src/pages/catalog/drawers/rateCardRate/constants.ts
ansmonjol: `values.rateModel` is gone: it goes through `toChargeModel()` now, an exhaustive enum-to-enum record, so a divergence between the two enums breaks the build. I left `values.properties as PropertiesZodInput` because it is the same call shape at every other caller (`planFormSchema.ts`, `ChargeFilterDrawerContent.tsx`) - worth fixing, but in one go across all of them rather than only here.

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
ansmonjol: Trimmed in 0966b4d2c - only the two that were not obvious from the name are left.

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
ansmonjol: Fixed properly: `ChargeModelSelector` only ever reads `localCharge.chargeModel`, so its prop is now `Pick<..., 'chargeModel'>`. A rate can pass just that field instead of faking a whole charge. No caller change needed.

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
ansmonjol: Same treatment across the component: `ChargeWrapperSwitch` and `SpendingMinimumOptionSection` are narrowed to the single field each actually reads, so all three synthetic `as unknown as LocalUsageChargeInput` casts are gone.

src/core/utils/escapeDoubleQuotes.ts
ansmonjol: Agreed, moved back to `~/core/utils/escapeDoubleQuotes.ts` in 0966b4d2c. Nothing about it is catalog-specific and that is where the other generic string helpers live.

src/pages/catalog/useRateCardRateTableActions.tsx
ansmonjol: Trimmed in 0966b4d2c.

src/pages/catalog/useRateCardTableColumns.tsx
ansmonjol: Shortened to two lines. I kept the gist because the wide fragment looks like over-fetching otherwise: Apollo replaces array fields wholesale, so selecting the narrower one here would strip range fields the rate pages cached.

src/pages/catalog/details/RateCardRateDetails.tsx
ansmonjol: Dropped in 0966b4d2c.

src/pages/catalog/details/RateCardRateDetails.tsx
ansmonjol: Fair - the banner comment above them was misleading, so I removed it (it also sat orphaned above a type declaration here). The constants themselves I kept: it is the pattern the whole catalog area already ships on main - 40 of the 48 `export const X = 'text_...'` in the repo are under `src/pages/catalog/` (`RateCardsList`, `RateCardDetails`, `RateCardDetailsOverview`, `RateCardDrawerContent`, `useRateCardDrawer`, `useDeleteRateCardDialog`, `useRateCardTableColumns`, `formatActiveRate`). Changing it in this PR alone would make the new files the inconsistent ones. Happy to drop it everywhere if you want, but as its own PR - want me to open one?

src/pages/catalog/details/RateCardRateDetailsOverview.tsx
ansmonjol: The test angle was just how the comment phrased it, and that comment is gone now. The actual reason they are exported is that the same copy is rendered by several surfaces (drawer, rates tab, row actions, details page), so they need a single home. See my reply on `RateCardRateDetails.tsx` for the wider convention question.

src/pages/catalog/details/RateCardRatesTab.tsx
ansmonjol: Same as the others: the banner comment is removed, the constants follow the existing catalog convention. Details in the `RateCardRateDetails.tsx` thread.

src/pages/catalog/drawers/rateCardRate/constants.ts
ansmonjol: These particular ones are the same label rendered in more than one place - the "Add a rate" / "Edit rate" button appears on the drawer, the rates tab, the row action menu and the details page. Giving each surface its own key would mean the same English string under several ids and translators translating it more than once. If the rule you have in mind is broader than that (no exported key constants at all), that is the same discussion as the `RateCardRateDetails.tsx` thread and I would rather settle it for the whole catalog area in one PR.

src/pages/catalog/drawers/rateCardRate/constants.ts
ansmonjol: Agreed, done in 0966b4d2c. `constants.ts` is constants only now, with `schema.ts`, `utils.ts` and `mapRateToFormValues.ts` beside it, each with its own unit test file.

src/pages/catalog/drawers/rateCardRate/useRateCardRateDrawer.tsx
ansmonjol: Done - `useRateCardRateForm` moved out into its own `useRateCardRateForm.tsx` with the gql document and the mutations, so `useRateCardRateDrawer.tsx` is just the drawer hook now (545 to ~145 lines).

src/pages/catalog/drawers/rateCardRate/useRateCardRateDrawer.tsx
ansmonjol: Done: `mapRateToFormValues` and `toFormProperties` are in `mapRateToFormValues.ts`, `deriveEffectiveFromBoundary` and the date/boundary helpers in `utils.ts`, both with their own unit tests.

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
ansmonjol: Went the other way in the end, see the thread below - the write path is wired now, so Custom stays available.

src/pages/catalog/drawers/rateCardRate/RateCardRateDrawerContent.tsx
ansmonjol: You read it right, and thanks for digging into it - I could reproduce exactly that. Fixed in 0966b4d2c by wiring it rather than filtering Custom out: the spec pins the model list to `getFixedChargeModelComboboxData()` / `getUsageChargeModelComboboxData()` verbatim, so dropping a model there would break that. Rather than copy the sub-drawer, I extracted it out of `UsageChargeDrawerContent` into a shared `useCustomChargeDrawer` hook and both drawers use it now. There is a test pinning that the callback reaches `ChargeWrapperSwitch` and that saving lands on `properties.customProperties`. The missing `error` prop on `CustomCharge` and the same gap in `ChargeFilterDrawerContent` I left alone - both pre-date this PR, agreed they are their own follow-up.

domenicofalco (approved):

PR #4200, opened 2026-08-28
fix(orders): send an explicit null when clearing the execution date

toommz:
## Context

Removing the execution date on an order had no effect: the save reported success and the date
was still there.

Clearing the date picker yields `undefined`, and `JSON.stringify` drops undefined-valued keys
from the request body, so `updateOrder` arrived with no `executeAt` key at all. The API reads an
absent key as "leave unchanged" and an explicit `null` as "clear", so the old value survived.

## Description

`buildUpdateOrderInput` now maps a cleared date to `null`.

`EditOrder` also handled no API failure at all, unlike the sign and execute pages: a rejection
resolves with `data: null` rather than throwing, and the handler had no `else`, so the user got
no inline error and no explanation. It now runs `getQuoteMutationErrors` and
`applyFormFieldErrors` and silences the codes it owns, matching `ExecuteOrder`.

Tests cover clearing, changing and rejecting the date. Reverting the one-line fix turns four of
them red.

### DIFF
--- a/src/pages/quotes/EditOrder.tsx
+++ b/src/pages/quotes/EditOrder.tsx
@@ -33,6 +33,11 @@ import {
   EditOrderFormValues,
   editOrderValidationSchema,
 } from './editOrder/validationSchema'
+import { applyFormFieldErrors } from './utils/applyFormFieldErrors'
+import {
+  getQuoteMutationErrors,
+  QUOTE_MUTATION_SILENT_ERROR_CODES,
+} from './utils/quoteMutationErrors'
 
 export const EDIT_ORDER_CLOSE_BUTTON_TEST_ID = 'edit-order-close-button'
 export const EDIT_ORDER_CANCEL_BUTTON_TEST_ID = 'edit-order-cancel-button'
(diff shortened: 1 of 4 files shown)


src/pages/quotes/editOrder/validationSchema.ts
```diff
@@ -33,5 +33,7 @@ export const buildUpdateOrderInput = (
 ): UpdateOrderInput => ({
   id: orderId,
   executionMode: values.executionMode,
-  executeAt: values.executeAt,
+  // An omitted key means "leave unchanged" server-side, so clearing the date must be sent as
+  // an explicit null.
```
domenicofalco: this comment can be removed. not needed

src/pages/quotes/EditOrder.tsx
```diff
@@ -96,6 +101,10 @@ const EditOrderFormContent = ({ order, loading }: EditOrderFormContentProps) =>
 
   const [updateOrderMutation] = useUpdateOrderMutation({
     refetchQueries: ['getOrders'],
+    // Handled locally by `getQuoteMutationErrors`, so the global error link must not
+    // also fire its generic toast (nor report these expected failures to Sentry).
+    // Copied: the link pushes its own force-silenced codes onto the array it receives.
```
domenicofalco: this comment can be removed. not needed

domenicofalco (approved):

PR #4201, opened 2026-08-28
docs(agents): apply flywheel feedback to the loop skills

ansmonjol:
Harvest of the nine proposals loop-run accumulated in `_flywheel.md` (BIL-501 run + PR #4193 review rounds). The skill that performs the harvest is #4202.

### Applied

| Proposal | Landed in |
| --- | --- |
| Fixture default switches off the branch under test | `.agents/docs/testing-practices.md` (new section) + pointers from `make-tests` and `loop-review` check 6 |
| Captured state goes stale (`MainHeader.Configure` snapshot key, drawer `open({ children })`) | `loop-build` implement step |
| `as unknown as` on a shared prop type / between twin enums | `loop-build` implement step, one bullet with both remedies (`Pick<...>`, exhaustive `Record<Source, Target>`) |
| Comment policy + pre-handoff deletion pass, no comment copied across files | `loop-build` implement step, now pointing at "Comments: Default to None" in the TS conventions |
| Comment budget as a review dimension | `loop-review` check 8, replacing the older redundant-comment rule |
| Optional `on*` props omitted on a reused form component = silent read-only | `loop-review` new check 12 |
| Calendar day vs instant, write zone and display zone pinned together | `loop-spec` explore step |

Two proposals about `as unknown as` merged into one rule; the "don't repeat a comment in N files" proposal folded into the comment bullet rather than added beside it.

The fixture rule was relocated: it was proposed for `loop-build`, but loop-build forbids hand-writing tests, so the rule lives where tests are authored and every session gets it, not just the loop.

The comment rules reference the "Comments: Default to None" section added in #4193, which is approved and mergeable.

`.agents/**` is outside every gate (`pnpm lint` covers `src/`, `cypress/` and `index.html`), so there is nothing to run here.


ansmonjol:
Superseded: these skill and doc edits belong on the PR whose runs produced the flywheel entries, so they moved to #4193. The harvest skill itself stays in #4202.

PR #4211, opened 2026-08-31
fix(auth): polish the Microsoft Entra ID login and settings surfaces

domenicofalco:
## Context

Three UI/wording returns on the Microsoft Entra ID SSO integration, grouped under LAGO-1790:

- The Entra logo on the login page reads far smaller than the Google and Okta ones next to it. It is also used in settings > team & security > Authentication, where it is fine and must stay unchanged.
- The Entra card's action menu wording is too long: "Edit Microsoft Entra ID connection" / "Delete Microsoft Entra ID connection" should read "Edit connection" / "Delete connection".
- On the Authentication settings page the SSO cards should read Password Login / Google / Microsoft Entra ID / Okta, and the provider should be named "Microsoft Entra ID" rather than "Entra ID".

This is a single PR closing the parent ticket and all three of its sub-issues (LAGO-1772, LAGO-1774, LAGO-1775) — they touch the same two files and splitting them would produce three one-line PRs.

## Description

**Logo sizing (LAGO-1772):** both SSO login surfaces rendered `microsoft-entra-id.svg`, a 40x40 rounded white plate whose Microsoft mark only occupies the inner 20x20, scaled down to 16px — so the mark itself painted at roughly 8px while the sibling `google` and `okta` glyphs fill their whole 16px box. Added a bare `microsoft` design-system icon (the same four squares, same colors, on a 16x16 viewBox) and switched both buttons to `startIcon="microsoft"`:

- `src/pages/auth/Login.tsx` — the login page, which the ticket reports.
- `src/pages/Invitation.tsx` — the invitation-accept page, same pattern, caught in pre-review. Beyond the sizing, using an inline child left `startIcon` unset, so `Button` routed this button's `loading` spinner to the **endIcon** slot (`Button.tsx:199-205`) and kept the squashed logo on the left; it now spins in place like the Okta button beside it.

The two remaining consumers of the plated asset — `Authentication.tsx` and `EntraIdAuthenticationDetails.tsx` — are deliberately untouched: both render it unconstrained inside a `connector-full` Avatar, which is the settings usage the ticket calls out as already fine.

**Menu wording (LAGO-1774):** the Entra entry now points at the existing `Edit connection` / `Delete connection` keys already used by the SSO details page, instead of the provider-suffixed ones. `text_1784307344255fc26gfvrmb5` stays as-is because it is also the `AddEntraIdDialog` edit title, where the provider name is still wanted; `text_1785339647668vsn4wyo0j7n` had no other consumer and is removed. Okta keeps its own labels — its name is short enough to fit, which is why only the Entra menu was reported.

**Order and naming (LAGO-1775):** the Entra entry moves ahead of Okta in `ssoProviders`, and `text_17843073442548zt904xoinv` becomes "Microsoft Entra ID" — it is the provider name everywhere it is used (selector title, details view name, add/delete dialogs, the login-method-not-authorized alert), so all of those read consistently now.

**Tests:** added coverage for the card order, the provider name, the shortened menu labels, and the icon on both SSO buttons. The Okta tests that reached for "the third selector" / "the third dots button" by index now use the stable per-provider test ids, so they no longer depend on the card order.

**Note for review:** the reference asset attached to LAGO-1772 is hosted on `uploads.linear.app`, which this sandbox's proxy blocks, so the new icon is derived from the in-repo SVG's own geometry and brand colors rather than from the Figma file. Worth an eyeball against the design.

Fixes LAGO-1790


domenicofalco:
Verified and fixed in `4a2120a`.

The finding was correct — `src/pages/Invitation.tsx:437` had the identical `<MicrosoftEntraId className="mr-2 size-4" />` pattern, sitting between a Google button and an `startIcon="okta"` button, so its mark painted at roughly half their size. Switched to `startIcon="microsoft"` and dropped the now-unused import.

Worth noting a second effect the pre-review didn't mention: as an inline child the logo left `startIcon` unset, so `Button` routed the `loading` spinner to the **endIcon** slot (`Button.tsx:199-205`) and kept the squashed logo on the left. It now spins in place, matching the Okta button beside it.

Added `should render the Microsoft logo at the same size as the other SSO buttons` to `Invitation.test.tsx`, mirroring the Login one.

The two remaining consumers of the plated asset — `Authentication.tsx:150` and `EntraIdAuthenticationDetails.tsx:72` — are deliberately left alone: both render it unconstrained inside a `connector-full` Avatar, which is the settings usage LAGO-1772 explicitly calls out as already fine.


---

ansmonjol (approved):

PR #4213, opened 2026-08-31
refactor(customer-portal): migrate wallet top-up form to TanStack Form

ansmonjol:
## Context

Part of the ongoing Formik -> TanStack Form migration (LAGO-1148). The customer-portal wallet top-up page was one of the remaining Formik + Yup forms in the app.

## Description

Migrated `src/components/customerPortal/wallet/WalletPage.tsx` from `useFormik` + Yup to `useAppForm` + Zod, following the established migration conventions:

- New `validationSchema.ts` ports the Yup `amount: number().required('')` rule to Zod. The Formik-era `amount <= 0` check (previously only enforced by the submit-button's external disabled logic) is folded into the same Zod refine, so the observable gating behavior is unchanged.
- `amount` uses `field.AmountInputField` under `form.AppField`; the business-rule bounds error from `topUpAmountError` is still wired through `errorOverride` / the submit button's `disabled` prop, exactly as before.
- `onCompleted` now calls `form.reset()` instead of `formikProps.resetForm()`.
- The mutation call inside `onSubmit` is now awaited, so the submit spinner (driven by `form.SubmitButton`'s `isSubmitting`) correctly covers the network request.
- Added a `<form onSubmit>` wrapper (this page didn't have one), which also means pressing Enter while focused in the amount field now submits the form, consistent with the rest of the app's TanStack forms.
- Added `src/components/customerPortal/wallet/__tests__/WalletPage.test.tsx` covering the loading/error states, rendering, submit, and the bounds-error disabled state.

No visual or UX changes intended beyond the Enter-to-submit addition.

Fixes LAGO-1221

### DIFF
--- a/src/components/customerPortal/wallet/validationSchema.ts
+++ b/src/components/customerPortal/wallet/validationSchema.ts
@@ -1,0 +1,22 @@
+import { z } from 'zod'
+
+// The amount field is invisible-error by design (rendered with displayErrorText={false} —
+// the visible label comes from topUpAmountError instead), so only the message KEY matters.
+export const walletPageTopUpErrorLabels = {
+  required: 'text_1771342994699klxu2paz7g8',
+} as const
+
+// `AmountInputField`'s `beforeChangeFormatter={['positiveNumber']}` stores `number | ''` at
+// runtime (parseFloat result, or '' when emptied) — see migrate-formik-to-tanstack Pattern 11.
+// The original Yup schema was only `number().required('')`; combined with the Formik-era
+// `amount <= 0` check baked into the submit-button disabled logic, the actual gate was
+// "a positive number", so both collapse into one refine here.
+export const walletPageTopUpValidationSchema = z.object({
+  amount: z
+    .union([z.number(), z.literal('')])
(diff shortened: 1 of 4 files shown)


src/components/customerPortal/wallet/validationSchema.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from 'zod'
+
+// The amount field is invisible-error by design (rendered with displayErrorText={false} —
+// the visible label comes from topUpAmountError instead), so only the message KEY matters.
+export const walletPageTopUpErrorLabels = {
+  required: 'text_1771342994699klxu2paz7g8',
+} as const
```
domenicofalco: is that assertion needed?

domenicofalco (approved):

src/components/customerPortal/wallet/validationSchema.ts
ansmonjol: Good catch, no. Nothing relied on the literal type. I removed it in 0162a56fc and replaced the exported labels object with a private message constant.

PR #4215, opened 2026-08-31
chore(docs): extract narrow-scope CLAUDE.md sections into auto-triggering skills

domenicofalco:
## Context

`CLAUDE.md` is loaded into every agent session, and 76% of it (385 of 508 lines) sat in
four sections that only matter while touching one subsystem: Pagination, Drawers, Dialogs
and Organization slug architecture. They inflated the context cost of every session and
diluted the signal of the genuinely universal rules.

A "read this doc on demand" pointer is unreliable exactly when it matters most: when the
task does not name the topic. "Create a new template path" does not read as an org-slug
task, but it is one. A skill description, by contrast, is surfaced every turn, so an
explicit TRIGGER clause fires on the signals themselves rather than on the agent noticing
a pointer.

## Description

- Move the four sections into dedicated skills under `.agents/skills/`: `lago-pagination`,
  `lago-drawers`, `lago-dialogs`, `lago-organization-slug`. Each carries the full detail —
  tables, code patterns, reference sites — transplanted verbatim, with heading levels
  shifted up one. No rule text changed.
- Each skill description ends with a `TRIGGER — ...` clause naming the concrete signals
  that should surface it: symbols, file paths, GraphQL shapes, task phrasings and reported
  symptoms. The org-slug trigger explicitly covers persistent identifiers that embed an org
  (a copied URL, a webhook or template path, an LS key, a filename, a mutation argument).
- Replace the four sections in `CLAUDE.md` with a short pointer block naming each skill, and
  add a "Maintaining this file" section recording the extraction rule: narrow scope + more
  than ~20-25 lines + concretely detectable triggers.
- Update `babysit` and `triage-frontend-ticket`, which asserted that `CLAUDE.md` itself
  holds the drawer, dialog, pagination and organization-slug rules.

`CLAUDE.md` goes from 508 to 154 lines. Every section that stays is byte-identical.

Fixes STA-12

**Follow-up candidate:** the `## Cypress e2e tests` section (27 lines) fits the same extraction rule but was left inline — the ticket names exactly four sections.


domenicofalco:
Not applied. Angle brackets in a skill description are not rejected: six skills already tracked in this repo carry them (`loop-build` — `"/loop-build <ISSUE-ID>"`, `loop-review`, `loop-revise`, `loop-run`, `loop-spec`, `triage-frontend-ticket`) and load fine, and there is no skill validator in the repo or in CI. `<Feature>` here is the symbol name `use<Feature>Drawer` as written throughout `CLAUDE.md`; replacing it would make the description less precise for a constraint that does not exist. Happy to revisit if you can point at the validator that rejects it.

ansmonjol (approved):

PR #4222, opened 2026-09-01
docs(agents): add qa-session skill

domenicofalco:
## Context

The loop pipeline covers spec, build, review and ship, but browser
verification of a change was still improvised on every run: steps guessed
instead of read from the routes and translations, no control run on main,
and verdicts that proved the diff rather than the reported symptom. The
skill existed only in one local checkout, so nobody else could use it.

## Description

Adds `.agents/skills/qa-session/SKILL.md`, the browser-QA phase of the loop
pipeline:

- two modes (`manual` steps for the operator, `auto` driving Chrome) and a
  target (ISSUE-ID, PR, branch or surface), both required
- resolving the app under test from the loop state dir, `.worktree-slots` or
  `docker ps`, including the vite-cache reset for stale containers
- a test plan written in UI language only, whose first check is always the
  reporter's own path, with feature-flag, permission and data preconditions
- a mandatory control run on main with two separate controls, the acceptance
  criteria and the reported repro, so a PASS cannot come from a fix that
  addresses a different defect
- verification rules for visibility over presence and scroll as a test
  dimension, a live triage table for common anomalies, and a verdict block
  written last

State-dir paths use `$LOOP_STATE_DIR` like the other loop skills instead of
a hardcoded home path.

### DIFF
--- a/.agents/skills/qa-session/SKILL.md
+++ b/.agents/skills/qa-session/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: qa-session
-description: Use when the operator wants to verify in the browser that a change works — asks "come lo testo in locale?", "dammi gli step", "QA this", references a worktree/PR/ticket built in this session, or reports something not working during manual testing (button does nothing, section missing, block blank).
+description: Use when the operator wants to verify in the browser that a change works — asks "how do I test this locally?", "give me the steps", "QA this", references a worktree/PR/ticket built in this session, or reports something not working during manual testing (button does nothing, section missing, block blank).
 ---
 
 # QA Session — browser verification of a change
@@ -42,6 +42,12 @@ diff and not the other way round. Enumerate the ticket's attachments (`get_issue
 `list_comments`): a video, screenshot or Slack thread often carries the only repro there is. You cannot watch
 a video — say so explicitly, and ask the operator for the gesture and the surface rather than substituting a
 plausible one.
+
+**No ticket (PR, branch or free-text target)**: the report is whatever the operator handed over — PR body,
+their own message. Ask for the exact gesture, payload and surface when it is not there; never substitute a
+plausible one. Everywhere below, "ticket" and "spec" mean that source instead, and with no acceptance
+criteria the AC control (Step 4a) collapses into the report control — say so rather than deriving ACs from
+the diff.
 
 Then check what gates the surface:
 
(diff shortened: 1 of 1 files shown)


domenicofalco:
Both HOLD points applied in 5f5d74e.

1. Non-ticket targets are now a real flow: Step 2 gained a no-ticket branch (the operator's PR body or message is the report, missing repro details get asked for, and with no acceptance criteria the AC control collapses into the report control instead of being derived from the diff), and Step 7 stores results in `$LOOP_STATE_DIR/qa/<branch-or-PR>.md` when there is no ISSUE-ID.
2. The control run is diff-aware: it dumps the branch diff to a patch, reverts with `git apply -R`, then re-applies the same patch forward to restore. That is symmetric for adds, deletes and renames, which `git checkout main -- <paths>` was not — verified on a scratch repo with all three cases, tree clean afterwards.

Also unrelated to the review: the description's trigger phrases were in Italian, now English like the rest of the file.

domenicofalco:
All three HOLD points applied in 7bba6b8.

1. **Step 1 target resolution** — a PR number or URL now resolves through its head branch to the state dir whose `branch:` matches, the same path `loop-revise` takes. Branch names are normalized (`/` → `-`) before a slot lookup, since that is the key `lago-worktree` actually registers (`name="$(echo "$branch" | sed 's|/|-|g')"`), so a slashed branch no longer misses its own slot. The slot registry is the monorepo root's `.worktree-slots` (`$LAGO_PATH/.worktree-slots`, i.e. `../.worktree-slots` from the `front/` checkout) rather than a `lago/`-prefixed guess, and the container line now covers the API side: `lago_api_wt_${SAN}`, or `lago_api_dev` when the slot has no `api_port` and the worktree proxies the shared stack.

2. **Step 2 feature flag** — enabled on the organization whose slug is in the URL under test via `Organization.find_by!(slug:)`, in the API container resolved in Step 1. Confirmed `Organization.slug` is a non-null field in the schema, so the URL slug is a sound key; `Organization.first` was an arbitrary row and could flip the flag on an org nobody was testing while leaving the tested route gated.

3. **Step 4 control run** — the patch is generated with `git diff --binary`, so a diff carrying an image or a font re-applies instead of failing, and the forward re-apply is now stated as mandatory on every exit path: the check reproducing, not reproducing, erroring, the browser leg dying mid-run in `auto` mode, and the Step 4 stop when the report control does not reproduce. That last one was the real hole — it was an early exit with the patch still reverted, leaving the branch on main's code under a dirty tree that reads as the operator's own uncommitted edits.

Gates: `pnpm run code:style` exits 0 with zero errors (the 57 warnings are pre-existing formik deprecations elsewhere in the tree). Markdown-only change, so no tests apply.

---

ansmonjol (approved):

PR #4227, opened 2026-09-02
fix(invoices): handle a stale draft-invoice fee with an actionable error

domenicofalco:
## Context

Sentry reported `createAdjustedFee` returning a 404 with `{"fee":["not_found"]}` while a user was editing a fee on a draft invoice.

Draft invoice fees are destroyed and recreated server-side when the draft is refreshed, which changes their ids. A row rendered before such a refresh still holds the old id, so opening it and submitting sends a fee id that no longer exists — the backend 404 is correct, the front-end is acting on stale data.

The error was not silenced, so the global Apollo error link fired the generic "an error occurred" toast, reported to Sentry, and left the drawer open with the stale `feeId` still in the form. Every retry failed the same way.

## Description

`EditFeeDrawer` now handles the not-found case itself:

- `context.silentErrorCodes` lists `NotFound`, so the global error link stops toasting it generically and stops reporting it to Sentry.
- `onError` tells the stale-fee case apart with `hasDefinedGQLError('NotFound', error, 'fee')`, guarded on edit mode since no `feeId` is submitted otherwise. It shows a dedicated message, closes the drawer, resets the form so the stale id is dropped, and refetches `getInvoiceDetails` / `getInvoiceFees` so the rows come back with the new ids.
- Any other error on the mutation keeps the generic toast, so nothing else changes behaviour.

Adds a test id on the drawer's submit button and a test file covering the toast, the drawer close, the refetch, and the fallback for unrelated errors.

Fixes LAGO-1868

---
_Generated by [Claude Code](https://claude.ai/code/session_01M4xZj5kLpJAPxecsH5qHHT)_

### DIFF
--- a/src/components/invoices/details/EditFeeDrawer.tsx
+++ b/src/components/invoices/details/EditFeeDrawer.tsx
@@ -1,4 +1,4 @@
-import { gql } from '@apollo/client'
+import { gql, useApolloClient } from '@apollo/client'
 import InputAdornment from '@mui/material/InputAdornment'
 import { useFormik } from 'formik'
 import { tw } from 'lago-design-system'
@@ -12,6 +12,7 @@ import { Skeleton } from '~/components/designSystem/Skeleton'
 import { Typography } from '~/components/designSystem/Typography'
 import { AmountInputField, ComboBox, ComboBoxField, TextInputField } from '~/components/form'
 import { DrawerLayout } from '~/components/layouts/Drawer'
+import { addToast, hasDefinedGQLError } from '~/core/apolloClient'
 import { ALL_CHARGE_MODELS, ALL_FILTER_VALUES } from '~/core/constants/form'
 import { TExtendedRemainingFee } from '~/core/formats/formatInvoiceItemsMap'
 import { getCurrencySymbol, intlFormatNumber } from '~/core/formats/intlFormatNumber'
@@ -24,6 +25,7 @@ import {
   FeeForCreateFeeDrawerFragment,
   FixedCharge,
   FixedChargeChargeModelEnum,
+  LagoApiError,
   useCreateAdjustedFeeMutation,
   useGetInvoiceDetailsForCreateFeeDrawerQuery,
 } from '~/generated/graphql'
@@ -32,6 +34,7 @@ import { OnRegeneratedFeeAdd } from '~/pages/CustomerInvoiceRegenerate'
 
 import { InvoiceTableSection } from './InvoiceDetailsTable'
 import { InvoiceDetailsTableBodyLine } from './InvoiceDetailsTableBodyLine'
+import { EDIT_FEE_DRAWER_SUBMIT_BUTTON_TEST_ID } from './invoiceDetailsTestIds'
 import {
   getChargesComboboxDataFromInvoiceSubscription,
   getChargesFiltersComboboxDataFromInvoiceSubscription,
@@ -206,6 +209,7 @@ type formikValues = Omit<Partial<CreateAdjustedFeeInput>, 'feeId'> & {
 
 export const EditFeeDrawer = forwardRef<EditFeeDrawerRef>((_, ref) => {
   const { translate } = useInternationalization()
+  const client = useApolloClient()
   const drawerRef = useRef<DrawerRef>(null)
   const [localData, setLocalData] = useState<EditFeeDrawerProps | undefined>(undefined)
   const isRegenerateMode = localData?.mode === 'regenerate'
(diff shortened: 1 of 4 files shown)


domenicofalco:
### `Run Test E2E` is failing on an API-side step, not on this diff

**What is failing:** step 11, `Populate db` — `docker compose ... exec api bundle exec rails db:seed`. It ran 06:44:53Z → 06:44:55Z (2s) and aborted with Rails' pending-migrations error, which dumps the full migration list then exits 1. Steps 12–14 (`Cypress install`, `Cypress run`) were all **skipped**, so no spec in this repo ever executed.

**Why it is not this PR's:**

- The failing command is a Rails task inside the `api` container, checked out from `getlago/lago-api` at its own HEAD. This PR changes four front-end files — `EditFeeDrawer.tsx`, `invoiceDetailsTestIds.ts`, one Jest test, and one translation key — none of which can affect whether the API container's database has its migrations applied.
- `Build Front local image` succeeded (06:39:06Z → 06:41:04Z), so the branch compiles and bundles; the failure is strictly downstream of the front-end build.
- Cypress never started, so this is not a failing e2e assertion.
- Every other check on this head is green: `Run linters`, `Typecheck`, all four `Tests` shards, `Run Codegen`, CodeQL, and the SonarQube quality gate (0 new issues, 100% coverage on new code).

**No fix exists to port.** The breakage is in the API repo's migration state, not in anything this PR or any front-end PR can carry. I am re-running the job once to confirm it reproduces identically; if it does, this needs someone with API-side context (or a later `lago-api` HEAD) rather than a change here.

---

domenicofalco:
Correcting my comment above: the failure did **not** reproduce on the re-run, so it was transient rather than a persistent API-side breakage as I predicted.

On attempt 2, `Populate db` succeeded in 7s (06:52:03Z → 06:52:10Z) against the same head SHA and the same `lago-api` checkout, and `Run Test E2E` went on to pass in full (06:46:59Z → 06:57:20Z, 10m21s). The likeliest reading is a race between the API container finishing its migrations and `rails db:seed` firing — attempt 1 reached `db:seed` about 30s after `Launch APP + API` returned, attempt 2 about 20s later in wall-clock terms.

The part that held: the failure was never in this diff. It aborted before Cypress started, in a Rails task inside the API container.

All 14 checks are now green on `bed610f`. No action needed here.

---

domenicofalco:
Thanks — fixed in `4ecd3ef`, though the premise needs one correction.

**On the two-toasts claim:** both handlers did run, but the user only ever saw one toast. `addToast` drops a toast whose `translateKey` already sits in `toastsVar` ([`toastVar.ts#L39-L44`](https://github.com/getlago/lago-front/blob/main/src/core/apolloClient/reactiveVars/toastVar.ts#L39-L44)), and the drawer's fallback used `text_622f7a3dc32ce100c46a5154` — the exact key the global error link toasts with. Same key, so the second call was a no-op.

**Why I changed it anyway:** the handler's correctness rested on that dedupe rather than on its own logic, which is a fair thing to flag. `onError` now returns early unless the error is the `not_found` it actually silenced:

```tsx
onError(error) {
  if (!hasDefinedGQLError('NotFound', error)) return
  ...
}
```

So the drawer only handles what it took ownership of, and every other code is left to the error link that still toasts *and* reports it to Sentry. No behavioural change for users; the implicit dependency is gone.

**On the test ask:** rather than a one-toast assertion (which would have asserted the dedupe), the suite now asserts the stronger property — for an unsilenced code the drawer does not toast **at all**, covering both a bare `unprocessable_entity` and one carrying details. 11 tests, all green.

---

ansmonjol (approved):

domenicofalco:
> Is there a way to make sure the drawer opens

Good catch, and it was a real hole — thanks. Fixed in `4d9a9d5`.

Nothing proved the drawer had opened, and the stale-fee case asserts its **absence**. So a drawer that silently failed to open would have satisfied `not.toBeInTheDocument()` without exercising the close path at all — a passing test proving nothing.

`renderAndOpenDrawer` now asserts the drawer body is present right after `openDrawer`, which guards the precondition for every case rather than just the close one. The add-mode test shares the helper instead of opening inline, and there's an explicit case naming the behaviour.

I checked the guard actually bites rather than assuming it: stubbing out the `openDrawer` call turns **all 12 tests red**, including `THEN should close the drawer instead of leaving it on the stale fee` — which passed before the change on a drawer that never opened.

```
✕ THEN should render the drawer body
✕ THEN should close the drawer instead of leaving it on the stale fee
✕ THEN should keep the drawer open so the user can retry
... 12 failed, 12 total
```

---

PR #4235, opened 2026-09-03
feat(connections): Add a code field to the connection drawers

domenicofalco:
## Context

A customer connection carries a `code`, the routing key that will let a customer
hold several connections of the same type and let billing objects point at one of
them. The backend already exposes it on every read and accepts it on every write —
the customer arrays (`paymentProviderCustomers` / `integrationCustomers`) and the
four dedicated per-connection mutations — and derives a default from the provider
code when it is omitted. Until now the front-end never surfaced it, so the value
was invisible and unmanageable.

The field ships ungated, ahead of the multi-connection-per-type work: it is
non-mandatory, the backend keeps deriving a default, so nothing changes for users
who ignore it.

## Description

The shared connection drawer (used by both the customer create/edit form and the
customer-information master-detail) gets an optional, editable `code` input,
placed under the connection selector and above the mapping fields.

- The input mirrors the code of the selected connection, so what the drawer shows
  is what gets stored. It follows a provider switch, comes back to the code the
  drawer opened on when the original provider is reselected, and never overwrites
  a code the user typed.
- The reserved `lago_manual` code is rejected: it identifies the manual payment
  row, and a connection saved under it would be dropped from the customer payload.
- Both persistence strategies carry the value: the customer create/edit form
  writes it into `paymentProviderCustomers[].code` / `integrationCustomers[].code`,
  and the customer-information view sends it on the create/update payment and
  integration connection mutations. An empty input never blanks a persisted code.
- The backend uniqueness rejection is surfaced where the user can act on it: under
  the Code input, with the drawer kept open, and cleared as soon as the code is
  edited. On the customer form, where the drawer is already closed by save time,
  it is reported on a toast — and the pre-existing `ValueAlreadyExist` handling is
  now scoped per field, so a duplicate connection code no longer blames the
  customer external id nor resets the form silently.
- `clearExistingCodeError` joins `applyExistingCodeError` in `src/core/form` so the
  guard-and-clear pair lives in one place.

Fixes ING-440


src/components/customerConnections/ConnectionCodeField.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const CONNECTION_CODE_FIELD_TEST_ID = 'connection-code-field'
+
+type ConnectionCodeFieldProps = {
+  form: CustomerConnectionDrawerFormApi
+}
+
+// Never pre-filled from the selected provider: an empty code is submitted as
+// null and `PaymentProviderCustomers/IntegrationCustomers::BaseCustomer#set_code`
+// backfills it from the provider, so the value shown always comes from the DB.
+export const ConnectionCodeField = ({ form }: ConnectionCodeFieldProps) => {
```
ansmonjol: we jsut use it once? was it worth creating a new component for it? maybe you plan to re-use it later?

domenicofalco: Good catch, thanks! You're right, and no reuse is coming: the call site can only ever be one, since `CustomerConnectionDrawer` is itself the piece shared between the customer create/edit form and the customer-information view, so the multi-connection-per-type work won't add a second mount either. Inlined into the drawer in 1db0b2a9.

For contrast, `ChargeCodeField` does earn its own file — the fixed- and usage-charge drawers both mount it. Here the drawer test already owned the field's behaviour, so the three cases that were unique to `ConnectionCodeField.test.tsx` (no code derived from the provider on switch, in both create and edit, plus the `clearExistingCodeError` wiring) moved into `CustomerConnectionDrawer.test.tsx` rather than being dropped.

---

ansmonjol (approved):

PR #4240, opened 2026-09-04
chore(skills): add a lago-forms skill for building new TanStack forms

domenicofalco:
## Context

Nothing documents how to write a *new* TanStack form. `migrate-formik-to-tanstack`
holds most of the conventions, but its trigger only fires on converting an existing
Formik form, so a "build me a form" task never reaches them.

That cost two rounds of human correction during ING-626 / ING-634, where a date field
was built to invalidate the form before the first submit — contradicting the
submit-first convention that is settled in this codebase but written down nowhere an
agent would look.

## Description

Adds `.agents/skills/lago-forms/SKILL.md`, sibling of `lago-dialogs`, `lago-drawers`,
`lago-pagination` and `lago-organization-slug`, with a `TRIGGER` clause naming the
concrete signals (`useAppForm`, `form.AppField`, `revalidateLogic`, `validationLogic`,
`SubmitButton`, `validationSchema.ts`, adding a form or a field) plus the symptoms that
should send a reader to it — submit stuck disabled with no error, a schema that never
fires, an error before the first submit, a raw "Invalid input" in a field.

A consolidation, not a design: every rule already existed in the migration skill, the
sibling skills or the code. What it captures:

- the skeleton — schema file (schema + `z.infer` type + defaults + form→API mapper) and
  the component wiring it, `<form id>` included
- the non-negotiables, each with the failure it prevents: always pass
  `validationLogic: revalidateLogic()` (omitted, `defaultValidationLogic` never runs
  `onDynamic` and the schema is silently skipped); submit-first, so the form is never
  invalid before the first submit attempt; a field publishes its value and the schema
  decides whether it is acceptable (the ING-634 rule, post-#4220); bare
  `<form.SubmitButton>` with no `isDirty` gate; `await` inside `onSubmit`; `useStore`
  for anything read in the render
- the stored value shape of all eleven registered field components, since a schema
  written against the API payload instead fails silently — the BIL-410 class of bug
- schema location and `zodCustoms` reuse, translation-key messages and the
  `DEFAULT_ZOD_ERROR_MESSAGE` fallback, `addUnsupportedDateIssue` for date floors
- the two non-interchangeable server-error shapes (`setFieldMeta` vs `setErrorMap`)
- `withForm` vs `withFieldGroup`, and how a form is wired into a dialog

Cross-references rather than duplicates: `lago-drawers` / `lago-dialogs` keep the
surrounding hook and the `import.meta` mock, `make-tests` keeps the suite, and
`typescript-conventions.md` keeps the style rules. `migrate-formik-to-tanstack` gains a
one-line pointer so greenfield work is redirected here. `CLAUDE.md` gains the 3-line
pointer the other subsystem skills have.

Each load-bearing claim was checked against source rather than carried over on trust:
`revalidateLogic`'s `submit` / `change` defaults and `defaultValidationLogic`'s
validator set in `@tanstack/form-core@1.33.0`, the `useFieldContext<T>` of every
`*FieldForTanstack` wrapper, `AmountInput` pushing `int` at `getCurrencyPrecision === 0`,
and the `z.looseObject` schemas backing multi-selects.

### Tests

Documentation only — no `src` file changes, so the coverage map is empty and no test is
added or needed.

`pnpm run code:style`: 0 errors, 56 warnings, unchanged from `main` (all pre-existing
`formik` deprecation warnings in untouched files).

Fixes ING-635

---
_Generated by [Claude Code](https://claude.ai/code/session_01ES5C9zxqZnCi1W1vLVBFRg)_

### DIFF
--- a/.agents/skills/lago-forms/SKILL.md
+++ b/.agents/skills/lago-forms/SKILL.md
@@ -92,7 +92,7 @@ with no error and no request (the BIL-410 regression).
 
 | Component | Stored value |
 | --------- | ------------ |
-| `TextInputField` | `string` — **unless** `beforeChangeFormatter` includes `'int'`, which stores `number \| ''` |
+| `TextInputField` | `string` — **unless** `beforeChangeFormatter` includes `'int'`, which stores `number \| ''`, plus a transient `'-'` while a negative number is being typed (pair it with `'positiveNumber'`, which strips the sign first, to rule that out) |
 | `AmountInputField` | `string \| number \| undefined` — `'int'` is auto-pushed for 0-decimal currencies (JPY, KRW…), so an amount flips shape with the currency |
 | `ComboBoxField` | `string \| undefined` |
 | `MultipleComboBoxField` | whole options, `{ value, label, … }[]` — schema `z.array(z.looseObject({ value: z.string() }))`, map to ids in `onSubmit`, seed defaults as options |
@@ -159,9 +159,13 @@ formApi.setErrorMap({
 })
 ```
 
-Clear the error when the user edits the field, or submit stays disabled. A duplicate
-`code` rejection has a helper that does both: `applyExistingCodeError(formApi)`
-(`~/core/form/existingCodeError`).
+Clear the error when the user edits the field, or submit stays disabled — setting and
+clearing are two separate jobs. For a duplicate `code`, `applyExistingCodeError(formApi)`
+(`~/core/form/existingCodeError`) does only the setting half, in the mutation catch on
+`LagoApiError.ValueAlreadyExist`. The clearing half belongs to the field, gated on
+`EXISTING_CODE_ERROR_MESSAGE` so it does not wipe the zod required-check — already
+implemented in `NameAndCodeGroup` and `ChargeCodeField`, so mount one of those rather
+than re-wiring the listener.
 
 Validator-produced errors live directly on `errorMap.onDynamic` keyed by field path; the
 `.fields` sub-shape exists only for errors set manually. Reading section validity for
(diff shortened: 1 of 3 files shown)


domenicofalco:
All four pre-review findings verified against source and fixed in 86523f3.

- **`withFieldGroup` call contract** — correct, and worth having caught. A group is mounted with `form={form}` + a `fields` mapping (`PlanSettingsSection.tsx:112`, `NameAndCodeGroup.test.tsx:46`); `group` is only what the group's own `render` receives. Also fixed the same wrong signature in the two places `migrate-formik-to-tanstack` carried it — the one file outside the stated scope I touched, since leaving it would contradict the correction in the doc this guide defers to.
- **Duplicate-code errors** — correct. `applyExistingCodeError` only sets; clearing-on-edit is a field-level `onChange` listener gated on `EXISTING_CODE_ERROR_MESSAGE`, implemented in `NameAndCodeGroup` and `ChargeCodeField`. Reworded to say setting and clearing are two jobs, and to mount one of those two rather than re-wire the listener.
- **`EditFeeBillingPeriod.tsx`** — correct: raw `DatePicker` inside `form.AppField` plus `message: ''`. Took the "explicitly disqualify" option: the reference is now `EditInvoiceItemDescriptionDialog.tsx` (registered field + `dialogFormProps`), and `EditFeeBillingPeriod` is kept as a named counter-example so the next reader doesn't pick it up on their own.
- **`TextInputField` + bare `int`** — correct. `formatValue` passes a lone `'-'` through instead of `parseInt`-ing it (`TextInput.tsx:85`, pinned by `FormatValue.test.ts:47`). Documented, with `'positiveNumber'` as the way to rule it out.

`pnpm run code:style`: 0 errors, 56 warnings, unchanged from `main`. CI re-running on the new head.

---

ansmonjol (approved):

PR #4243, opened 2026-09-04
feat(connections): Add the default flow to the customer connections

domenicofalco:
## Context

Customer connections were rebuilt in Milestone 1 as a list plus a per-type
drawer, but with no notion of a default: the information tab rendered an empty
Status column as a placeholder and nothing surfaced which connection a billing
object would actually route to. The backend side is ready — the two
set-as-default mutations shipped on api/main, and the connection read now
carries `isDefault` with the first connection of a type auto-becoming default.

This is the Milestone 2 default flow for provider and integration connections,
gated behind the `multi_connection` feature flag: with the flag off, behaviour
is unchanged. The manual payment connection is deliberately out of scope — it
has its own gate and timeline.

## Description

- The shared connections list gains a `Default` badge in its Status column and
  a `Set as default` entry in the row "…" menu, between Edit and Delete. The
  entry is disabled, not hidden, when the row already is the default or has no
  persisted connection to address.
- `useSetConnectionAsDefault` is a new form-agnostic hook owning the two
  dedicated mutations, so both connection surfaces share one writer. The
  customer information tab composes it and follows a successful write with its
  silent customer refresh; the customer create/edit form calls it directly and
  mirrors the flag into its form arrays without dirtying the form, since the
  change is already persisted.
- The connection settings panel reports `Default on customer`, omitted for a
  payment row that carries no provider-customer record at all — asserting "No"
  there would claim a healthy connection is not the default.
- The connection drawer shows the same `Default` badge on its locked selector.
- `isDefault` is added to the six integration inline fragments and read back
  into the create/edit form model. It is never sent through the customer
  arrays: the dedicated mutations are the only writers of default state.

Fixes ING-441


src/components/customerConnections/useSetConnectionAsDefault.ts
```diff
@@ -0,0 +1,88 @@
...
+      id
+      isDefault
+    }
+  }
+
+  mutation setCustomerIntegrationConnectionAsDefault(
+    $input: SetIntegrationCustomerAsDefaultInput!
+  ) {
+    setIntegrationCustomerAsDefault(input: $input) {
+      __typename
```
ansmonjol: This mutation selects only `__typename`, so its result never lands in the normalized cache - asymmetric with `setPaymentProviderCustomerAsDefault` right above, which selects `{ id, isDefault }`.

On the customer-details surface it's masked by the `refreshCustomer()` that `useCustomerConnectionsPersistence` runs after success. On the create/edit surface there is no refresh, and the `dontUpdateMeta: true` mirror in `ExternalAppsAccordion.handleSetDefault` keeps the field untouched - so a subsequent payment set-default (which does write the cache) makes `useGetSingleCustomerQuery` re-emit, `FormApi.update` re-applies `defaultValues` over the untouched field, and the mirrored integration default silently disappears from the UI until reload.

All six union members expose `id` and `isDefault`, so:

```graphql
setIntegrationCustomerAsDefault(input: $input) {
  id
  isDefault
}
```

fixes the cache and makes the local mirror unnecessary.

domenicofalco: Good catch on the cache — fixed in eb8798342. Two notes on the details:

`IntegrationCustomer` is a union, not an interface, so `{ id isDefault }` directly on it does not compile (`pnpm codegen`: `Cannot query field "id" on type "IntegrationCustomer"`). Went through one inline fragment per member instead, which gets the same normalization.

Kept the local mirror though: `SetAsDefaultService` flips the siblings server-side but the mutation only returns the connection it promoted, and `FormApi.update` re-applies default values only while the form is untouched — so on a form the user has already edited, the cache write never reaches the rendered values. The two are complementary: with the selection fixed, a default-values refresh now re-applies the same value the mirror wrote instead of a stale one, which is the bug you spotted.

ansmonjol (approved):

PR #4245, opened 2026-09-04
refactor(form): Normalize numeric text input values

AllanMichay:
## Context

`formatValue` in `src/components/form/TextInput/TextInput.tsx` was not
type-consistent. Every formatter branch returned a string except `int`, which
returned `parseInt(...)` — a number — and the result was laundered into the
public `onChange(value: string)` contract with an `as string` cast.

`AmountInput` made it worse: it auto-pushes the `int` formatter when
`getCurrencyPrecision(currency) === 0` (JPY, KRW, ISK, DJF, GNF, KMF), so the
same amount field emitted a string or a number depending on the currency the
user had selected.

The consequences were spread across the codebase: seven schema fields carried a
`z.union([z.string(), z.number()])` purely to model the ambiguity, and several
`z.string()` schemas silently received numbers — broken validation on 0-decimal
currencies and on `int`-formatted fields.

The exposed fields are the `AmountInput` ones with no `chargeDecimal` formatter,
since `AmountInput` early-returns on `chargeDecimal` before the 0-decimal `int`
push: `MinimumCommitmentDrawer` and the subscription-level progressive billing
form. Both were reproduced against `main` in the browser — a JPY amount is
rejected with "This value is not valid" and the submit button stays disabled.
The plan-level progressive billing drawer passes `chargeDecimal` and was never
affected, contrary to what the ticket originally claimed. Nothing crashed on
`.length` / `.trim()` either: the `z.string()` parse fails first, so those calls
are never reached.

A numeric text input now emits exactly one type: `string`, always.

## Description

Input layer — the only behavioural change:

- `formatValue` has an explicit `string | null` return type and no branch can
  return a number. The `int` branch stringifies and swallows a `NaN` parse
  explicitly, so the literal `'NaN'` can never leak into form state, and the
  no-formatter passthrough returns `String(value)`.
- The `as string` cast in `TextInput` and its apology comment are gone.
- The wider `onChange` annotations in `TextInputField`, `AmountInputField` and
  `AmountInputFieldForTanstack` are narrowed to `string`.
- Callers may still seed numbers from the API — only the emitted value
  normalizes.

Schema sweep — every input-backed union collapsed to `z.string()`, keeping each
`refine` / `superRefine` and its translation key: coupon, add-on, customer-portal
wallet, `AddCouponToCustomerDialog` and the quote discount drawer. The
`number | ''` unions are rewritten as `z.string()` plus a refine: both
grace-period dialogs, the net payment term custom period (whose refine checked
`typeof === 'number'`) and the subscription fee `trialPeriod`, which now keeps a
string form state behind an unchanged numeric public contract.

Consumer sweep — TypeScript cannot catch most of this, since the BigInt GraphQL
scalars generate as `any` and the plan/wallet schemas are `z.custom` with no
field-level checking, so every `int`-formatted field was traced by hand to its
submit boundary. `serializePlanInput` now coerces `packageSize` and `freeUnits`,
matching the `Number(...)` treatment its neighbours already had. Wallet
`priority` is typed as a string in the form types and coerced in both mappers.
`PackageCharge` compares its numeric values rather than the raw strings, and the
coupon code snippet emits `percentage_rate` and `frequency_duration` unquoted —
both had started printing a string for fields the API declares as numbers.

Tests pin the new contract: `formatValue` never emits a number for any
formatter, `NaN` never escapes, and the JPY / cleared-field / zero cases that
motivated the change are covered. New suites cover the billing entity grace
period dialog, the net payment term dialog and the coupon code snippet.

Two regression tests pin the composition that actually broke — a 0-decimal
currency making `AmountInput` push `int` — and both fail on `main`:
`AmountInputFieldForTanstack` asserts the field emits a string for JPY exactly
as for EUR, and `SubscriptionProgressiveBillingForm` drives a JPY threshold
through the real schema to the mutation.

Fixes LAGO-1759


src/components/customers/AddCouponToCustomerDialog.tsx
```diff
@@ -90,32 +90,19 @@ gql`
...
+const numericInputSchema = z.string().optional()
+
+const isAtLeast = (value: string | undefined, minimum: number): boolean =>
   Number(value ?? '') >= minimum
 
 const defaultFormValues = {
   couponId: '',
   couponType: CouponTypeEnum.FixedAmount as CouponTypeEnum,
-  amountCents: undefined as NumericInputValue,
+  amountCents: undefined as string | undefined,
```
domenicofalco: any chance to get rid of the assertion within this whole object?

domenicofalco (approved):

src/components/customers/AddCouponToCustomerDialog.tsx
AllanMichay: Good catch, done in 7e8137100.

Those nine assertions were only there to widen literal / `undefined` inference in an untyped object literal, so I declared an `AddCouponFormValues` type and annotated the constant instead — the shape is now stated once and checked, rather than asserted per field. `pnpm types` passes without them.

There are still six `as` in the rest of the file (the `validationSchema as unknown as never` for the TanStack validator, the field-meta error reads, and `frequency as CouponFrequency` at submit). Those are pre-existing and outside what you flagged, so I left them — happy to take them on here if you would rather they go too.

src/core/serializers/serializePlanInput.ts
```diff
@@ -128,7 +128,10 @@ export const serializeProperties = (properties: Properties, chargeModel: ChargeM
         }
       : { volumeRanges: undefined }),
     ...(chargeModel === ChargeModelEnum.Package
-      ? { freeUnits: properties?.freeUnits || 0 }
+      ? {
+          packageSize: Number(properties?.packageSize) || undefined,
```
ansmonjol, line 132: Non-blocking. `Number(...) || undefined` maps a `packageSize` of `0` to `undefined`, and loses precision above `2^53` on the `BigInt` scalar.

Both are unreachable today: `chargePropertiesSchema.validatePackage` rejects `packageSize < 1` before submit, including for rate-card rates via `buildRateCardRateSchema`. Worth keeping in mind if that validation ever moves or is relaxed.

src/components/form/TextInput/TextInput.tsx
```diff
@@ -56,80 +56,79 @@ const numberFormatter = new RegExp(
...
+]
+
+const hasNumberFormatter = (formatterFunctions: FormatterFunctions): boolean =>
+  numberFormatter.test(
+    typeof formatterFunctions === 'string' ? formatterFunctions : formatterFunctions.join(''),
+  )
+
+const applyNumericFormatters = (value: string, formatterFunctions: FormatterFunctions): string => {
+  // A lone '-' is a valid intermediate state: the user is typing a negative number.
+  if (value === '-') return value
```
ansmonjol, line 83: Non-blocking. This early return now also covers `sextDecimal`, which the old code excluded: it was the one decimal branch written without the `formattedValue !== '-'` guard. So `formatValue('-', 'sextDecimal')` returns `'-'` where it used to return `''`.

Unreachable today: `sextDecimal` is only ever paired with `positiveNumber` (`FixedChargeDrawerContent`), and `positiveNumber` strips the `-` before this runs.

ansmonjol (commented):
Two non-blocking observations from a full pass over the diff. Both look unreachable today, flagging them only so the assumption is written down somewhere.

src/core/serializers/serializePlanInput.ts
AllanMichay: Thanks for writing both of these down — checked them, and I agree they hold but stay theoretical for now.

On `0`: `validatePackage` rejects `Number(packageSize) < 1` before submit, so the serializer never sees it. Emitting `0` would actually be worse than omitting the field, since the API rejects it too.

On the `2^53` precision: the rounding already happens one layer up — `PackageCharge` sets `beforeChangeFormatter={['positiveNumber', 'int']}` and `int` runs `parseInt` on every keystroke (same on `main`), so form state holds the rounded value and `Number()` here adds nothing. Fixing it for real means dropping `int` and carrying the digits as a string to the BigInt scalar — worth doing if that formatter ever goes away, but bigger than this PR.

src/components/form/TextInput/TextInput.tsx
AllanMichay: Good eye — that one is deliberate. On `main`, `sextDecimal` was the only decimal branch written without the `!== "-"` guard that `decimal` / `triDecimal` / `quadDecimal` / `chargeDecimal` all had, so table-driving the five into `DECIMAL_TRUNCATIONS` normalised it. I read the old asymmetry as the accident rather than the intent.

Agreed it is unreachable either way: the sole `sextDecimal` call site is `FixedChargeDrawerContent` with `["positiveNumber", "sextDecimal"]`, and `positiveNumber` strips the `-` first. The new behaviour is pinned in `FormatValue.test.ts` so it will not drift back silently.

PR #4258, opened 2026-09-07
fix(devtools): stop the panel crashing on a shared inspector link

domenicofalco:
## Context

Opening a copied inspector link (`?devtool-tab=…`) while logged out, then logging in, showed a generic error toast instead of the developer tools panel. The panel stayed closed until the user refreshed the page.

Two defects combined:

1. `usePanel.openPanel` resized the panel imperatively and synchronously. `react-resizable-panels` asserts (and throws) `Panel size not found for panel "devtools-panel"` while the panel is not yet in the group's committed layout — which is exactly the state when `openPanel` is flushed from a passive effect during the commit that mounts the panel. Thrown from an effect, it reached `DevtoolsErrorBoundary`.
2. `checkParamsFromUrl` lived in the `useDeveloperTool` consumer hook, so it ran in all 14 consumers. Each read `window.location.search` and called `openPanel()` on its own, so which consumer stripped the param was a race — and the consumers inside the devtools MemoryRouter could never strip it from the browser URL, leaving `openPanel()` free to fire again right after the panel mounted. `DevtoolsView` being a consumer also pointed the devtools MemoryRouter at an app route it does not define, leaving the panel blank.

## Description

`usePanel` now routes every imperative resize through `safeResize`: the call is deferred to the next frame, so the group recomputes its layout first, and a `try/catch` reports a library assert to Sentry instead of letting it reach an error boundary. `expandPanel` already deferred its resize and now shares the helper. The resize is kept rather than dropped because the AI agent panel is mounted permanently at size 0 and relies on it to open.

The `devtool-tab` bridge moves out of the consumer hook into a dedicated `useDevtoolTabParam`, mounted once by `RouteWrapper`. It reads the router's `search` instead of `window.location.search` and guards on the tab it already consumed, so the param is handled exactly once. `useDeveloperTool` is now a pure context accessor, which also removes the MemoryRouter mis-navigation.

`useDevtoolTabParam` lives in `RouteWrapper` rather than in `DeveloperToolProvider` as the ticket proposed: the provider wraps the `PanelGroup` that holds both routers, so it has no Router context and could only strip the param via `window.history.replaceState` — which leaves React Router's location, and therefore `useLocationHistory`'s `goBack`, still carrying it. `RouteWrapper` is the single stable host inside the BrowserRouter.

### Tests

Unit: a new `usePanel` suite covering the deferred resize, the swallowed assert and the fullscreen paths; the `useDeveloperTool` suite reworked around the new bridge, including a regression check that the consumer hook alone no longer touches the panel. Coverage on the two changed hooks is 98.8% statements.

E2E: `cypress/e2e/t50-devtools-inspector-link.cy.ts` covers the panel opening on the linked tab from an inspector link, and the ticket's own reproduction — the link opened while logged out and restored through `location.state.from` after logging in. Devtools test ids moved to a standalone `utils/dataTestConstants.ts` so the spec can import them (Cypress cannot import the components), and the devtools tabs gained `dataTest` values since their titles are translation keys.

Fixes LAGO-1913

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -81,6 +81,7 @@ jobs:
             cypress/e2e/t10-*
             cypress/e2e/t20-*
             cypress/e2e/t30-*
+            cypress/e2e/t50-*
 
       - name: Upload screenshots
         uses: actions/upload-artifact@v7.0.1
--- a/cypress/e2e/t50-devtools-inspector-link.cy.ts
+++ b/cypress/e2e/t50-devtools-inspector-link.cy.ts
@@ -1,0 +1,50 @@
+import {
+  DEVTOOLS_COPY_INSPECTOR_LINK_TEST_ID,
+  DEVTOOLS_TAB_WEBHOOKS_TEST_ID,
+} from '~/components/developers/utils/dataTestConstants'
+
+import { userEmail, userPassword } from '../support/reusableConstants'
+
+// The devtools address a copied inspector link carries, as `?devtool-tab=` holds it.
+const devtoolsAddress = '/devtool/webhooks'
+const inspectorLinkPath = `/analytics?devtool-tab=${encodeURIComponent(devtoolsAddress)}`
+
+const expectPanelOpenOnWebhooks = (): void => {
+  cy.get(`[data-test="${DEVTOOLS_COPY_INSPECTOR_LINK_TEST_ID}"]`).should('exist')
+  cy.get(`[data-test="${DEVTOOLS_TAB_WEBHOOKS_TEST_ID}"]`).should(
+    'have.attr',
+    'aria-selected',
+    'true',
+  )
+  // The panel used to stay closed behind a generic error toast from `DevtoolsErrorBoundary`.
+  cy.get('[data-test="toast/danger"]').should('not.exist')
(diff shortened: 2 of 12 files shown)


domenicofalco:
`Tests (shard 2/4)` failed on `6c0b8d2` — 1 failed of 3231, in `src/pages/wallet/components/__tests__/RecurringRuleDrawer.test.tsx:487`, a `waitFor` on the `paidCredits` input timing out.

Not this PR's failure:

- That file is not in this diff. The change touches `src/hooks/ui/usePanel.tsx`, `src/hooks/useDeveloperTool.tsx`, `src/components/RouteWrapper.tsx` and their tests — nothing under `src/pages/wallet/`.
- All four shards were green on the previous head `48c11e1`, which already carried **every** source change in this PR. The only delta in `6c0b8d2` is a deleted code comment, which cannot alter test behaviour.
- The suite passes locally on this branch: 29/29, including that assertion.

So it is a timing-sensitive `waitFor` flaking under CI load. No fix exists to port, and hardening an unrelated wallet drawer test would widen this PR beyond its scope, so I have re-run the failed job once rather than changing it. If it fails again on the re-run the failure is real and I will root-cause it.

---

ansmonjol (approved):

PR #4269, opened 2026-09-07
fix(customers): count completed integration polls

ansmonjol:
Customer integration links are created asynchronously after editing a customer, so the first details query can arrive before those links exist. The edit flow opens the customer page with `shouldPollIntegrations: true`, asking it to refresh briefly while the backend finishes.

## Why the previous counter was wrong

Apollo did the polling through `startPolling()` while a separate effect counted attempts:

```tsx
useEffect(() => {
  if (!shouldPollIntegrations) return

  pollingAttemptsRef.current += 1

  // Stop when an integration exists or the counter reaches 3.
}, [shouldPollIntegrations, hasAnyIntegrationCustomer, stopPolling, navigate, location.pathname])
```

That counter measured effect executions, not completed requests. The project's `useNavigate()` wrapper returns a new function on each render, and `navigate` was a dependency, so unrelated renders could rerun the effect and consume the allowance:

| Event | Old counter | Completed polling requests |
| --- | ---: | ---: |
| Initial effect runs | 1 | 0 |
| Another render changes `navigate` | 2 | 0 |
| Another render changes `navigate` | 3, polling stops | 0 |

The initial effect already consumed an attempt before any polling response. Stabilizing `navigate` alone would not have fixed the counting: repeated responses with no integration leave the effect's dependencies unchanged, so the counter would stop advancing instead.

## What replaces it

The whole lifecycle moves out of `CustomerDetails` into `useCustomerIntegrationPolling` (`src/hooks/customer/`), a sequential refetch loop that ties counting to `refetch()` completion:

1. Start only when a customer ID and the navigation flag are present.
2. While the initial query is loading, check again after one second without consuming an attempt, bounded by `MAX_INTEGRATION_POLLING_LOADING_WAITS` so a request that never settles cannot keep the cycle alive.
3. If the current customer already has an integration, finish immediately.
4. Await one refetch. Only once it resolves or rejects does `completedPolls` increase.
5. Stop when an integration appears or three attempts have completed.
6. Otherwise schedule the next request one second after the previous one settled.

```tsx
try {
  const result = await refetch()
  integrationFound = hasIntegrationCustomer(result.data?.customer)
} catch {
  // `errorPolicy: 'all'` (apolloClient/init.ts) resolves GraphQL errors with no data,
  // so only network failures reject here and the query's error state surfaces them.
}

if (cancelled) return

completedPolls += 1

if (integrationFound || completedPolls >= MAX_INTEGRATION_POLLING_ATTEMPTS) {
  finishPolling()
  return
}

timeout = setTimeout(() => void poll(), INTEGRATION_POLLING_INTERVAL)
```

Concretely: identical responses still count as completed attempts; loading transitions and unrelated renders do not consume attempts; failed requests count toward the limit, preventing endless retries; a cycle waits for its current request before scheduling another; and finishing clears the navigation flag with `replace`, adding no history entry.

The limit remains three completed attempts, not a three-second deadline. Request duration and initial loading add to the elapsed time.

## `pollingContextRef`

```tsx
const pollingContextRef = useRef({ customer, loading, navigate })

useEffect(() => {
  pollingContextRef.current = { customer, loading, navigate }
}, [customer, loading, navigate])
```

This lets the active loop read the latest query state and navigation function without restarting whenever those values change. The loop restarts only on its own dependencies: customer, route pathname, navigation key, polling flag, or `refetch`.

## Counter lifetime and cleanup

`cancelled` and `completedPolls` are declared inside the effect, so they belong to one effect execution, and that execution's scheduled callbacks close over them.

| Situation | Counter |
| --- | --- |
| Ordinary render, dependencies unchanged | Active loop keeps its counter |
| A dependency changes | Old effect cleaned up, new execution starts at `0` |
| Component unmounts | Old cycle cancelled |
| Component mounts again | Fresh execution starts at `0` |

React runs the old cleanup before setting up an effect with changed dependencies, and on unmount ([effect lifecycle](https://react.dev/reference/react/useEffect#parameters)). So cleanup only has to neutralize what could still run:

```tsx
return () => {
  cancelled = true
  clearTimeout(timeout)
}
```

`clearTimeout()` stops a scheduled request from starting. A request already in flight can still settle and briefly retain the old closure; the `if (cancelled) return` after it stops that continuation from incrementing the counter, navigating, or scheduling another request. Resetting `completedPolls` in cleanup would only reset the old, cancelled counter: it would not affect the next cycle and would not guard against late responses.

One limitation: cleanup does not abort the network request nor stop Apollo from processing its response. It only prevents this polling loop from acting on the completion.

## Tests

Cover unrelated renders, initial loading, a query that never settles, a refetch resolving without data, identical responses, pending requests, rejected requests, Strict Mode replay, unmount before a timer fires, unmount during a request, and navigation to another customer with a fresh three-attempt allowance.

### DIFF
--- a/src/pages/CustomerDetails.tsx
+++ b/src/pages/CustomerDetails.tsx
@@ -1,5 +1,5 @@
 import { gql } from '@apollo/client'
-import { useEffect, useRef } from 'react'
+import { ReactElement, useEffect, useRef } from 'react'
 import { useParams } from 'react-router-dom'
 
 import { useAddCouponToCustomerDialog } from '~/components/customers/AddCouponToCustomerDialog'
@@ -15,6 +15,7 @@ import { CUSTOMERS_LIST_ROUTE, useLocation, useNavigate } from '~/core/router'
 import {
   AddCustomerDrawerFragmentDoc,
   CustomerMainInfosFragmentDoc,
+  GetCustomerQuery,
   LagoApiError,
   useGetCustomerQuery,
 } from '~/generated/graphql'
@@ -66,9 +67,11 @@ gql`
   ${CustomerMainInfosFragmentDoc}
 `
 
-const CustomerDetails = () => {
+const hasIntegrationCustomer = (customer: GetCustomerQuery['customer'] | undefined): boolean =>
+  !!customer?.integrationCustomers?.length
+
+const CustomerDetails = (): ReactElement => {
   const { openAddCouponToCustomerDialog } = useAddCouponToCustomerDialog()
-  const pollingAttemptsRef = useRef(0)
   const { translate } = useInternationalization()
   const navigate = useNavigate()
   const location = useLocation()
@@ -77,7 +80,7 @@ const CustomerDetails = () => {
   const shouldPollIntegrations = (location.state as { shouldPollIntegrations?: boolean })
     ?.shouldPollIntegrations
 
-  const { data, loading, error, startPolling, stopPolling } = useGetCustomerQuery({
+  const { data, loading, error, refetch } = useGetCustomerQuery({
     variables: { id: customerId as string },
     skip: !customerId,
     notifyOnNetworkStatusChange: true,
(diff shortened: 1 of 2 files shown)


src/pages/CustomerDetails.tsx
```diff
@@ -86,36 +90,87 @@ const CustomerDetails = () => {
...
-    if (shouldPollIntegrations && !hasAnyIntegrationCustomer) {
-      pollingAttemptsRef.current = 0
-      startPolling(INTEGRATION_POLLING_INTERVAL)
-    }
+    pollingContextRef.current = { customer, loading, navigate }
+  }, [customer, loading, navigate])
 
-    return () => {
-      stopPolling()
+  useEffect(() => {
```
domenicofalco: Non-blocking suggestion: could we extract this lifecycle into a dedicated `useCustomerIntegrationPolling` hook under `src/hooks/customer/`? It now owns timers, completed-attempt counting, initial-loading waits, error handling, and cancellation. Encapsulating that behavior would keep `CustomerDetails` focused on page composition and let us test the polling lifecycle independently with `renderHook`.

A customer-specific hook seems sufficient here: the integration checks and navigation-state cleanup belong to this flow. We can consider a shared polling hook later if another use case needs the same lifecycle.

domenicofalco (approved):

src/pages/CustomerDetails.tsx
ansmonjol: indeed it makes sense! on it

ansmonjol: Extracted in 5ab644a47: `src/hooks/customer/useCustomerIntegrationPolling.ts` now owns the timers, completed-attempt counting, initial-loading waits, error handling and cancellation, and the 21 polling tests moved to a `renderHook` suite under `src/hooks/customer/__tests__/`.

domenicofalco (approved):

PR #4275, opened 2026-09-07
feat(payments): add filters to the payments list

sarkissianraffi:
The payments list now has eleven URL-persisted filters alongside search: status, amount, receipt number, created date, provider, method, currency, invoice number, customer, payment type and payable type. Filter changes reset pagination and issue one getPaymentsList request.

The `pa` adapter maps each chip to the API argument. Payment statuses include processing, enum options come from generated GraphQL types, and exact amount strings are preserved through URL parsing and BigInt variables, including 9223372036854775807 cents. Existing invoice/credit-note amount behavior is unchanged via a payment-only precision opt-in. The existing invoice-number selector remains unchanged on other pages; payments use the exact-text variant.

## Verification

- Full regression run: `TZ=UTC pnpm exec jest --config jest.config.ts --coverage --maxWorkers=4` — **1,023 suites / 12,772 tests pass**.
- 710 tests across the filter system and Payments/Invoices/CreditNotes pages pass; final focused run 241 tests passes.
- Changed executable lines: 124/127 covered (97.64%), branches 89/94 (94.68%), combined coverage **96.38%**; both operation amount variables are asserted to be BigInt.
- Types pass. Lint: no errors, 57 existing warnings; changed files clean. Translations added through the repository script; inspect and consistency checks pass.
- Generated GraphQL file rebuilt against the local API.
- 47 live browser checks: every filter, URL/reload persistence, four amount modes including reversed between bounds, full int64 values, inclusive dates, three filters plus search, clear all, page-two reset, count/row parity with REST, and one request per change. No export button or DataExport code.
- CLI and all six SDKs match the same seven succeeded EUR payment IDs.

[API and client QA](https://github.com/getlago/lago-api/blob/feat/payments-list-filters/qa/payments-filters/README.md), [79 HTTP assertions](https://github.com/getlago/lago-api/blob/feat/payments-list-filters/qa/payments-filters/api-qa.json), [cross-client IDs](https://github.com/getlago/lago-api/blob/feat/payments-list-filters/qa/payments-filters/cross-client-qa.json), [UI QA and recording](https://github.com/getlago/lago-front/blob/feat/payments-list-filters/qa/payments-filters/README.md)

Reproduce on a running development API (synthetic fixtures, no gateway calls):

```sh
lago exec api bundle exec rails runner script/seed_payments_filters.rb
python3 script/qa_payments_filters.py
lago exec api bundle exec rails runner script/benchmark_payments_filters.rb
```

[Seed script](https://github.com/getlago/lago-api/blob/feat/payments-list-filters/script/seed_payments_filters.rb): 30 payments / 29 visible, three customers, stub Stripe and GoCardless, all statuses/types/methods, receipts, two-invoice PaymentRequest, two currencies, dates at organization-timezone boundaries, zero and amounts through 9223372036854775807. Credentials stay in ignored mode-0600 files.

## Screenshots and recording

Before:
Before

After:
After

[Filter panel](https://github.com/getlago/lago-front/blob/feat/payments-list-filters/qa/payments-filters/payments-panel.png) · [Screen recording](https://github.com/getlago/lago-front/blob/feat/payments-list-filters/qa/payments-filters/payments-filters.webm)

## Compatibility notes

Requires the linked API schema. [Feature-branch codegen and TypeScript CI passed](https://github.com/getlago/lago-front/actions/runs/34168929265); GitHub lint and CodeQL pass. Standard Tests/Cypress/Sonar jobs skip draft PRs, so the full test and coverage results above were also run locally. The default codegen workflow targets API main; merge the API dependency before promoting this PR. An unchanged DatePicker test constructs a historical date before its timezone setup, so the full local suite uses TZ=UTC to match CI; the component and its test are unchanged. The provider method enum is named `PaymentProviderMethodTypeEnum` to avoid the existing manual/provider enum. The shared payment table's existing display formatter rounds amounts above JS safe integer precision; filter inputs, chips, URLs and query bounds stay exact. That shared display behavior is outside this change. No payment response fields or export added.

## Related PRs

- [lago-api](https://github.com/getlago/lago-api/pull/6325)
- [lago-front](https://github.com/getlago/lago-front/pull/4275)
- [lago-openapi](https://github.com/getlago/lago-openapi/pull/579)
- [lago-go-client](https://github.com/getlago/lago-go-client/pull/365)
- [lago-python-client](https://github.com/getlago/lago-python-client/pull/419)
- [lago-ruby-client](https://github.com/getlago/lago-ruby-client/pull/373)
- [lago-rust-client](https://github.com/getlago/lago-rust-client/pull/61)
- [lago-javascript-client](https://github.com/getlago/lago-javascript-client/pull/99)
- [lago-php-client](https://github.com/getlago/lago-php-client/pull/5)
- [lago-cli](https://github.com/getlago/lago-cli/pull/27)

Merge order: API, then front; OpenAPI before client/CLI releases. JavaScript CI uses the pinned feature spec during that rollout.

## Performance follow-up (2026-09-09)

- `paymentMethodType` filter removed, together with its panel item, value labels, translations and tests: the API dropped `payment_method_type` for performance reasons (getlago/lago-api#6325, "Not shipped for performance reasons"). Generated GraphQL types regenerated from the API schema.
- The payments collection metadata now carries `totalCountCapped` and `hasNextPage` like invoices; the header shows "10,000+" above the API's counting limit and pagination advances on `hasNextPage`.
- `pnpm code:style` green (0 errors), 239 filter tests and the payments page tests green.
- Depends on getlago/lago-api#6325; indexes ship first in getlago/lago-api#6341.


src/pages/PaymentsPage.tsx
```diff
@@ -59,24 +99,51 @@ const PaymentsPage = () => {
...
+      page,
+      searchTerm,
+    }),
+    [searchParams, organization?.defaultCurrency, pageSize, page, searchTerm],
+  )
+
+  useEffect(() => {
+    if (!searchPending) return
+
+    const timeout = setTimeout(() => {
```
domenicofalco, line 118: **Blocking** — This debounce restarts on every re-render, not just on keystrokes.

`goToPage` is in the dependency array and is a new function on every render: `usePageSearchParam` lists `navigate` in its `useCallback` deps (`src/components/designSystem/Pagination/usePageSearchParam.ts:33`), and `~/core/router/useNavigate` returns a fresh closure each render with no memoization (`src/core/router/useNavigate.ts:38`). Any re-render inside the 500 ms window clears the timeout and starts a new one, and with `notifyOnNetworkStatusChange: true` + `fetchPolicy: 'network-only'` those renders arrive routinely, so the typed term can stay uncommitted.

We already have two shared solutions, both immune to this because the debounce is created once with empty deps:

- `useDebouncedSearch` — for search-only lists (`CustomersList`, `PlansList`, `CouponsList`).
- `CustomerInvoicesTab.tsx:108-117` — the closer precedent, since it has URL filters *and* search: `useMemo(() => debounce((value) => setSearchTerm(value || undefined), DEBOUNCE_SEARCH_MS), [])` plus a `cancel()` on unmount.

Please use one of them rather than a bare `setTimeout` keyed on render-unstable deps.

src/pages/PaymentsPage.tsx
```diff
@@ -59,24 +99,51 @@ const PaymentsPage = () => {
...
+
+    return () => clearTimeout(timeout)
+  }, [nextSearchTerm, searchPending, goToPage])
+
+  useEffect(() => {
+    const signature = JSON.stringify(queryVariables)
+
+    if (queryPending || lastQuery.current === signature) return
+
+    lastQuery.current = signature
```
domenicofalco, line 131: **Blocking** — The established shape is to put the filters in the query's `variables` and let Apollo re-execute — `InvoicesPage.tsx:159-176` with `useGetInvoicesListLazyQuery`, and `CustomerInvoicesTab.tsx:86-106` with a plain `useQuery` (`fetchPolicy: 'network-only'`, `searchTerm` and the formatted filters passed straight in). That also removes the need for `queryPending`.

Beyond the divergence, this manual dedup has three concrete defects:

- The signature is stored **before** the request settles, so a failed `getPayments` is never retried for that signature — the effect skips it forever.
- The signature depends on the key insertion order of `formatFiltersForPaymentsQuery`'s reduce, so reordering URL params yields a different signature for identical variables.
- It loses the loading anti-blink behaviour `useDebouncedSearch` provides.

Related, on line 87: `searchInput.trim().length >= 3` duplicates `MIN_SEARCH_CHARS` (`src/hooks/useDebouncedSearch.ts:8`), which isn't exported precisely so callers don't re-derive it.

src/pages/PaymentsPage.tsx
```diff
@@ -97,13 +164,18 @@ const PaymentsPage = () => {
...
         }}
         filtersSection={
-          <SearchInput
-            onChange={(value) => {
-              goToPage(1)
-              paymentsDebounceSearch?.(value)
-            }}
-            placeholder={translate('text_17370296250897aidak5kjcg')}
-          />
+          <Filters.Provider
```
domenicofalco, line 167: **Blocking** — The ten filters added here need the list's placeholders updated too. `PaymentsList.tsx:305,317` branch only on `variables?.searchTerm`, so any `pa_*` combination with no matches renders "No payments have been recorded for invoices. Please record a payment or connect your customer to a payment provider…" — wrong when payments exist but are filtered out.

Two existing ways to fix it: `InvoicesList.tsx:105` feeds `searchParams` into `getEmptyStateConfig({ hasSearchTerm, searchParams, translate })`, and `CustomerInvoicesTab.tsx:119` derives an `isFiltering` flag from search + filters.

src/pages/PaymentsPage.tsx
```diff
@@ -59,24 +99,51 @@ const PaymentsPage = () => {
...
+    if (queryPending || lastQuery.current === signature) return
+
+    lastQuery.current = signature
+    void getPayments({ variables: queryVariables })
+  }, [getPayments, queryVariables, queryPending])
+
+  const paymentsIsLoading = loading || queryPending
 
   const paymentsTotalCount = data?.payments?.metadata?.totalCount
+  // Above the API's counting limit the total is a lower bound: print "10,000+", as invoices do.
```
domenicofalco, line 138: **Nit** — This repeats the JSDoc of `formatCountToMetadata` (`src/components/MainHeader/formatCountToMetadata.ts:4-12`) and justifies the change rather than explaining the code — that belongs in the commit body, per `.agents/docs/typescript-conventions.md` → "Comments: Default to None".

Same for `// Keep payment amounts exact through URL parsing, interval ordering and cents conversion.` in `paymentFilterValues.ts:13`, which restates what the function names already say. (The jsdom / react-virtual comment in the new test is fine — that's a constraint the reader can't see.)

src/pages/PaymentsPage.tsx
```diff
@@ -30,11 +48,23 @@ gql`
...
+      invoiceNumber: $invoiceNumber
+      paymentType: $paymentType
+      payableType: $payableType
     ) {
       metadata {
         currentPage
         totalPages
         totalCount
+        totalCountCapped
+        hasNextPage
```
domenicofalco, line 67: **Info** — Side note: these two fields, and the new filter arguments above, don't exist on API main — they come from getlago/lago-api#6325. Front codegen CI can't pass until that merges, so the merge order needs to be explicit on the PR.

src/components/Filters/graphql/FiltersPanelItemTypeSwitch.tsx
```diff
@@ -94,6 +106,7 @@ export const FiltersPanelItemTypeSwitch = ({
   ...props
 }: FiltersPanelItemTypeSwitchProps) => {
   const { translate } = useInternationalization()
+  const paymentFilters = useContext(FilterContext)?.filtersNamePrefix === PAYMENT_LIST_FILTER_PREFIX
```
domenicofalco, line 109: **Blocking** — Branching the shared registry on the URL prefix has three problems:

- The registry dispatches on `AvailableFiltersEnum`; this makes one member mean two different things depending on which page renders it. A second consumer of these filters (a customer payments tab, a payments dialog) silently gets the invoice behaviour.
- The same prefix test then repeats in `graphql/utils.ts` (`formatActiveFilterValueDisplay`, twice), so the branching spreads across the engine.
- It bypasses the sanctioned `useFilterContext()` accessor (`presentation/context.tsx:61`), which throws when the provider is missing — so the `?.` here is dead defensiveness.

`FiltersProvider` has no per-page override mechanism by design (`context.tsx:11-18`: only `availableFilters`, `staticFilters`, `displayInDialog`), and the codebase already solves this exact collision with distinct enum members — `productProductCategory` / `productFilterProductCategory` / `rateCardProductCategory` (`presentation/types.ts:104-108`), same translation key, different behaviour. This PR adds five dedicated members already, so please do the same for the three colliding ones (`payablePaymentStatus`, `paymentInvoiceNumber`, and either `paymentAmount` or an enum-keyed precision flag) and drop every `=== PAYMENT_LIST_FILTER_PREFIX` check from shared code.

src/components/Filters/graphql/utils.ts
```diff
@@ -221,6 +234,17 @@ export const isValidDateRangeValue = (value?: string): boolean => {
   return fromDate <= toDate
 }
 
+export const isValidPaymentDateRangeValue = (value?: string): boolean => {
```
domenicofalco, line 237: **Suggestion** — The only difference from `isValidDateRangeValue` right above is that one bound may be empty. The fork forces `FiltersPanelPopper` to add both an explicit `filterType !== AvailableFiltersEnum.paymentCreatedAt` exclusion and a second `.when` block, so the next open-ended date filter has to remember both edits. Prefer `isValidDateRangeValue(value, { allowOpenEnded })`, driven by a set alongside `FiltersItemDates`.

Same shape with `orderPaymentAmountBounds` vs `orderIntervalBounds` (`utils.ts:131`) and `parsePaymentAmountValue` vs `parseFromToValue` (`utils.ts:137`): the existing ones are `Number`-based, so the fork has a real reason — but an `{ exact: true }` option on the shared helpers keeps one implementation instead of two that must stay in sync (`formatActiveFilterValueDisplay` now ternaries between them).

src/components/Filters/graphql/paymentFilterValues.ts
```diff
@@ -0,0 +1,86 @@
...
+    case AmountFilterInterval.isAtLeast:
+      return { amountFrom, amountTo: null }
+    case AmountFilterInterval.isUpTo:
+      return { amountFrom: null, amountTo }
+    default:
+      return { amountFrom: null, amountTo: null }
+  }
+}
+
+export const paymentAmountToCents = (
```
domenicofalco, line 52: **Blocking** — The `Number`-based `serializeAmount` genuinely can't carry int64 amounts, so a Decimal converter is the right call — but this is the third private copy of the same concern: `formatAmountCurrency` already exists in `InvoicesPage.tsx:116` and `CreditNotesPage.tsx:89`, both carrying a `// TODO: This is a temporary workaround` marker this copy drops. Please put the exact converter next to `serializeAmount` in `src/core/serializers/serializeAmount.ts` and converge the three call sites.

The functional bug stands either way: the currency comes from `filters.currency || organization.defaultCurrency` while the amount filter is selectable without a currency filter, so on a multi-currency org "amount >= 10" with a USD default sends `amountFrom: 1000` — 100x off for JPY rows (precision 0) and 10x off for BHD. Either require a currency alongside the amount filter in the panel, or scope the conversion per row currency.

src/components/Filters/graphql/filtersElements/FiltersItemPayablePaymentStatus.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const FiltersItemPayablePaymentStatus = ({ value, setFilterValue }: Props) => {
+  const { translate } = useInternationalization()
+
+  const options = Object.values(PayablePaymentStatusEnum).map((option) => ({
+    value: option,
+    label: translate(paymentStatusLabels[option]),
+  }))
+
+  return (
+    <MultipleComboBox
```
domenicofalco, line 21: **Bug** — Missing `displayInDialog`. Every existing enum item reads `const { displayInDialog } = useFilters()` and forwards `PopperProps={{ displayInDialog }}` to `MultipleComboBox` (`FiltersItemHttpMethods.tsx:14,19`, `FiltersItemCurrencies`, `FiltersItemActivityTypes`, …), so the option popper is positioned correctly when the filters panel renders inside a dialog.

Same for `FiltersItemPayableType`, `FiltersItemPaymentProviderType` and `FiltersItemPaymentType` — one line each.

src/components/Filters/presentation/types.ts
```diff
@@ -425,6 +443,11 @@ const translationMap: Record<AvailableFiltersEnum, string> = {
   [AvailableFiltersEnum.paymentDisputeLost]: 'text_66141e30699a0631f0b2ed32',
   [AvailableFiltersEnum.paymentOverdue]: 'text_666c5b12fea4aa1e1b26bf55',
   [AvailableFiltersEnum.paymentStatus]: 'text_63eba8c65a6c8043feee2a0f',
+  [AvailableFiltersEnum.receiptNumber]: 'text_17416040051091zpga3ugijs',
+  [AvailableFiltersEnum.paymentProviderType]: 'text_634ea0ecc6147de10ddb6631',
+  [AvailableFiltersEnum.paymentType]: 'text_1788818972604tg2ogag4h2r',
+  [AvailableFiltersEnum.payableType]: 'text_17888189726040erbp5xejy9',
+  [AvailableFiltersEnum.paymentCreatedAt]: 'text_664cb90097bfa800e6efa3f5',
```
domenicofalco, line 450: **Nit** — That key is "Date", while the sibling created-at filters use `text_1776870266380s3zbpmnfrhj` ("Created at") — see `quoteCreatedAt` / `orderFormCreatedAt` a few lines below. The chip should read "Created at".

src/pages/__tests__/PaymentsPage.test.tsx
```diff
@@ -77,6 +80,74 @@ describe('PaymentsPage', () => {
...
+    expect(testMockNavigateFn).toHaveBeenCalledWith(
+      { search: 'pa_currency=EUR' },
+      { replace: true },
+    )
+  })
+
+  it('does not expose an export action', () => {
+    render(<PaymentsPage />)
+    expect(capturedConfig?.actions?.items).toHaveLength(1)
+    expect(screen.queryByRole('button', { name: /export/i })).not.toBeInTheDocument()
```
domenicofalco, line 150: **Bug** — This assertion can never fail: `MainHeader.Configure` is mocked at `:14-17` to render only `props.filtersSection`, so no action button ever reaches the DOM — it would still pass if an export action were added. It also pins the absence of a feature that was never in scope.

Keep the `toHaveLength(1)` line and drop this one; or, if the header really should be asserted, render the real `MainHeader` the way `TeamAndSecurity.test.tsx:9-14` does.

src/components/Filters/graphql/filtersElements/__tests__/PaymentFilterItems.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+        setFilterValue={setFilterValue}
+      />,
+    )
+
+    expect(setFilterValue).toHaveBeenLastCalledWith('isEqualTo,90071992547409.92,90071992547409.92')
+    fireEvent.change(screen.getByRole('textbox'), { target: { value: '90071992547409.93' } })
+    expect(setFilterValue).toHaveBeenLastCalledWith('isEqualTo,90071992547409.93,90071992547409.93')
+  })
+
+  it('uses exact invoice text only in the payments panel', () => {
```
domenicofalco, line 157: **Suggestion** — The `only` half of the name is never exercised: the `in`-prefix case, which must still get the `FiltersItemInvoiceNumber` combobox, isn't rendered — and that's precisely the regression the prefix branching risks. Please add it (best case this test disappears along with the prefix branching).

src/generated/graphql.tsx
```diff
@@ -902,6 +905,20 @@ export type CashfreeProvider = {
   successRedirectUrl?: Maybe<Scalars['String']['output']>;
 };
 
+/** A product-catalog plan */
+export type CatalogPlan = {
```
domenicofalco, line 909: **Blocking** — This regeneration carries schema drift unrelated to payments: `CatalogPlan`, `createCatalogPlan`/`updateCatalogPlan` changing from `Plan` to `CatalogPlan`, new `AccountTree` and `CatalogPlan` enum members, a `taxCodes` input field, and `... on CatalogPlan { id }` added to the activity-log documents. Please regenerate against the schema this PR targets so the generated diff holds only the payments filter arguments and `PaymentCollectionMetadata` — otherwise it conflicts with whichever PR owns the catalog changes.

qa/payments-filters/README.md
```diff
@@ -0,0 +1,67 @@
+# Payments filters QA
```
domenicofalco, line 1: **Blocking** — This directory adds ~1.37 MB to git history permanently: `payments-filters.webm` (794 KB), three PNGs (~527 KB), a 1285-line `ui-qa.json`, plus `cli-comparison-ui.json` and `cross-client-qa.json`. The evidence is useful — please move it to the PR description or an attachment and strip the directory from the branch (`.gitignore` has no rule for `qa/` or `.webm` today, so it's worth adding one). Or just delete such folder unless you have some specific action in mind.

ansmonjol: I would recommend to melt those test in our existing test structure, following our implementation examples, and completely delete this new `qa/*` folder.
If you can prove that those assertions cannot be replicated using our jest or cypress test suite for some reasons, we could consider evolving our test structure

PR #4277, opened 2026-09-08
fix(a11y): render navigation tabs, table rows and row menus as real links

domenicofalco:
## Context

Alexis Tacnet (Mistral AI) reported that most navigation targets in the app are JS click handlers instead of anchors, so they cannot be opened in a new tab, middle-clicked, right-click "Open link in new tab", copied as a URL or previewed in the browser status bar, and screen readers do not announce them as links.

Three surfaces were affected: the navigation tabs, the table rows (~50 lists), and the row action menus on the right.

## Description

**Navigation tabs** (`NavigationTab`, `NavigationTabBar`) render their MUI `Tab` with `component={Link} to={tab.link}` instead of a `button` + imperative `navigate`, so the URL is a real `href` and the browser owns modifier clicks. Clicking the tab that already matches the URL still no-ops, unless a modifier is held. A disabled tab stays a button: MUI only forwards `disabled` to real buttons, so an anchor would still be followable.

**Table rows** wrap the first cell's content in the slug-aware `Link`, giving each row a real, org-prefixed anchor while keeping the row-level click for everything else. The anchor is `tabIndex={-1}` so the row stays the single tab stop, and it stands down when the click came from a control nested in that cell (inline copy buttons). Row click suppression now uses `closest('a, button')` rather than an `instanceof` check on the event target, so a click landing on a label inside a cell button no longer navigates the row.

**Row action menus**: `ActionItem` becomes a discriminated union of `onAction` and a new `link`, rendered through `ButtonLink`; the 18 row-menu entries that only navigated now declare a link.

Two follow-ons the row anchor required:

- `onRowActionLink` is called while rendering to build the href, so it must be pure. Five call sites hid a side effect in it (the log tables' mobile "slide forward", the activity logs panel). Those move to `onRowActionClick`, which now runs alongside the navigation instead of replacing it.
- Pressing Enter on a focused row built the link and discarded it, so keyboard activation did nothing. It now navigates.

The tabs keep `role="tab"` inside the MUI `tablist`: they are real anchors now, so open-in-new-tab, middle-click, copy-link and the status-bar preview all work, without breaking the tab pattern the surrounding markup declares.

Fixes LAGO-1916

https://claude.ai/code/session_01PMSrxK3psmBAPAmD6jsrn2

---
_Generated by [Claude Code](https://claude.ai/code/session_01PMSrxK3psmBAPAmD6jsrn2)_


domenicofalco:
### `Run Test E2E` is failing repo-wide, not on this PR

The job dies in the **API seed step**, before Cypress starts (no screenshots or videos are produced):

```
bin/rails aborted!
ArgumentError: wrong number of arguments (given 0, expected 1) (ArgumentError)
    callee: /app/app/services/catalog_plans/create_service.rb:7
    |     def initialize(args, send_webhook: true)
/app/app/services/catalog_plans/create_service.rb:7:in 'CatalogPlans::CreateService#initialize'
/app/app/services/base_service.rb:178:in 'BaseService.call'
/app/db/seeds/30_product_catalog.rb:103:in '<main>'
```

`CatalogPlans::CreateService` takes a positional `args`, but `BaseService.call` invokes it with none — a `lago-api` arity mismatch in `db/seeds/30_product_catalog.rb`.

**Why it is not this PR's:** the diff here is frontend TypeScript only and touches nothing that runs before Cypress boots. The same job fails identically on an unrelated PR on the same base:

| Run | PR | Duration | Result |
| --- | --- | --- | --- |
| [34198351210](https://github.com/getlago/lago-front/actions/runs/34198351210) | this PR (#4277) | 07:14:29Z → 07:19:42Z (5m13s) | failure, same seed error |
| [34199432811](https://github.com/getlago/lago-front/actions/runs/34199432811) | #4272 (unrelated) | 07:27:38Z → 07:32:49Z (5m11s) | failure |

**No fix to port:** the change belongs in `lago-api`, which is outside this repo. I have not re-run the job — a Ruby arity error is deterministic, so a re-run would only burn a cycle.

Not treating this as a flake and not disabling anything. The remaining check on this PR, the SonarCloud coverage gate, is mine and I am pushing a fix for it.

---

domenicofalco:
### Correction: the API seed is fixed, and E2E then found a real regression here

My earlier comment said `Run Test E2E` was failing for reasons outside this PR. That was accurate for attempts 1 and 2, but it no longer holds — updating the record.

On **attempt 3** (12:18Z) the `Populate db` step passed at 12:23:34Z, so the `lago-api` seed has been fixed upstream. Cypress then actually ran for 7m17s and surfaced a genuine regression from this PR:

```
AssertionError: Timed out retrying after 4000ms:
Expected to find element: `button[role="tab"]`, but never found it.
  at cypress/e2e/10-resources/t20-create-customer.cy.ts:43:12
```

`10-resources/t20-create-customer.cy.ts` selected the settings tab as `button[role="tab"]`. This PR turns navigation tabs into anchors, so the tag no longer matches — the role does, but the selector was coupled to the element type.

Fixed in `707c73c` by dropping the tag: `cy.get('[role="tab"]')`. The three other tab selectors in the suite (`t80`, `t90` ×2) already query `[role="tab"]` alone, so this was the only coupled one.

Worth noting the rest of the run was clean: **52 of 56 tests passed, 3 pending, 1 failing** — the only failure was this selector. The 16 other specs, including the ones that navigate tables, passed against the new row anchors.

---

domenicofalco:
Confirmed and applied in `6ae9e7f`.

`ActionItemButton` took the `link` branch regardless of `disabled`, and `ButtonLink` keeps its `href` — `pointer-events-none` only blocks pointer hit-testing, so activation from assistive tech or script could still follow a route a disabled button never exposed. Four entries were affected: the two credit-note actions (`InvoicesList`, `CustomerInvoicesList`), the dunning campaign edit and the invoice custom section edit, all gated on permissions.

A disabled action now renders as a disabled `Button` whatever its shape, which is also what `NavigationTab` does for a disabled tab. Regression test asserts the open menu exposes no `link` role and no `href` at all, only a disabled button; verified by inverting the guard — the test fails without it.

---

ansmonjol (approved):

domenicofalco:
Swept the automated pre-review feedback on this PR. Every HOLD except one was addressed by a later commit, and the most recent pre-review verdict (`fde08a0`) is PASS with all 14 checks green on the current head.

The one point that stays open is the `<a><button>` nesting in row-menu link actions, raised on `9aa4b96` and again on `17c3e5b`. The finding is correct: `ButtonLink` renders `<Link>` wrapping `MuiButton`, so the anchor contains interactive content — which is exactly the rule `Table.hasInteractiveContent` enforces for first-cell content in this same PR.

Leaving it as is here: that structure is pre-existing in `ButtonLink` and shared by ~60 call sites (sidebar, settings nav, wallet and role details, auth pages). Making the anchor the single control means reworking the design-system component itself, which is well beyond the scope of LAGO-1916 and would touch every one of those call sites. Better as a follow-up on `ButtonLink` — @domenicofalco will decide whether to split that off.

---

PR #4282, opened 2026-09-09
fix(catalog): relations grid and rate card field locking

ansmonjol:
## Context

A batch of adjustments on the product catalog detail pages and drawers. Three
things were off: the product category / product / product filter rows were
duplicated across four overview components and had drifted apart, rate card
field locking did not match what the backend actually accepts on update, and a
duplicate-code rejection could land under the Code input even when the conflict
was on another field.

## Description

**Shared relations grid.** The four detail overviews (`ProductDetailsOverview`,
`ProductFilterDetailsOverview`, `RateCardDetailsOverview`,
`RateCardRateDetailsOverview`) each built their own product category / product /
product filter rows. They had drifted: only two of them preferred
`invoiceDisplayName` over `name`, only two carried a `data-test` on the
"no category" placeholder, and two rendered a dash for a missing product filter
while the others omitted the row. `CatalogRelationsInfoGrid` now owns all of it,
and the fragments select `invoiceDisplayName` everywhere the link is rendered.

Two user-visible consequences, both intentional: the product filter row is now
hidden when there is no filter rather than showing a dash, and the pricing unit
row on the rate card overview always renders (with a dash) instead of
disappearing when no pricing unit is set.

**Rate card locking.** `isLocked` conflated two distinct backend rules. Per
`RateCards::UpdateService`, `LOCKED_WITH_RATES` freezes the billing-semantic
fields once a rate exists, and the currency additionally freezes once the rate
card is attached to a plan or subscription. Split into `hasRates` (from the new
`ratesCount` selection) and `isAttached`. The old flag also froze the code
input, which the backend does accept on update, so `code` is now sent on both
rate card and rate card rate updates and the input is only disabled on
attachment.

**Duplicate-code errors.** `hasDefinedGQLError('ValueAlreadyExist', errors)`
matches any field in `extensions.details`, so a conflict on a different field
surfaced under the Code input. All four catalog drawers now pass the `'code'`
key, matching the pattern already used in `CreateCustomer` and
`useEditOrganizationSlugDialog`. The rate card rate code field also clears the
error as the user types, which the other drawers already did.

**Smaller items.** `NameAndCodeGroup` accepts `nameDataTest` / `codeDataTest` so
the code inputs are addressable from tests; the rate card zod schema uses the
shared required-field message instead of three keys, one of which was borrowed
from the coupon form; small chips get 8px horizontal padding instead of 4px.

`src/generated/graphql.tsx` also picks up unrelated schema drift from the
regeneration (the `catalogPlan` / `catalogPlans` queries, `Contract.plan`
retyped to `CatalogPlan`, and `productCategoryId` dropped from `plans`).
Verified that nothing in `src/` passed `productCategoryId` to the `plans` query,
and the typecheck is clean.

### DIFF
--- a/src/components/form/NameAndCodeGroup/NameAndCodeGroup.tsx
+++ b/src/components/form/NameAndCodeGroup/NameAndCodeGroup.tsx
@@ -14,6 +14,8 @@ type NameAndCodeGroupProps = {
   disableAutoGenerateCode?: boolean
   nameProps?: Partial<TextInputProps>
   codeProps?: Partial<TextInputProps>
+  nameDataTest?: string
+  codeDataTest?: string
 }
 
 const defaultValues: NameAndCodeGroupValues = {
@@ -35,6 +37,8 @@ const NameAndCodeGroup = withFieldGroup({
     disableAutoGenerateCode,
     nameProps,
     codeProps,
+    nameDataTest,
+    codeDataTest,
   }) {
     const { translate } = useInternationalization()
 
@@ -68,6 +72,7 @@ const NameAndCodeGroup = withFieldGroup({
               label={translate('text_629728388c4d2300e2d38091')}
               placeholder={translate('text_629728388c4d2300e2d380a5')}
               {...nameProps}
+              data-test={nameDataTest}
             />
           )}
         </group.AppField>
@@ -77,8 +82,9 @@ const NameAndCodeGroup = withFieldGroup({
               label={translate('text_629728388c4d2300e2d380b7')}
               beforeChangeFormatter="code"
               placeholder={translate('text_629728388c4d2300e2d380d9')}
-              disabled={disableCodeInput}
               {...codeProps}
+              disabled={disableCodeInput}
+              data-test={codeDataTest}
             />
           )}
         </group.AppField>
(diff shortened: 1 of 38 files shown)


ansmonjol:
Applied in e3386b3d6 — the pre-review was right, and it under-reported by one site.

`configSnapshot` in `MainHeaderConfigure.tsx` drops any value with `$$typeof`, so `entity.metadata` never reaches the serialized string: a code-only change produces a byte-identical snapshot, the `useLayoutEffect` dep never moves, `setConfig` is skipped and the header keeps the old element. `MainHeaderConfigure.test.tsx` ("WHEN content changes but no snapshotKey is set") already pins exactly that behaviour for the analogous `content` case. Same mechanism strips the action `onClick`s, so the edit closure keeps handing the drawer a stale `ratesCount`.

Both fields are now in the key, extracted as `buildRateCardSnapshotKey` next to the existing `buildRateCardRateSnapshotKey`, with unit tests covering code, rate count, description, attachment and stability.

The extra site: `RateCardRateDetails` renders the *parent card's* code in its own `entity.metadata`, but `buildRateCardRateSnapshotKey` only tracked `rate?.code`. Added `rateCard?.code` there too, with its own test.

domenicofalco (approved):

PR #4294, opened 2026-09-11
refactor(forms): unify the duplicate-code error handling across code fields

domenicofalco:
## Context

"The backend rejected this code as already existing" was implemented eight different ways across the app, split over three competing errorMap conventions: `setFieldMeta` on `onDynamic` with a manual clear, `setFieldMeta` on `onSubmit` with a manual clear, and `setErrorMap` with no clear. Only the product-catalog rate drawer had been unified, on purpose, to keep that diff reviewable.

Two of those conventions are also broken. Verified against `@tanstack/form-core@1.33.0`: `setFieldMeta` does not stamp `errorSourceMap`, and `determineFormLevelErrorSourceAndValue` (`dist/esm/utils.js:266-281`) preserves any error whose source is not `'form'`. These forms put the schema on the form and declare no validator on the `code` field, so the field-level clearing pass never runs and the manual error is permanent. Every site that set it therefore had to hand-roll a clear, and each hand-rolled clear was a separate copy of the same message-gated check. Where the clear was missing — `usePlanForm` — the submit button stayed disabled for the life of the form, since `SubmitButton` gates on `canSubmit`.

`setErrorMap` (`FormApi.js:828-849`) stamps `errorSourceMap[key] = 'form'`, so an error applied that way is dropped by the next validation on its own. Measured both conventions on a form wired like the real ones (schema on `onDynamic`, `revalidateLogic`): the `setFieldMeta` error survives the edit with `canSubmit === false`, the `setErrorMap` one clears and submit re-enables.

## Description

Reshaped `applyExistingCodeError` onto `setErrorMap`, which removes the need for a clearing counterpart anywhere, and deleted `clearExistingCodeError`.

Collapsed every hand-rolled implementation onto it — `NameAndCodeGroup`, `ChargeCodeField` and `CreateCustomSection` as the ticket asks, plus five more copies found on the way: `usePlanForm`, `CreateTax`, `CreateCoupon`, `CreateAddOn`, `CreatePricingUnit` and `CreateWallet`. That drops five `codeValue` + `useEffect` clearing pairs, `CreateAddOn`'s `setCodeExistsError`, and `CreateCustomSection`'s mount-run guard, whose only reason to exist was the clear it guarded. `CreatePricingUnit` also stops pre-translating the message, which the field wrapper translates again.

Folded the last hardcoded copy of `text_632a2d437e341dcc76817556` onto the exported `EXISTING_CODE_ERROR_MESSAGE` in `alerts/utils.ts` and `CreateCustomSection` — the two the ticket names. `chargeCode.ts` no longer re-exports the helpers; the two charge drawers import them from `~/core/form/existingCodeError` directly.

The ticket's fifth point — give `NameAndCodeGroup` a code-only mode, or have it and `ChargeCodeField` share one listener — is resolved by not needing a listener in either: both are now plain fields.

### Tests

`existingCodeError.test.tsx` was rewritten around the real cycle (submit → backend rejects → error shown and submit blocked → edit → error dropped and submit re-enabled → clearing the code hands over to the schema error), which is what guards the stuck-submit bug. A new `usePlanForm` suite pins the same thing on the plan form, the one site that had no clear at all. `CreateTax`, `CreateAddOn`, `CreateCoupon` and `CreateWallet` each gained a duplicate-code case.

The component suites that asserted the removed clearing listeners were reduced to what those components still own, and `CreateCustomSection`'s and `CustomerConnectionDrawer`'s error cases now drive an actual submit first — without one, `revalidateLogic` is still in submit mode and no validation runs, so they were asserting against a state the app never reaches.

Coverage on new code: 13/13.

Also updated the `lago-forms` skill, which documented the convention being removed.

### Deliberately not included — tracked in LAGO-1933

**Scoping the ~35 unkeyed `hasDefinedGQLError('ValueAlreadyExist', errors)` calls** to their field (ticket point 4). A key narrows the check to `details[key]`, so passing `'code'` where the backend reports the collision under another name silently stops the error from showing. Nothing in this repo says which `details` key each mutation returns, and the five catalog drawers that already pass `'code'` were confirmed against the API, not inferred. Doing the sweep blind would trade a misplaced error for a missing one; it needs the per-mutation answer from the backend.

**The thirteen inlined `setErrorMap` copies in `src/components/settings/integrations/Add*Dialog.tsx`** (raised by the pre-review bot below, and correct). Those already use the convention this PR standardises on, so they self-clear properly today — the duplication is worth removing, but it is not a defect. Eleven of the twelve files have no test file and nothing in the repo exercises a `ValueAlreadyExist` path in any of them, so folding them in adds thirteen uncovered lines and re-fails the coverage gate that the second commit here just fixed. They sit on the same files as point 4, so LAGO-1933 does both with the tests they need.

**The four Formik pages** that still hardcode the message key (`FeatureForm`, `BillingEntityCreateEdit`, `CreateDunning`, `CreateBillableMetric`). Folding them onto the constant changes no behaviour and cannot be covered honestly — Formik's `TextInputField` renders a field error only once the field is touched, and touching it re-runs validation, which drops a manually-set error. That belongs with the Formik→TanStack migration of those pages, where the error handling is rewritten and the pages get tests.

Fixes LAGO-1925

### DIFF
--- a/.agents/skills/lago-forms/SKILL.md
+++ b/.agents/skills/lago-forms/SKILL.md
@@ -159,13 +159,15 @@ formApi.setErrorMap({
 })
 ```
 
-Clear the error when the user edits the field, or submit stays disabled — setting and
-clearing are two separate jobs. For a duplicate `code`, `applyExistingCodeError(formApi)`
-(`~/core/form/existingCodeError`) does only the setting half, in the mutation catch on
-`LagoApiError.ValueAlreadyExist`. The clearing half belongs to the field, gated on
-`EXISTING_CODE_ERROR_MESSAGE` so it does not wipe the zod required-check — already
-implemented in `NameAndCodeGroup` and `ChargeCodeField`, so mount one of those rather
-than re-wiring the listener.
+Prefer `setErrorMap` for a server error: it stamps the error as form-owned, so the next
+validation drops it and submit re-enables on its own. A `setFieldMeta` error reads as
+field-owned, and on these forms — schema on the form, no validator on the field — nothing
+ever clears it, so submit stays disabled for the life of the form.
+
+For a duplicate `code` that is `applyExistingCodeError(formApi)`
+(`~/core/form/existingCodeError`), called in the mutation catch on
+`LagoApiError.ValueAlreadyExist`. It needs no clearing counterpart; do not add a listener
+to clear it.
 
 Validator-produced errors live directly on `errorMap.onDynamic` keyed by field path; the
 `.fields` sub-shape exists only for errors set manually. Reading section validity for
--- a/src/components/alerts/utils.ts
+++ b/src/components/alerts/utils.ts
@@ -1,3 +1,6 @@
+import { AnyFormApi } from '@tanstack/react-form'
+
+import { applyExistingCodeError } from '~/core/form/existingCodeError'
 import { deserializeAmount } from '~/core/serializers/serializeAmount'
 import { AlertThreshold, CurrencyEnum, ThresholdInput } from '~/generated/graphql'
 
@@ -91,23 +94,12 @@ export const createThresholdSetters = (form: ThresholdsFormApi): ThresholdSetter
   },
 })
 
-/** The slice of an alert form the duplicate-code error handler relies on. */
-type CodeErrorFormApi = {
-  setErrorMap: (errorMap: {
-    onDynamic: { fields: { code: { message: string; path: ['code'] } } }
-  }) => void
-}
-
 /**
  * Marks the code field with the "value already exists" error and scrolls back
  * to it, as both alert forms do when the API rejects a duplicate code.
  */
-export const setCodeAlreadyExistsError = (formApi: CodeErrorFormApi): void => {
-  formApi.setErrorMap({
-    onDynamic: {
-      fields: { code: { message: 'text_632a2d437e341dcc76817556', path: ['code'] } },
-    },
-  })
+export const setCodeAlreadyExistsError = (formApi: AnyFormApi): void => {
+  applyExistingCodeError(formApi)
 
   document.getElementById('root')?.scrollTo({ top: 0 })
 }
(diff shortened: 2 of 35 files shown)


domenicofalco:
On the pre-review HOLD: the finding is factually correct, and I'm not disputing it. Verified — thirteen call sites across those twelve files (`AddAvalaraDialog` has two) inline the identical shape, all of them also pre-translating the message:

```
setErrorMap({ onDynamic: { fields: { code: {
  message: translate('text_632a2d437e341dcc76817556'), path: ['code'],
} } } })
```

I'm still leaving them for a follow-up, for two reasons.

**They are duplication, not a defect.** Those thirteen already use `setErrorMap`, which is the convention this PR standardises on. They self-clear correctly today. What this PR fixes is the `setFieldMeta` sites, where the error was never cleared and the submit button stayed disabled — none of the integration dialogs are in that group. So folding them in removes repetition; it does not change behaviour or fix a bug.

**It cannot ride along without failing the quality gate.** Eleven of the twelve files have no test file, and no test in the repo exercises a `ValueAlreadyExist` path in any of them. Replacing the inline blocks adds thirteen uncovered lines against this PR's current 13/13 on new code, taking it to roughly 50% and re-failing the gate that the second commit here just fixed. Clearing it means building duplicate-code harnesses for about seven integration dialogs — a bigger change than the one this ticket describes, and one that would bury the form-core fix that is the actual point of the PR.

It also pairs naturally with the other deferred item. Those same dialogs are among the unkeyed `hasDefinedGQLError('ValueAlreadyExist', errors)` call sites from LAGO-1925's point 4, so one follow-up can route them through `applyExistingCodeError`, scope their error check to the right field once the backend confirms the `details` key, and add the tests that both changes need. I'll raise that ticket.

Happy to pull it in here instead if a reviewer would rather have it in one go — say the word and I'll add the dialog tests alongside.

---

ansmonjol (approved):

PR #4295, opened 2026-09-11
fix(invoices): pin date picker to its declared timezone

ansmonjol:
## Context

On the one-off invoice billing period, picking a start date recorded the day before the one clicked (click 17 Sep, get 16 Sep), for a user in a timezone ahead of UTC. The end date was unaffected, and a second click on the same day corrected it.

## Description

`DatePicker` declares a `defaultZone` but only applied it by mutating the global `Settings.defaultZone` from an effect, so its first render parsed the value in the ambient zone and MUI froze its calendar on that zone. Clicking a day then carried the local time-of-day onto it, and publishing `.toUTC()` landed one UTC day off (one day back in zones ahead of UTC, one day forward in zones behind it). The value is now parsed in `defaultZone` and the zone is pinned on the MUI picker, which only changes behaviour for the consumers that already pass `defaultZone`. On top of that, `FeesSection` seeded the fee period from the ambient zone while the row and the picker both render it in UTC, and `EditFeeBillingPeriod` normalised the day bounds against the ambient zone; both are now explicit. Covered by three regression tests that each fail on `main`.

Fixes BIL-677

### DIFF
--- a/src/components/form/DatePicker/DatePicker.tsx
+++ b/src/components/form/DatePicker/DatePicker.tsx
@@ -79,12 +79,15 @@ export const DatePicker = ({
   /**
    * Date will be passed to the parent as ISO
    * So we need to make sure to re-transform to DateTime for the component to read it
+   *
+   * Parsed in `defaultZone` rather than the ambient zone: the `Settings.defaultZone`
+   * effect below only runs after the first render, and MUI freezes the calendar on it.
    */
   const getValueFormatted = useCallback(() => {
     if (!value) return null
 
-    return typeof value === 'string' ? DateTime.fromISO(value) : value
-  }, [value])
+    return typeof value === 'string' ? DateTime.fromISO(value, { zone: defaultZone }) : value
+  }, [defaultZone, value])
 
   const [localDate, setLocalDate] = useState<DateTime | null>(getValueFormatted())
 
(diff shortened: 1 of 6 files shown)


domenicofalco:
Pushed a test-only commit (e517f16ae) — the fix itself looks right to me, but the coverage was arguing for it from a state the fix makes unreachable.

The end bound is the strongest case here: it is stored at `23:59:59.999Z`, so in any zone ahead of UTC it renders on the next day, and on the last day of a month the calendar opens on the next month. Picking day 17 on the `to` picker, with a period already on the UTC day:

- without the `DatePicker` change: `2026-10-16T23:59:59.999Z`
- with it: `2026-09-17T23:59:59.999Z`

So the dialog test now opens on a UTC-day period (what `FeesSection` seeds after your fix) and drives both calendars. The previous premise, a period seeded at ambient-zone midnight, cannot happen once `FeesSection` seeds in UTC.

Also moved the `Settings` restore into `afterEach` so a throwing await cannot leak the zone into the sibling tests, and dropped the comment in `FeesSection`.

Two things I hit while checking: `DatePicker.test.tsx` fails on a machine in Europe/Paris on `main` too (LMT offset on a year-26 date) — pre-existing, green in CI. And the `DatePicker` change reaches every consumer passing `defaultZone` (subscription dates, rate card rate, all date filters), so that is where I would put the manual QA.

domenicofalco (approved):

PR #4298, opened 2026-09-11
feat(catalog): plan list, drawer and details page

ansmonjol:
Fixes BIL-261

Adds the catalog-plan area under `/plan-pricing`: the list, the create/edit drawer, and the plan details page. Second real entity in the Products & rate cards initiative after rate cards, and it follows the same shape as the five sibling catalog entities.

Everything is behind `FeatureFlagEnum.ProductCatalog`, and permissions reuse the legacy `plans*` set because that is what the server enforces (`plans:view` / `create` / `update` / `delete` on every catalog-plan resolver and mutation).

## Plan list

Full-page list at `/plan-pricing`: search, numbered pagination, four columns (name + code, `# rate cards`, `# subscriptions`, creation date), and empty / search-empty / error states.

Layout follows `SubscriptionsPage` rather than the product-catalog tab lists: `insetPager`, sticky pager, `containerSize={{ default: 16, md: 48 }}`, and the search passed as `MainHeader.Configure`'s `filtersSection`. Those tab lists use an inset wrapper and `network-only` because a route-based tab remount would otherwise flash a stale cached page; a full-page list does not remount that way.

## Create / edit drawer

`useFormDrawer` + `useAppForm` + zod, dual-mode, with the shared "Create more" system.

The field locks mirror what the API actually enforces, and nothing more:

| field | locked when |
| --- | --- |
| `name`, `description`, `invoiceDisplayName` | never |
| `code` | plan attached to contracts |
| `currency` | attached to contracts, or already priced by rate cards |

Two consequences worth knowing when reading the diff:

- **The update payload deliberately sends every field**, including an unchanged `code` and `currency`. The backend's locks are change-gated (they compare against the stored value) and it has a spec pinning that resubmitting an unchanged code alongside an editable field succeeds. Diffing fields out would be the intuitive move and would be wrong.
- **The submit handler ends with a generic danger toast** when the mutation returns no plan and the error was not `ValueAlreadyExist`. `plan_locked` and `not_editable_with_applied_rate_cards` arrive as silenced `unprocessable_entity` details and are not `LagoApiError` members, so `hasDefinedGQLError` cannot name them. Without the fallback an unrecognised rejection is swallowed and the save silently appears to do nothing, which is a hole the five sibling drawers still have.

## Details page

Three tabs (Overview, Contract, Activity logs). The Overview tab carries a second navigation level: a static left sub-nav that swaps the right-hand pane, driven by a `:section` path segment so each view is deep-linkable and starts at the top.

Two details worth a reviewer's attention:

- **`snapshotKey` includes `code`.** The `MainHeader` config snapshot strips functions and React elements, so a field being *displayed* is not the same as it being *encoded*. The code is rendered as `entity.metadata` (a ReactNode), so leaving it out meant a code-only edit left the Copy-code and Edit-plan closures on the old value, the next save silently reverted the rename, and the Contract tab kept querying the stale `planCode`. This was caught in review, not in testing.
- **The Overview tab declares `match: [overviewPath, \`${overviewPath}/:section\`]`.** `matchPath` is exact by default, so the bare path alone does not match `/overview/rate-cards`; without the second entry, matching falls through and `useMainHeaderTabContent` silently renders the first visible tab. A content-based test cannot catch that because the fallback masks it, so the test asserts the array itself.

Activity logs work with no new backend surface: catalog plans log under the existing `plan_created` / `plan_updated` / `plan_deleted` types, which `useActivityLogsInformation` already maps.

## Deliberately out of scope

The rate-cards section is an inert empty state. `planAppliedRateCards` and `createPlanAppliedRateCard` both exist and are catalog-plan scoped, so a real list and attach flow are buildable, but neither has a design yet. The "Add rate card" and Quick-actions `+` controls are placeholders for that flow.

## Two changes that are not catalog-plan feature code

- `NavigationTabBar` now renders linked tabs as anchors, so cmd-click and middle-click open a tab in a new browser tab. Disabled tabs stay buttons, because `disabled` on an anchor is a no-op for assistive tech.
- `DEBOUNCE_SEARCH_MS` is `0` under Jest as well as Cypress. It was 500ms under Jest, so every test exercising a debounced search burned half a second of real time. The 8-file verification set went from ~12.1s to ~9.1s, and it removed an obsolete `asyncUtilTimeout: 5000` override from two preview suites. Happy to split either out if you would rather review them separately.


src/pages/catalog/details/CatalogPlanDetails.tsx
```diff
@@ -0,0 +1,166 @@
...
+                <CatalogPlanDetailsOverview
+                  rateCardsCount={catalogPlan?.appliedRateCardsCount}
+                  loading={loading}
+                />
+              </DetailsPage.Container>
+            ),
+          },
+          {
+            title: translate('text_17891318128636r6g9igqqeq'),
+            link: buildTabLink(CatalogPlanDetailsTabsOptionsEnum.contracts),
```
domenicofalco, line 141: [P2] Gate this tab on `contractsView`: users with `plansView` but without `contracts:view` can open it, then the contracts query returns `forbidden`. Add `contractsView` to `MembershipPermissions` as well—it is not fetched yet.

src/pages/catalog/dialogs/useDeleteCatalogPlanDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+        const destroyedId = data?.destroyCatalogPlan?.id
+
+        if (!destroyedId) return
+
+        evictFromCache(client, {
+          id: destroyedId,
+          __typename: 'CatalogPlan',
+          listFieldName: 'catalogPlans',
+          listQueryDocument: [CatalogPlansDocument],
+        })
```
domenicofalco, line 65: [P2] Refresh the list after deletion. `evictFromCache` removes the row but leaves pagination metadata unchanged and does not trigger a refetch. Deleting the only plan on the last page leaves an empty page with stale totals, so `PaginatedContent` cannot move back to the previous page.

src/pages/catalog/details/CatalogPlanRateCardsSection.tsx
```diff
@@ -0,0 +1,41 @@
...
+        action={{
+          title: translate('text_1789030049529b0zmy1slfxl'),
+          dataTest: CATALOG_PLAN_ADD_RATE_CARD_TEST_ID,
+          onClick: () => undefined,
+        }}
+      />
+
+      <Table
+        name="catalog-plan-rate-cards"
+        data={[]}
```
AllanMichay, line 33: The Overview sub-nav renders the live `appliedRateCardsCount` right next to this always-empty table, so the two contradict each other.

A plan with `appliedRateCardsCount: 2` shows "2" on the `rate-cards` nav item (`CatalogPlanOverviewNav.tsx:58`/`107`), and clicking it lands on this section, which says "No rate cards have been added yet". `CatalogPlanOverviewSection` also hides the "Assign your first rate cards" quick action once the count is `> 0`, so the user ends up with a "2", an empty list, and no affordance at all.

I know the section is deliberately inert until the design lands — the issue is only that the count beside it is live. Either suppress the nav count while the section is a placeholder, or word the empty state as "not available yet" rather than "none added".

src/components/customerConnections/__tests__/ProviderSelectionSection.test.tsx
```diff
@@ -150,9 +150,8 @@ describe('ProviderSelectionSection', () => {
 
         // Options themselves are virtualized (not mounted in jsdom):
         // the open state is the behaviour under test
-        // 1s default: the popper loses the race under a loaded parallel run
-        expect(await screen.findByRole('listbox', {}, { timeout: 5000 })).toBeInTheDocument()
-        expect(input).toHaveAttribute('aria-expanded', 'true')
+        await waitFor(() => expect(input).toHaveAttribute('aria-expanded', 'true'))
```
AllanMichay: This drops the `asyncUtilTimeout: 5000` headroom and falls back to RTL's 1s default, which reinstates the exact flake the removed comment documented:

> 1s default: the popper loses the race under a loaded parallel run

On a loaded 4-shard runner the MUI Popper can mount more than 1s after the click, and this `waitFor` times out. Nothing in the catalog-plan feature touches this file's timing, so the change is pure risk here — worth keeping the explicit timeout on the `aria-expanded` wait.

Worth noting the current run already has `Tests (shard 2/4)` red on a popper race in `EditQuoteAside.test.tsx` (unrelated file, pre-existing pattern).

src/hooks/useDebouncedSearch.ts
```diff
@@ -3,7 +3,7 @@ import { debounce, DebouncedFunc } from 'lodash'
 import { DateTime } from 'luxon'
 import { useCallback, useEffect, useRef, useState } from 'react'
 
-export const DEBOUNCE_SEARCH_MS = window.Cypress ? 0 : 500
+export const DEBOUNCE_SEARCH_MS = window.Cypress || process.env.NODE_ENV === 'test' ? 0 : 500
```
AllanMichay: `DEBOUNCE_SEARCH_MS` is not search-only — `useCreditNoteFormCalculation.ts:141` uses it as the debounce for the `creditNoteEstimate` query, and `CustomerInvoicesTab` / `CustomerCreditNotesList` use it for their own local debounces.

Under Jest that estimate now fires once per fee edit instead of once per burst, so the test environment behaves as if there were no debounce at all. A future regression that actually breaks the coalescing in production (an unstable `debouncedQuery` identity, say) would be invisible to the suite.

The speed-up is worth having — it just wants a narrower seam: a search-specific constant, or the `0` override applied inside `useDebouncedSearch` only, leaving the exported constant at 500 for the other consumers.

src/pages/catalog/details/CatalogPlanRateCardsSection.tsx
ansmonjol: Yea it's in my plan to ship that as a whole in next iteration. It's fine to have it like this for now

src/pages/catalog/details/CatalogPlanDetails.tsx
ansmonjol: Done in 04ba387ad. `contractsView` is now part of the `MembershipPermissions` selection, and the tab carries `hidden: !hasPermissions(['contractsView'])`.

Checked the resolver to be sure of the failure mode: `Resolvers::ContractsResolver` declares `REQUIRED_PERMISSION = "contracts:view"`, so an ungated tab opened straight onto `forbidden`. And the second half mattered as much as the first: `hasPermissions` maps an unfetched permission to `false`, so gating without adding the field to the fragment would have hidden the tab from everyone.

There is a test that fails without the `hidden` line.

src/pages/catalog/dialogs/useDeleteCatalogPlanDialog.tsx
ansmonjol: Fixed in 104e3af37, but in `evictFromCache` rather than here.

The mechanism you describe is exactly right. What made me move the fix is that it isn't specific to this dialog: all five catalog delete dialogs go through the same helper, and `ProductCategoriesList.tsx:132` even carries a comment stating that no refetch is needed. A refetch wired into this one call site would fix plans and leave the other four with the same empty last page.

So the helper now keeps `metadata` in step with the collection it just shortened: `totalCount` is decremented, and when the removal empties a page past the first, `totalPages` drops to `currentPage - 1`, which is precisely the condition `PaginatedContent`'s out-of-range clamp waits for. It pages back on its own, with no extra round trip, and the optimistic removal you get today is preserved.

`evictFromCache.test.ts` covers the three shapes: a page that keeps rows, a page emptied past the first, and the last row of the whole list.

src/components/customerConnections/__tests__/ProviderSelectionSection.test.tsx
ansmonjol: Reverted in d34365be5 - the file is byte-identical to `origin/main` again, explicit `timeout: 5000` and the comment included.

src/hooks/useDebouncedSearch.ts
ansmonjol: Reverted in d34365be5. `DEBOUNCE_SEARCH_MS` is back to `window.Cypress ? 0 : 500`.

The blast radius is as you describe. Consumers I confirmed: `useCreditNoteFormCalculation.ts:141` for the `creditNoteEstimate` debounce, `CustomerInvoicesTab.tsx:109` and `CustomerCreditNotesList.tsx:110` for their own.

It also had a third casualty beyond the two I had already patched: `EditQuoteAside.test.tsx`, which was the red `Tests (shard 2/4)` on this run. Same popper race you point at. Reverting the line takes it back to 55/55.

Rather than narrow the seam here, I dropped the speed-up from this PR entirely - it can come back on its own, where the reduced coverage you describe can be weighed properly.

domenicofalco (approved):

PR #4300, opened 2026-09-14
refactor(form): stop DatePicker from mutating the global Luxon default zone

domenicofalco:
## Context

`DatePicker` wrote the process-wide `Settings.defaultZone` from a mount effect. A form component owning global state had three consequences: the empty dependency array never re-applied a changed `defaultZone`; two pickers declaring different zones fought over it, and the first unmount reset the ambient zone while the other was still mounted; and while any picker was mounted, every other ambient-zone reader in the app (`DateTime.now()`, a zone-less `DateTime.fromISO()`) ran in a zone chosen by a form component. That was the root cause behind BIL-677.

BIL-677 made the picker zone-correct locally — it parses its value in `defaultZone` and pins `timezone` on the MUI picker — so the global write has no remaining job.

## Description

Removes the mount/cleanup effect, along with the `useOrganizationInfos` read and the `OrganizationForDatePicker` fragment it was the only consumer of. The fragment was spread into `MainOrganizationInfos`, which already selects `id` and `timezone` directly, so the query's selection set is unchanged.

Audited the ambient-zone readers that were implicitly relying on the write and pinned the two that were:

- `getBillingTimeHelperKey` derives the anniversary day from `subscriptionAt`, a UTC calendar day. Read in the ambient zone it shifted a day early for any organization west of UTC.
- `SubscriptionDatesOffsetHelperComponent` compared `subscriptionAt` against today's start of day, so the same date read as future in an organization ahead of UTC.

Verified as already explicit or instant-based (zone-independent), so unchanged: `EditFeeBillingPeriod`, `RateCardRateDrawerContent` and its utils, `subscriptionFormSchema`, `useUpdateSubscriptionForm`, `CreateSubscription`. `FiltersItemIssuingDate` deliberately declares no zone and is org-zone-relative; it now reliably gets the organization's zone instead of whatever a sibling picker last wrote.

**One deviation from the ticket's scope:** it asked to drop `parseBound` in `DateRangeFilterFields` "once it is provably redundant". It is not — with the effect gone the ambient zone is *always* the organization's and never the picker's declared one, so parsing the bounds in `defaultZone` is permanently load-bearing rather than a first-render workaround. `parseBound` is kept; its comment and the one in `DatePicker` are rewritten to stop describing the effect.

Three regression tests cover the change, each verified to fail on `main`:

- `DatePicker` leaves the ambient zone untouched across mount and unmount.
- `getBillingTimeHelperKey` keys the anniversary on the UTC calendar day under an ambient zone behind UTC.
- `SubscriptionDatesOffsetHelperComponent` renders the same helper text whatever the ambient zone.

The ticket also calls for a QA round with an organization timezone ahead of UTC (e.g. `Europe/Paris`) — with the organization on UTC this whole class of defect is invisible.

Fixes BIL-686


src/components/form/DatePicker/DatePicker.tsx
```diff
@@ -104,17 +90,6 @@ export const DatePicker = ({
     return helperText
   }, [error, helperText, isInvalid, showErrorInTooltip, translate])
 
-  useEffect(() => {
-    if (defaultZone) Settings.defaultZone = defaultZone
```
ansmonjol, line 108: So we can remove the `defaultZone` then? How do we configure the timezone now?

domenicofalco: No — `defaultZone` stays, and it is still exactly how you configure the picker's timezone. What this PR removes is only the *mechanism* it used to reach the picker through, not the prop.

Before #4295 (BIL-677), `defaultZone` was never passed to MUI at all. The effect wrote it into the process-wide `Settings.defaultZone`, and MUI picked it up from the ambient Luxon zone — which is why it leaked into every other `DateTime.now()` / zone-less `DateTime.fromISO()` in the app while a picker was mounted.

#4295 made the prop apply locally, in two places that are both on `main` today:

- `DatePicker.tsx:75` — the incoming ISO value is parsed in `defaultZone`
- `DatePicker.tsx:133` — `timezone={defaultZone}` pins MUI's calendar to it

With those two in place the effect is dead weight: it only sets a global that nothing in this component reads any more. So the picker's timezone is configured the same way as before by every call site (`EditFeeBillingPeriod`, `SubscriptionInformationFormSection`, `RateCardRateDrawerContent`, `FiltersItemDate` / `FiltersItemWebhookDate` / `FiltersItemLoggedDate`), and `FiltersItemIssuingDate` still deliberately passes none so it stays org-zone-relative — it just now gets the organization's zone reliably, instead of whichever sibling picker last wrote the global.

Removing the prop itself would mean every picker rendered in the org zone, which would reintroduce BIL-677 on the UTC-backed call sites.

---

ansmonjol (approved):

PR #4302, opened 2026-09-14
feat(wallets): connection routing on wallets and recurring rules, read and write

domenicofalco:
## Context

Per-object routing puts the connection first: the payment method becomes a sub-choice of the resolved connection. The wallet is the first billing object whose GraphQL door is open — lago-api#6387 adds `connections: ConnectionsInput` to the wallet create/update inputs and to both recurring-rule inputs, and lago-api#6401 exposes the stored routing back.

The existing payment settings drawer is in production, so the connection-first drawer ships as a separate component rather than as a flag threaded through the old one: with `multi_connection` off the mount points open literally today's code path, which makes the flag-off behaviour structurally impossible to regress and turns the eventual flag removal into a deletion.

## Description

Three surfaces, on the shared primitives in `src/components/connectionSelection/`.

**Shared primitives.** `ConnectionBehaviorFields` is a category-agnostic three-branch control — use the customer default / use a specific connection / skip — taking the category's own copy, plus two slots: what the choice itself owns (rendered above the rule that closes it) and what follows the rule. `CustomerPaymentConnectionComboBox` and `CustomerIntegrationConnectionComboBox` sit over the customer's connections, with a `Default` chip and an empty state. `fromConnectionRouting` maps the read side back into form values.

**Connection-first payment drawer** (`paymentSettings/connectionFirst/`), mounted on the wallet form and inside the recurring-rule drawer, each rule carrying its own choice. The payment-method choice nests under the inherit and specific branches and renders only once a connection resolves; manual moves up to the connection level, so the method control drops its manual branch. Its list comes from the customer's methods filtered by the resolved connection — the connection-scoped hook reads through the singular `providerCustomer` and can only answer for one connection.

**Additional integration settings drawer** (`additionalIntegrationSettings/`) does the same for accounting, CRM and tax, without the method sub-choice. Three sections on the same control, which is what proves it is genuinely category-agnostic.

**External apps tab** on the wallet and on each recurring rule, reading the stored routing.

**Payload.** Each category is serialised as exactly one of `{ behavior }` or `{ code }`, as the backend validator requires, and omitted entirely when untouched — `inherit` destroys the stored override row. Reading back branches on `behavior` before `code`: an inherited category reports the customer default's code, and carrying it would turn inheritance into an explicit override on the next save. With the flag off the mappers drop `connections` from the payload altogether, since the backend refuses any payload carrying it while gated.

## Outside the feature

Two fixes to the loop scripts, forced by breakage hit while working here rather than planned: `scripts/loop-plan-check.test.sh` cleared the git environment it inherits from hooks (it was addressing the real repository instead of its sandbox), and `agents:check` left the pre-push gate, which now runs the skill budget alone.

## Verification

Behind the `multi_connection` flag. Developed against the lago-api branch carrying #6387 and #6401; the `Run Codegen` check builds the schema from lago-api main and stays red until those merge.

Round trip verified in the browser against a running API: a specific connection saved on a wallet comes back as `behavior=specific` with its code, and the drawer reopens on that choice.

Fixes ING-519
Fixes ING-520
Fixes ING-527


ansmonjol (approved):

domenicofalco:
Quality gate is green on this run: 0 new issues, 0 security hotspots, 94.2% coverage on new code and 0.0% duplication. Nothing actionable in this report, so no change on the branch for it.

---

PR #4303, opened 2026-09-14
refactor(catalog): make rate-card regroupPaidFees nullable

rsempe:
## Context

The rate_card `regroupPaidFees` changed on the API ([LAGO-1811](https://linear.app/getlago/issue/LAGO-1811) — getlago/lago-api#6398): the enum dropped its `none` value and the field became **nullable**. `null` now means the paid fee stays standalone, aligning the rate card with the legacy charge contract.

This is the FE half of #6398 and must merge with it.

## Description

- Dropped `None` from `RateCardRegroupPaidFeesEnum` and made `regroupPaidFees` nullable across the generated types (object type + all query/fragment result types). Edited surgically to avoid sweeping unrelated schema drift.
- Updated the 5 test fixtures that used the removed `RateCardRegroupPaidFeesEnum.None` to `null`, and removed the now-unused enum imports.
- **No component change needed** — the drawer form logic (`constants.ts`) already maps the non-regrouped strategies (`invoiceable` / `none`) to `regroupPaidFees: null` and only ever writes `.Invoice`, so it was already null-native.

## Verification

- `tsc --noEmit`: clean (0 errors).
- Jest: constants, RateCardsList, RateCardDetailsOverview, RateCardPreview, useRateCardTableActions/Columns — 62 passed.
- eslint clean.

## Deploy coupling

Merge together with getlago/lago-api#6398. The FE `Codegen` CI builds lago-api from `main`, so this PR's codegen only goes green once #6398 is merged (same order as the previous pairs).


src/pages/catalog/__tests__/RateCardsList.test.tsx
```diff
@@ -107,7 +102,7 @@ const buildRateCard = (
   description: null,
   billingTiming: 'advance' as RateCardForListFragment['billingTiming'],
   displayOnInvoice: true,
-  regroupPaidFees: RateCardRegroupPaidFeesEnum.None,
+  regroupPaidFees: null,
```
ansmonjol, line 105: Type expected is undefined instead no?

ansmonjol: Oh the error is because of the codegen, but I guess you ran it against the BE PR

ansmonjol (approved):

src/pages/catalog/__tests__/RateCardsList.test.tsx
rsempe: yep 👍

PR #4309, opened 2026-09-15
feat(wallets): add the additional integration settings drawer

domenicofalco:
## Context

Wallets and their recurring rules can only route their payment connection.
Accounting, CRM and tax connections still fall back to the customer defaults
with no way to override or skip them per wallet, and the backend now accepts
those three categories on the wallet and recurring-rule inputs.

## Description

Adds `AdditionalIntegrationSettingsDrawer`: one drawer holding three
`ConnectionBehaviorFields` sections (accounting, CRM, tax), each a
fall back / specific / skip choice with a customer-connection combobox on the
specific branch and no nested sub-choice. The drawer is opened from a new
`AdditionalIntegrationSettingsSelector` card mounted on the wallet form and
inside the recurring rule drawer, so a rule's choices stay independent from the
wallet-level ones.

- new lean query `CustomerIntegrationConnections` behind
  `useCustomerIntegrationConnections`, bucketing `integrationCustomers` per
  category and resolving each connection's name and provider group from the
  org-level integrations
- new `CustomerIntegrationConnectionComboBox`, category-scoped, with a stale
  value guard
- `findPaymentRouting` generalised into `findConnectionRouting(connections,
  category)`, both call sites updated
- `formatConnections` widened to the four categories, emitting one key per
  touched category only, so an untouched category keeps whatever is stored
- read-back seeds the three new form fields from the routing array, an
  inherited row still yielding no override
- everything behind the `multi_connection` feature flag: flag off renders
  nothing new

Fixes ING-520

---

Stacked on #4302 (ING-519) — review that first. `Run Codegen` stays red until lago-api#6387 merges: CI builds the schema from lago-api `main`, which has no `connections` input yet.

Copy strings were derived from the ING-519 wording and the existing category label keys — the Figma could not be opened (seat tool-call limit), so labels deserve a designer's eye. The new card also ships without the subhead + caption block its sibling sections have on the wallet form; confirm with design whether it should get one.


src/hooks/customer/useCustomerIntegrationConnections.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export const useCustomerIntegrationConnections = ({
+  customerId = '',
+  category,
+  skip = false,
+}: UseCustomerIntegrationConnectionsArgs): UseCustomerIntegrationConnectionsReturn => {
+  const { data, loading } = useCustomerIntegrationConnectionsQuery({
+    variables: { customerId },
+    skip: skip || !customerId,
+  })
```
ansmonjol, line 97: `useConnectionOptions()` fires **four** org-wide queries (`paymentProviders` plus three separate `integrations(limit: 1000)` operations) purely to resolve a display name and group, and folds all four into `loading`.

Two consequences:

- The accounting/CRM/tax combobox sits in its loading state until the entirely unrelated `paymentProviders` query resolves, so a slow or failing payment-provider fetch blocks a select that needs none of it.
- The hook is mounted 3-6 times per drawer (once per `AdditionalIntegrationSettingsSection`, again inside each revealed `CustomerIntegrationConnectionComboBox`), so opening the drawer can mount up to 24 query subscriptions.

Note also that `skip` only gates the customer query: passing `skip: true` still runs all four org queries.

The sibling `useCustomerPaymentConnections` deliberately takes only `usePaymentProviders()`. This one should take the single category-specific provider hook (`useAccountingProviders` / `useTaxProviders` / `useCrmProviders`) instead.

src/components/additionalIntegrationSettings/AdditionalIntegrationSettingsSection.tsx
```diff
@@ -0,0 +1,152 @@
...
+  value,
+  onChange,
+  error,
+}: AdditionalIntegrationSettingsSectionProps) => {
+  const { translate } = useInternationalization()
+  const { defaultConnection } = useCustomerIntegrationConnections({ customerId, category })
+
+  const renderBadge = (optionBehavior: ConnectionBehavior) => {
+    if (optionBehavior !== ConnectionBehavior.INHERIT) return null
+
```
ansmonjol, line 81: `renderBadge` treats "connections not loaded yet" and "customer has no default" identically: it ignores the `loading` the hook returns (and the hook discards the query `error` entirely).

Scenario: open the drawer on a slow connection and all three inherit options, the options selected by default, are labelled "No default connection assigned on customer", the exact opposite of the truth, until the query lands. If the query errors it stays wrong permanently, and the user can reasonably pick "skip" or "specific" on that false premise.

Destructure `loading` alongside `defaultConnection` on line 67 and render nothing (or a skeleton) while it is true.

src/components/connectionSelection/CustomerIntegrationConnectionComboBox.tsx
```diff
@@ -0,0 +1,47 @@
...
+  customerId,
+  category,
+  value,
+  onChange,
+  error,
+  PopperProps,
+}: CustomerIntegrationConnectionComboBoxProps) => {
+  const { translate } = useInternationalization()
+  const { options, loading } = useCustomerIntegrationConnections({ customerId, category })
+
```
ansmonjol, line 32: This stale-value guard is visual only. `selectedValue` blanks a `value` missing from `options`, but the form still holds `{ code: 'X' }`, and `connectionChoiceSchema` only refines on `!value?.code`.

Scenario: a wallet stores `accounting: { code: 'netsuite_old' }` and that customer connection is later deleted. Reopening the drawer shows "Use a specific connection" selected with an empty select and **no** validation error, and Save succeeds, silently re-persisting the dangling code.

Either clear the form value when the guard trips (`onChange('')`), or surface the mismatch as an error once `loading` is false. The same pattern exists in `CustomerPaymentConnectionComboBox` from ING-519, so a fix likely belongs in both.

ansmonjol (commented):
Review of the drawer. The build is solid overall: the mapper round-trip (`formatConnections` per-category emission, `findConnectionRouting` read-back, rule-vs-wallet independence, the `...rest` destructure keeping FE-shaped fields out of the payload), the flag gating and the drawer wiring all check out, and the translation keys resolve. Nesting the drawer inside `RecurringRuleDrawer` is safe: `useFormDrawer` mints a per-`useId` NiceModal, `drawerStack` supports stacking, and `BaseDrawer` returns `null` when unmounted so `ConnectionBehaviorFields`' internal state is re-seeded on each open. `shouldPromptOnClose: () => form.state.isDirty` is fine too: TanStack 1.33 derives `isDirty` from sticky per-field flags that `form.reset` clears, so seeding does not produce a false prompt.

Three issues inline. Two I checked and deliberately did not flag: `fieldMeta[category]?.errors?.[0]?.message` does work here (`form.setFieldValue` registers `fieldInfo`, and it is always called before a "specific with no code" error is reachable), matching the ING-519 pattern; and re-sending accounting/CRM/tax on flag-off orgs is harmless, since those routings read back as `inherit` -> `undefined` -> omitted from the payload.

ansmonjol (approved):

src/hooks/customer/useCustomerIntegrationConnections.ts
domenicofalco: Good catch, thanks! Applied in ac5e0b1e4. The hook now goes through a new `useCategoryIntegrationOptions`, which only queries the provider list of the category it was asked for — `paymentProviders` is gone and the other two categories are skipped, so an accounting select no longer waits on queries it does not use. `skip` gates the org query too now, and `useConnectionOptions` reuses the extracted option mapper so there is only one copy of it.

src/components/additionalIntegrationSettings/AdditionalIntegrationSettingsSection.tsx
domenicofalco: Good catch, thanks! Applied in ac5e0b1e4. `loading` is destructured alongside `defaultConnection` and the badge renders nothing while it is true, so the inherit option no longer claims there is no default before the answer is in.

src/components/connectionSelection/CustomerIntegrationConnectionComboBox.tsx
domenicofalco: Good catch, thanks! Applied in ac5e0b1e4. The combobox now clears the form value (`onChange('')`) once loading is done and the stored code is not among the options, so the schema's empty-code refinement fires and Save is blocked instead of re-persisting a dangling code. It deliberately does not clear while the options are still loading. Left `CustomerPaymentConnectionComboBox` untouched on purpose: it belongs to #4302, which this PR is stacked on — happy to mirror the fix there so it lands in one place rather than showing up in this diff.

PR #4310, opened 2026-09-16
fix(integrations): scope the ValueAlreadyExist check to the code key

domenicofalco:
## Context

Six integration dialogs (Anrok, Avalara ×2, Hubspot, Netsuite, Salesforce, Xero) called `hasDefinedGQLError('ValueAlreadyExist', errors)` with no key, so a collision on any other unique field would paint itself onto the Code input. The catalog drawers already pass `'code'`; these six were the gap.

Narrowing the check was held back because nothing in the frontend records which `details` key each mutation reports the collision under — passing `'code'` blindly risks turning a misplaced error into a missing one. That dependency is gone: on the five mutations that do not silence the rejection, a collision under any other key now falls through to the error link, which raises the same generic message and captures a Sentry event. That event is how we find out which key the API actually uses.

No production occurrence of the misdirected error is known — this is defensive, and an alignment with the rest of the codebase.

## Description

- Scope the check to `details.code` at all seven call sites.
- Anrok (both branches) and Avalara on edition pass `silentErrorCodes: [UnprocessableEntity]`, which mutes the error link, so those two carry an explicit generic toast for the non-`code` case — the shape `useCustomerConnectionsPersistence` already ships. The other five need nothing: the link already handles it.
- Tests: a per-dialog negative case (a collision under a non-`code` key leaves the Code input clean, and surfaces where that dialog expects it), sharing one `describeDuplicateRejectionRouting` contract. Verified by mutation — unscoping the check fails exactly those 7 assertions.

Fifteen lines of production change, no new helper, no new module.

Fixes LAGO-1935


domenicofalco:
**SonarCloud Quality Gate — Duplication on New Code** is the only failing check (everything else, including E2E, is green). I can't get it under the 3% bar within this PR's scope. Here is what I measured and why.

### Root cause

`sonar-project.properties` excludes `**/__tests__/**` from `sonar.sources`, so duplication is computed over the **production** lines of this diff only — about 59 lines of code. Five of them fall inside the 10–12 line **identical import header** shared by `AddHubspotDialog.tsx`, `AddNetsuiteDialog.tsx` and `AddXeroDialog.tsx`:

```
import { Alert } from '~/components/designSystem/Alert'
…
import { addToast, envGlobalVar } from '~/core/apolloClient'          ← touched (drops hasDefinedGQLError)
import { evictFromCache } from '~/core/apolloClient/evictFromCache'
import { applyExistingCodeErrorOrToast } from '~/core/form/existingCodeError'   ← touched (renamed symbol)
```

5 / 59 = 8.5%, which matches the reported 8.6%. The duplication is pre-existing copy-paste between sibling dialogs; this PR only *touches* two lines of it per file, which reclassifies them as new duplicated lines.

### What I tried

I extracted the four byte-identical `handleError` bodies into a shared `handleDuplicateCodeError` module (`fc246c0`). That made it **worse — 12.2%**: it shrank the new-source-line denominator and moved the remaining new lines into the larger copy-pasted mutation block. Reverted in `ffb44be`, so the PR is back to the minimal in-scope diff.

I also de-duplicated the test side (`aa65535`, `describeDuplicateRejectionRouting`) — worth keeping on its own merits, but it doesn't move this metric since tests aren't in `sonar.sources`.

I'm not re-running the check: the analysis is deterministic, not flaky.

### Why it isn't fixable here

Any correct change to this error handling has to edit code that is copy-pasted across four sibling dialogs, and the diff is deliberately small — so duplicated-new-lines density is structurally high no matter how the change is written. Getting under 3% means factoring out the shared dialog scaffolding (the identical import header, the Nango connect block, the `onCompleted` toast + navigate), which is a much larger refactor than "scope the check to the code key".

### Asking for a call

Either accept the gate failure on this PR, or I split the scaffolding refactor into its own PR and rebase this on top. Happy to do the latter — just say which.

---

domenicofalco:
Resolved — no decision needed, the question above is stale.

Reworking the change removed the duplication at the source rather than working around it. The helper that raised a toast from `~/core/form/` is gone (no other helper there raises one, and its "callers must silence the error at the link" contract was coupling it could not enforce), and with it the `silentErrorDetails` added to five mutations: those mutations don't mute the error link, so a collision under a non-`code` key already surfaces the same generic message — and the Sentry event it captures is how we find out which key the API actually uses. Only Anrok and Avalara-on-edition, which do mute the link, carry an explicit fallback.

What's left is 15 lines of production change across 6 files — on five call sites literally `, 'code'`. Quality Gate now passes: **0.0% duplication on new code**, 0 new issues, 100% coverage on new code. The PR body is updated to match.

---

ansmonjol (approved):

PR #4312, opened 2026-09-16
chore(agents): harden the loop pipeline with mechanical caps and a plan-first build

domenicofalco:
## Context

The loop is the pipeline that takes a Linear ticket and produces a PR on its own: it writes a spec, builds the change, reviews it in a separate agent, then opens the PR. Its instructions live in the `loop-*` skills under `.agents/skills/`.

Over the last six weeks those instructions kept growing. Every time a run went wrong, a new rule was added; nothing was ever removed. The reviewer went from about 5 checks to 15. This has three costs:

- **Rules stop working.** An agent reading 15 checks gives each one a fraction of the attention it would give 5. The "no useless comments" rule is written in three places and comments still show up on every PR.
- **Runs get more expensive.** Every phase reads its whole skill on every run, re-fetches the ticket from Linear, re-fetches the styleguide from Notion, and spawns a second review agent that has stalled on every recorded run.
- **The code drifts from the ticket.** The most expensive PR the loop produced solved a problem the ticket described but the codebase did not have: nobody checked the ticket's claim against the code, and the reviewer had no way to tell that 60 lines could have been 15.

This PR does not add rules. It removes some, caps the rest, and replaces prose with checks that run as scripts.

## Description

**1. The instructions cannot grow past a cap**

- A new script, `scripts/skill-budget.sh`, counts the checks in each loop skill and fails if there are more than allowed: 7 in the reviewer, 7 in the builder, 5 in the spec phase, and the orchestrator skill cannot exceed 150 lines. It also fails if any Linear ticket ID appears under `.agents/` (a ticket ID in a skill is a dangling reference the agent cannot follow). It runs on every push through `code:style`, together with two small test suites for the gate scripts. The whole step takes about one second.
- To get under the cap, the reviewer's 15 checks became 7: the three test-related checks are one, the four "follow the call into existing code" checks are one, and lint/types/translations left the reviewer entirely because they were already scripts.
- Every review failure is now tagged with the check that found it, and the run journal records those tags. That is what lets us delete a check later: one that has found nothing in 15 runs goes.
- The flywheel skill, which turns run feedback into skill edits, now has to follow four rules: a new check enters only by removing one; a proposal must cite a failure and a second plausible case, not just "this would have helped here"; anything that can be a script or a lint rule must be written as one; checks the journal shows to be dead get pruned.
- All real ticket IDs were removed from skills and docs. Examples use `<ISSUE-ID>`.

**2. Each run costs less**

- The orchestrator skill went from 262 to 123 lines. The bash it carried inline (container restart, CI log capture, journal row) moved to three small scripts it now calls by name.
- The spec phase copies the ticket into `spec.md`. The builder and the reviewer read it from there and never call Linear or Notion again.
- The Notion coding styleguide is mirrored in `.agents/docs/frontend-coding-styleguide.md` and refreshed by the flywheel, instead of being fetched on every build.
- The reviewer's second pass (`/code-review`) runs in the same agent. It used to run as a separate subagent, which stalled every time and came back late against code that had already changed.

**3. The code stays close to the ticket**

- The spec phase now verifies every factual claim the ticket makes about the code and records where it is confirmed, or marks it `unverified`. A criterion built on an unverified claim is flagged before anyone writes code.
- Before writing code, the builder writes a short plan: which files it will change, which new files and new exports it intends to add, and why. The default for new files and exports is "none".
- A script compares the finished diff with that plan. If a new file or export appeared that the plan did not declare, a second review agent is dispatched. It works out the smallest possible change for the ticket without looking at the actual diff, then compares. If inlining the new abstraction gives a smaller diff, that is a failing finding. This is the one check that catches over-engineering, and it only runs when there is something to catch.
- A `--confirm-spec` flag pauses the run once, after the spec, so a human can ask "why does this exist?" before any code is written. Off by default; nothing changes for runs that do not use it.

**4. Comments**

The "default to no comments" rule is now a gate, not a sentence. `scripts/diff-hygiene.sh` flags every comment the diff adds unless the builder's plan lists it with one of three reasons: an external constraint the reader cannot see, why the obvious alternative is wrong, or a trap that bites on edit. Comments in a `type`/`interface` body (prop documentation), above an import, or one-liners that just repeat the export's name are refused even if declared. Tool directives like `eslint-disable` are exempt. Expected result on a normal diff: zero comments, and each one that survives has a written reason next to it.

**Not in this PR**

- Validating the new pipeline on two real tickets. That happens on the next two runs and the journal records it.
- The daily scheduled task that runs the loop. It does not live in this repo and still needs to be pointed at `/loop-run` so it writes journal rows.

Fixes STA-13


domenicofalco:
Thanks, all five were right. Addressed in 98d4e9400:

- **Styleguide mirror missing** — good catch, and the cause is worth knowing: the root `.gitignore` has a `docs/` pattern that also ignores new files under `.agents/docs/`, so the file was written but never staged. Committed with `git add -f`.
- **Adversarial trigger vs build gate** — `loop-plan-check.sh` now distinguishes the two: exit 1 = undeclared (gate fails), exit 3 = new files/exports exist and all are declared (loop-run dispatches the adversarial pass), exit 0 = nothing new. Skills updated to match.
- **Re-exports** — `export { a, b as c }` (each name, alias when present) and `export * from '<module>'` (module path must be in the plan) are detected.
- **Hardcoded `revised`** — loop-revise now passes the actual outcome.
- **Coverage and wiring** — `loop-helpers.test.sh` covers `loop-journal.sh`, `loop-restart.sh` and `loop-ci-log.sh` with stubbed `docker`/`gh`; `agents:check` runs all three suites at push (about one second).

ansmonjol (approved):

PR #4315, opened 2026-09-16
fix(scripts): clear the inherited git env in the plan-check test

domenicofalco:
## Context

Git exports `GIT_DIR` — and `GIT_INDEX_FILE`, `GIT_WORK_TREE`, … — to its hooks, and those variables **take precedence over `git -C`**.

`scripts/loop-plan-check.test.sh` runs inside `agents:check`, which `code:style` runs, which the pre-push hook runs. So every `git -C "$wt"` in it addressed the real repository rather than the sandbox it had just created: `add -A && commit -qm base` recorded the whole checkout as deleted, and `update-ref refs/remotes/origin/main` repointed that checkout's ref at the fixture.

That is what happened on a worktree yesterday, twice. The sandbox guard from #4313 detects it — but detecting it during a hook means the hook exits 1, so **any push from a branch carrying that guard is now refused**. This restores pushes and removes the cause.

## Description

The test clears the inherited git environment before doing anything, so `git -C "$wt"` addresses the sandbox whether the script is invoked by hand, through `agents:check`, or by a hook. The guard stays, now as the assertion that this actually held.

## Verification

| invocation | before | after |
|---|---|---|
| standalone | passes | passes |
| with `GIT_DIR` set to the real repo (what a hook does) | refuses, exit 1 | passes, exit 0 |

Working tree untouched in both cases. The three sibling test scripts (`skill-budget`, `loop-helpers`, `conductor-front-container`) already pass under a hook environment — checked, not assumed.


domenicofalco:
Chiusa: il commit `b8cc20520` è già incluso nel branch di ING-519 ([#4302](https://github.com/getlago/lago-front/pull/4302)), che lo porterà su main al merge.

PR #4330, opened 2026-09-21
fix(invoices): Preserve zero values when regenerating invoices

domenicofalco:
## Context

Void and regenerate displayed a zero-priced fee but omitted its numeric zero price from the submitted payload. If the underlying charge price had increased, saving could charge $2,000 despite the $0 preview.

## Description

Preserve zero prices and quantities while omitting null, undefined and empty strings. Keep the input field types through a generic Partial<T> return type. Cover the rendered submit flow with regression tests for numeric zero, string zero and zero units, using a stable submit button test identifier. Replace copied redirect logic in tests with assertions against the real mutation completion callback.

Isolate the loop plan checker test from Git variables inherited by pre-push hooks so its temporary repository cannot resolve to the real checkout.

Validation: the invoice regeneration suite passes all 26 tests; pnpm code:style passes. The sandbox test also passes with inherited GIT_DIR and GIT_WORK_TREE while leaving the real checkout unchanged. Browser QA of the ticket's price-change scenario remains pending.

Fixes BIL-711


domenicofalco:
Thanks for the analysis! The quality gate passes with no new issues and 100% coverage on new code; no changes needed in response.

domenicofalco:
Thanks for flagging the loss of field types. Addressed in e7bcb132b: `removeEmptyKeys` now returns `Partial<T>` for its input type, without type assertions, while preserving numeric zero and omitting only null, undefined and empty strings.

The previous commit's typecheck was green, so I could not confirm the reported compilation failure; the type-preservation improvement is now applied. Local `pnpm code:style` and all 26 regeneration tests pass. Please rerun the advisory pre-review on the new commit.

domenicofalco:
All CI checks now pass on e7bcb132b, including Typecheck, the full unit-test pipeline and Cypress. Thanks for the updated SonarQube report: no new issues and 100% coverage on new code, so no further changes are needed for that report.

The previous HOLD references 0db6c802f; its requested generic return type is addressed in e7bcb132b. Awaiting the refreshed advisory pre-review.

ansmonjol (approved):

PR #4333, opened 2026-09-21
refactor(invoices): migrate the invoice metadata drawer to TanStack Form

ansmonjol:
## Context

The invoice "Add / Edit metadata" drawer (`src/components/invoices/AddMetadataDrawer.tsx`) was one of the remaining Formik + yup forms flagged by the `no-restricted-imports` lint warning. It also used the oldest drawer generation — `~/components/designSystem/Drawer` + a `DrawerRef` held by the parent — which CLAUDE.md rules out for new code.

## Description

Migrates the drawer to TanStack Form + zod and to the canonical `use<Feature>Drawer` hook, following the already migrated `ItemMetadataDrawer`.

**New `src/components/invoices/addMetadataDrawer/`**

- `useAddMetadataDrawer.tsx` — owns the query, the mutation, the `useAppForm` form (`revalidateLogic()` + `onDynamic`) and the `useFormDrawer` opening. Returns `{ openDrawer }`, so consumers no longer hold a ref nor render a drawer element.
- `AddMetadataDrawerContent.tsx` — the drawer body, split out with `withForm`.
- `constants.ts` — form id, `MAX_METADATA_COUNT` (5), `METADATA_VALUE_MAX_LENGTH` (255), default values.

**Shared rows** — both metadata drawers render the same key/value editor, so the rows grid, the per-field error tooltips and the add/remove controls now live in one `MetadataRowsFieldGroup` (`withFieldGroup`) used by this drawer and by `ItemMetadataDrawerContent`. Copy, limits, row cap, grid spacing and the add-button test id stay per drawer through props, so both keep their current UI and their own validation limits.

**Validation** reuses the existing `zodMetadataSchema(255, METADATA_KEY_MAX_LENGTH)` from `src/formValidation/metadataSchema.ts`, rule-for-rule equivalent to the yup `metadataSchema({ valueMaxLength: 255 })` it replaces: key required + unique across rows + 20 characters max, value required + 255 characters max. The shared yup schema is untouched — four other forms still use it.

**Behaviour**

- Same copy and same creation / edition modes. The mode is captured when the drawer opens, so the success toast keeps matching the state the user started from even after the cache updates.
- Same 5-row cap, same add / remove row.
- The mutation now sends an explicit `{ id, key, value }` per row instead of spreading the query fragment (which carried `__typename`), so existing pairs are still updated by id.
- The drawer stays open when the mutation fails and closes on success; leaving with unsaved changes prompts for confirmation; pressing Enter in an input submits, since the drawer renders a real `<form>`.
- Following the migration convention, the submit button is no longer gated on `isValid` / `dirty`: it stays enabled and validation errors surface on submit, with `scrollToFirstInputError` bringing the first invalid row into view. Creation mode therefore seeds one empty row, so saving an untouched drawer reports the missing key and value instead of silently sending an empty payload.
- A row carrying several issues at once (an empty key that is also a duplicate) now shows the first known message instead of falling through to no error state at all — this also fixes the same gap in the item metadata drawer.
- Saving replaces the whole metadata array, so the drawer never edits a list it has not read: opening seeds from the query, fetching it first when it has not landed yet (the invoice details page does not prime that fragment), and reporting an error instead of opening when the metadata cannot be read.

**Consumers** — `Metadatas.tsx` and `CustomerInvoiceDetails.tsx` call the hook and drop their refs and drawer elements.

**Tests** — `AddMetadataDrawer.test.tsx` is replaced by `addMetadataDrawer/__tests__/useAddMetadataDrawer.test.tsx` (19 cases): query silencing and skipping, both modes, the save payload, add / remove row, the 5-row cap, prompt-on-close, a failed mutation, the three validation rules, and the three seeding paths (cached, fetched, unreadable). `Metadatas.test.tsx` now asserts the drawer wiring instead of stubbing it away. The existing item-metadata and plan-metadata suites cover the shared rows component.

Fixes LAGO-1206


ansmonjol:
Confirmed and fixed in fa74ac4 — thanks, this was a real regression.

The Formik version had `enableReinitialize: true`, so a metadata response arriving after the drawer opened re-seeded the form. The migrated hook snapshotted `data` at open instead, and `CustomerInvoiceDetails` does not prime `InvoiceMetadatasForMetadataDrawer` (only `Metadatas` does), so opening from the invoice details popper before that query landed seeded an empty form and a save replaced the existing pairs.

`openDrawer` now seeds from the query result, and fetches it first when it has not landed yet; a failed fetch still opens the drawer empty rather than leaving the click dead. The cached path stays synchronous, so the common case is unchanged. Two regression tests cover both branches (fetch-then-seed, and open-anyway on a failed fetch).

---

ansmonjol:
Fair — fixed in b4ab1e1. Since a save replaces the whole metadata array, opening on a failed fetch let the user commit a list built from nothing and drop the pairs that only failed to load.

The drawer now refuses to open when the metadata cannot be read (fetch rejected, or the invoice comes back empty) and the action reports the error instead, so the existing pairs stay untouched until the list is known. The test that expected the empty fallback is replaced by two that assert the drawer stays closed and the error is surfaced.

---

domenicofalco (approved):

